PR #428, opened 2015-02-26
add metadata_xform.py script to bin/

sklump:
Converts ckan-v2.1 style metadata to ckan-v2.3 with wet4-scheming.


bin/meta_xform.py
```diff
@@ -0,0 +1,185 @@
...
+
+# new schema description, data field choices
+sd_new = None
+sd_new_dfc = None
+
+def _is_gzipped(fpath):
+    """
+    Return True if input path is a gzipped file, False otherwise
+    """
+    cmd = shlex.split('file --mime-type {0}'.format(fpath))
```
wardi: this will fail if your file path contains spaces (and other special shell characters)

I think it's better to remove this `_is_gzipped` function and let the code fail in the gzip module instead. If you would like a friendlier error just catch the exception raised from gzip in the main function.

bin/meta_xform.py
```diff
@@ -0,0 +1,180 @@
...
+    _HERE = path.dirname(path.abspath(__file__))
+    _SCHEMAS_DIR = path.join(path.dirname(_HERE),'ckanext/canada/schemas')
+    _JSON_NAME = path.join(_SCHEMAS_DIR, 'raw.json')
+
+    with open(_JSON_NAME) as j:
+        sd_new = simplejson.load(j)
+    sd_new_dfc = {
+        df['field_name'] : {
+            ch['label']['en'].replace(',', '') :
+                ch['value'] for ch in df['choices']}
```
wardi: this generator expression is incompatible with python 2.6, please use a list comprehension for now (just add [] around it)

wardi: sorry, dict comprehension , not generator expression. You can use dict([(key, value) for ... ]) instead

PR #436, opened 2015-03-12
bilingual message for 'Bilingual (English and French)', recombinant error conditions

sklump:
changes to canada.po


wardi:
@sklump this PR is mostly commit noise. Can you collapse it to a single commit against master and resubmit?

PR #437, opened 2015-03-12
merge from master

wardi:


wardi:
@thriuin look over the diff here please

thriuin:
@wardi  Looks good to me!

PR #441, opened 2015-03-26
Support auto-post of combined datasets from registry to portal

sklump:
ckanext-recombinant:
+ recombinant_tables.json supports "target_dataset" per dataset_type (e.g., "ati" for "ati-summaries" and "ati-none")
+ paster recombinant command target-datasets command to return all target datasets defined
+ paster recombinant command dataset-types lists all dataset types per input target

ckanext-canada:
+ ckanext/canada/common code between ati.py, pd.py refactored and generalized in dataset.py
+ ckanext/canada/ati.py, pd.py know their respective target datasets and pass them to dataset.py, which gets corresponding dataset types on the fly
+ bin/reg2portal.sh and reg2portal.py combine dataset resources and upload to corresponding portal datasets
+ bin/csv2solr.sh on portal will update solr core from uploaded csv files, but is still a work in progress

### DIFF
--- a/bin/csv2solr.sh
+++ b/bin/csv2solr.sh
@@ -1,59 +1,100 @@
 #!/bin/bash
 
 function usage() {
-    echo "Usage: ${0} CKAN-INI-FILE [VIRTUAL-ENV-HOME]"
-    echo "The script:"
+    echo "Usage: ${0} CKAN-INI-FILE \\"
+    echo "    TARGET-DATASET:PACKAGE-ID ... [VIRTUAL-ENV-HOME]"
+    echo
+    echo "This script:"
     echo "- activates virtual environment if specified"
     echo "- pushes datasets from .csv files to the solr core"
+    echo
+    echo "e.g., ${0} ../development.ini ati:ati_id pd:pd_id /opt/my-venv"
+
     exit -1
 }
 
-#
-#
-# THIS SCRIPT AS IT IS WRITTEN IS HOPELESS: NEEDS TO USE CKANAPI FILE STORAGE
-#
-#
+function log() {
+    DT=$(date +'%F %T')
+    LEVEL=${1}
+    MSG=${2}
+    echo ${DT} ${LEVEL} [$(basename "${0}")] ${MSG}
+}
 
-if [ "$#" -eq 0 ]
+LAST_TARG_ARG_IDX=$#
+if [ "$#" -lt 2 ]
 then
     usage
-elif [ "$#" -gt 1 ]
+elif [[ ! "${!#}" =~ ":" ]]
 then
-   VENV_HOME=$(echo ${2} | sed -e 's./$..')
-   source ${VENV_HOME}/bin/activate
+    LAST_TARG_ARG_IDX=$(( $# - 1 ))
+    VENV_HOME=$(echo ${!#} | sed -e 's:/$::')
+    source ${VENV_HOME}/bin/activate
 fi
 
+# Establish bin directory
 cd "$(dirname ${0})"
 BIN_HOME=$(pwd)
 
-# Set ini file
+# Set .ini file, get port of operation from [server:main] section
 cd "$(dirname ${1})"
 INI_FILE=$(basename ${1})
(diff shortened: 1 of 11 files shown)


bin/reg2portal.py
```diff
@@ -0,0 +1,72 @@
...
+# Instantiate remote ckanapi
+portal_site = ckanapi.RemoteCKAN(
+    portal_url,
+    apikey=api_key,
+    user_agent=user_agent)
+
+# Update package: add resources
+try:
+    portal_site.call_action(
+        'package_update',
```
wardi: you want package_patch here instead. update will clear all the other fields too.

solr_schema.xml
```diff
@@ -133,7 +136,7 @@
 
     <field name="_version_" type="string" indexed="true" stored="true"/>
 
-    <dynamicField name="*_date" type="date" indexed="true" stored="true" multiValued="false"/>
+    <!--<dynamicField name="*_date" type="date" indexed="true" stored="true" multiValued="false"/>-->
```
wardi: revert the changes to this file, this is the solr schema for ckan not for ATI/PD search

bin/csv2solr.sh
```diff
@@ -0,0 +1,59 @@
...
+# Record ckanext-canada home
+cd "${BIN_HOME}/.."
+CXC_HOME=$(pwd)
+
+# Go to ckanext-recombinant
+cd "${BIN_HOME}/../../ckanext-recombinant"
+CXR_HOME=$(pwd)
+
+# Get target dataset names
+TARGET_DATASETS=$(paster recombinant target-datasets -c "${INI_PATH}")
```
wardi: Is there a recombinant PR that this relies on?

bin/csv2solr.sh
```diff
@@ -0,0 +1,59 @@
...
+    for F in ${DATASET_TYPES}
+    do
+      CSV_FILES="${CSV_FILES} ${TARG_DS}.${F}.csv"
+    done
+    CSV_FILES=${CSV_FILES/ /}
+
+    # rebuild solr core from identified .csv files
+    cd "${CXC_HOME}"
+    paster "${TARG_DS}" rebuild -f ${CSV_FILES} -c "${INI_PATH}" ${TARG_DS}.${DS_TYPE}.csv
+done
```
wardi: I think this script is complicated enough to warrant rewriting it as a `paster canada` subcommand instead. That way it's not tied to a particular directory layout and can just import and use the necessary recombinant functions.

wardi: on second thought I've made a bunch of changes to commands.py for 2.3. It's ok to leave this as is. Just add some docs to the README about how you run the scripts.

ckanext/canada/ati.py
```diff
@@ -55,56 +66,82 @@ def command(self):
...
+                        csv_file,
+                        e.message))
+                except AssertionError as e:
+                    logging.warning('On {0:s}, encountered: {1:s}'.format(
+                        csv_file,
+                        e.message))
+                except IOError as e:
+                    logging.error('On {0:s}, encountered: {1:s}'.format(
+                        csv_file,
+                        e.strerror))
```
wardi: This swallows the tracebacks and only keeps the error message. I'm ok with letting the traceback rise normally and cancel the whole process. We don't want to ignore errors here (and we certainly want all the information necessary to fix them).  Please remove the except: clauses here.

bin/csv2solr.sh
sklump: Yes - I had neglected it. Please check https://github.com/open-data/ckanext-recombinant/pull/25

sklump:
Ian,

I'm not certain about the solr-schema changes: the solr_schema.ati_pd.xml works on the portal. as is, but there may be less intrusive changes accomplishing the same thing.

I put myself in fits trying to squash and clean up the commit noise, but as the world can plainly see, it got the best of me. Anyway I am going to stop it here before I make things even worse. I was the subversion administrator for Bell Managed Services -- this is embarrassing, but at least the content is good.

wardi:
recombinant work was merged, now working on this pr over here: #451

PR #446, opened 2015-04-23
limit recombinant previews to 500 records

wardi:


wardi:
This is a fix for BH#785 and an ordering bug on selecting those records

PR #450, opened 2015-05-15
add missing dataset fields for importing existing datasets

wardi:


wardi:
The new metadata conversion script is now run as:

paster canada metadata-xform od-do.jl.gz new.jsonl.gz

And the datasets can then be loaded from new.jsonl.gz into the this version of the extension with the usual:

ckanapi load datasets -I new.jsonl.gz

PR #455, opened 2015-06-08
OGS maps cart

wardi:
@chrismajewski these changes look ok to me, so I've created the PR

I don't think any of the shopping cart ties in to scheming, because scheming just affects the metadata values shown in the dataset/resource view/edit pages and your changes just affect the search results page.


wardi:
As discussed, merging to wet4-maps instead and deploying to dev6

PR #457, opened 2015-06-18
hard code atom feed urls to open.canada.ca/data/api...

wardi:


wardi:
This works testing locally at /feeds/dataset.atom on my 2.1 environment

PR #483, opened 2015-09-30
Disable Shopping Cart till Open Maps schema is integrated

chrismajewski:
ATM I'm commenting it out, we'll eventually Jinja it out when Scheming can account for it.

### DIFF
--- a/ckanext/canada/schemas/presets.yaml
+++ b/ckanext/canada/schemas/presets.yaml
@@ -117,7 +117,7 @@ presets:
     # copied from fluent_markdown preset
     form_snippet: fluent_markdown.html
     display_snippet: fluent_markdown.html
-    error_snippet": fluent_text.html
+    error_snippet: fluent_text.html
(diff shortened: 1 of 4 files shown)


ckanext/canada/schemas/presets.yaml
```diff
@@ -2091,6 +2138,21 @@ presets:
...
+      en: Date Published
+      fr: Date de publication
+    help_text:
+      en: Date of formal issuance (e.g., publication) of the resource.
+      fr: Date officielle d’émission (c.-à-d. de publication) de la ressource.
+    required: true
+    # copied from date preset
+    form_snippet: date.html
+    display_snippet: date.html
+    validators: scheming_required isodate
```
wardi: I think for resources we need to add an extra step to make sure the result is a string or this will cause a failure

wardi:
merging anyway, the problem is outside the -canada extension

PR #508, opened 2015-10-14
PD: Extract year for use in Solr query

chrismajewski:
* YAML definable (extract_date_year: true) Solr year extract
* Also, a small name change to a PD type


ckanext/canada/pd.py
```diff
@@ -220,6 +220,10 @@ def unique_id(r):
             if f.get('datastore_type') == 'date':
                 try:
                     value = date2zulu(value)
+                    # CM: If this only applies to PD types this should be accurate
+                    # CM: This should only apply if valid (as per date2zulu) else NULL
+                    if key == 'date' or key == 'file_id_date' or key == 'contract_date':
```
wardi: kind of nasty.  Would be a little nicer if we flagged the field in our yaml files.

```python
if f.get('extract_date_year'):
```

PR #521, opened 2015-10-23
OI Reg: [pending] Fix dataset text in org list

chrismajewski:
Get dataset_type to work in this snippet and reload the .po and this should work


chrismajewski:
Confirm that it's dataset / asset and not just asset.

PR #578, opened 2016-02-05
PD: Travel annual overhaul

chrismajewski:
Lots of changes, mostly tripling fields accounting for current year, previous year and total.


wardi:
So many problems here.

This table is hugely duplicative. Why are we asking for the change for every type? Why are entering the previous and next years values instead of just the current year? It looks like some of the labels are also duplicated e.g. public_servant_current_year and public_servant_compared_fiscal_year
have the same label

But the biggest problem is that the year itself isn't being recorded.

Can we throw this one out and start again?

PR #582, opened 2016-02-10
Adding Aggregated Contracts schema

chrismajewski:


ckanext/canada/recombinant_tables.yaml
```diff
@@ -697,6 +848,13 @@
     datastore_type: text
     xls_column_width: 20
 
+  - datastore_id: date
+    label: Date / Date
+    datastore_type: date
+    xls_column_width: 20
+    extract_date_year: true
+    extract_date_month: true
```
wardi: can't we call this year? why are they year and month being extracted if this is just a year field?

ckanext/canada/recombinant_tables.yaml
```diff
@@ -541,6 +541,157 @@
   default_preview_sort: contract_period_start desc, delivery_date desc
 
 
+- dataset_type: contracts-aggregated
```
wardi: just a preference, but I'm aligning these types to short all-lowercased names that match our solr cores. How about `contractsa` (like existing `travela`) or `contractsagg` ?

PR #619, opened 2016-03-08
Added the megamenu for the registry

LaurentGoderre:


LaurentGoderre:
Can we merge now that it passed?

PR #645, opened 2016-04-05
Move our custom dataset deletion from GCWeb.

TkTech:
GCWeb is shared with other projects, we only want this functionality in
the public-facing portal.


ckanext/canada/templates/public/scheming/package/read.html
```diff
@@ -1,3 +1,14 @@
 {% ckan_extends %}
 
 {% block package_additional_info %}{% endblock %}
+
+{% block actions_content %}
+  <ul class="nav nav-pills">
+      {% if h.check_access('package_update', {'id':pkg.id }) %}
+        {% block actions_content_inner %}
+        <li>{% link_for _('Edit'), controller='package', action='edit', id=pkg.name, class_='btn btn-default' %}</li>
```
wardi: edit shouldn't be available on the portal should it?

ckanext/canada/templates/public/scheming/package/read.html
```diff
@@ -1,3 +1,14 @@
 {% ckan_extends %}
 
 {% block package_additional_info %}{% endblock %}
+
+{% block actions_content %}
+  <ul class="nav nav-pills">
```
LaurentGoderre: If you want your buttons to work off the bat, change the classes here to `list-unstyled list-inline`

PR #653, opened 2016-04-25
fix titles, script blocks

wardi:


ckanext/canada/templates/internal/user/login.html
```diff
@@ -29,12 +29,12 @@
   <div class="panel">
 
       {% if h.lang() == 'en' %}
-<h2>Welcome to the Government of Canada Open Information Portal Registry</h2>
```
LaurentGoderre: Why is this not using gettext?

wardi: that's a good question!

PR #654, opened 2016-04-25
Added temporary styles to make panel heading match the menu

LaurentGoderre:


ckanext/canada/templates/public/base.html
```diff
@@ -16,6 +16,14 @@
...
 {%- endblock -%}
 
+{%- block custom_styles -%}
+  {{ super() }}
+  {# Temporary styles until the panels styles are resolved in GCWeb #}
+  <style>
+    .panel .panel-heading {background: #335075; border:1px solid #335075}
+  </style>
+{% endblock %}
+
```
wardi: this would apply to the internal (registry) too right? that's not what we want

LaurentGoderre: It shouldn't...it in the public folder...

LaurentGoderre: .....Oohh...you're right...

PR #721, opened 2016-09-20
Hide Shopping Cart if not catalogue type "FGP"

chrismajewski:


chrismajewski:
Course change, net chnage = 0

PR #751, opened 2016-10-28
Updated translations

LaurentGoderre:


LaurentGoderre:
Don't merge yet. I will keep updating this with my progress.

ckanext/canada/schemas/presets.yaml
```diff
@@ -210,10 +210,10 @@ presets:
     fluent_form_label:
       en:
         en: Contributor (English)
-        fr: Contributeur (anglais)
+        fr: Contributor (Anglais)
```
wardi: these are all (anglais) and (français) in the other fields not (Anglais) and (Français)

ckanext/canada/templates/internal/recombinant/resource_edit.html
```diff
@@ -18,7 +18,7 @@
     <div class="panel panel-primary">
       <div class="panel-heading">{{ _("Guide") }}</div>
       <div class="panel-body">
-        <p>{% trans -%}For information on how to prepare, maintain and upload templates to the Registry, refer to the appropriate documentation posted on <a href="http://www.gcpedia.gc.ca/wiki/Proactive_Disclosure_on_Open.Canada.ca">GCPEDIA</a>{% endtrans %}</p>
+        <p>{{ _('For information on how to prepare, maintain and upload templates to the Registry, refer to the appropriate documentation posted on {link}').format(link=_('<a href="http://www.gcpedia.gc.ca/wiki/Open_Data_Inventory_/_l%E2%80%99inventaire_de_donn%C3%A9es_ouvertes">GCPEDIA</a>'))|safe }}</p>
```
LaurentGoderre: I don't quite understand what you mean. They wanted to change the link, and to make it work I split it into two

wardi: I bet they wanted to change the link only on the open information page, not all the PD pages too. We'd be better off putting this help link in the recombinant definition file, hacking the special case into the template will get kind of nasty

LaurentGoderre: Not sure what that would involve so I will remove this change from this PR and we can fix it later.

LaurentGoderre:
Ok, updated

PR #756, opened 2016-11-24
fix password reset message

fanjinfei:


ckanext/canada/templates/internal/user/login.html
```diff
@@ -52,7 +52,7 @@
     {% endif %}
 
     {% block form %}
-      {{ super() }}
+      {% snippet 'user/login_form.html', action=c.login_handler, error_summary=error_summary %}
```
wardi: you moved the snippet from user/snippets/login_form.html to user/login_form.html. Was there a reason for that?

ckanext/canada/templates/public/package/base.html
```diff
@@ -3,9 +3,11 @@
...
 
 {% block breadcrumb_content %}
-  {% set dataset = pkg.title[client_lang] %}
   {% set pkg_url = h.url(controller='package', action='read', id=pkg.id) %}
   <li>{% link_for _('Search Open Data'), named_route='dataset_search' %}</li>
-  <li>{% link_for dataset|truncate(30), controller='package', action='read', id=pkg.name %}</li>
+  {% if pkg.title_translated is defined %}
+    {% set dataset = pkg.title_translated[client_lang] %}
+    <li>{% link_for dataset|truncate(30), controller='package', action='read', id=pkg.name %}</li>
+  {% endif %}
```
wardi: please use the new `h.get_translated(pkg, 'title')` helper instead of this logic

ckanext/canada/templates/internal/user/login.html
fanjinfei: I will move it to /user/snippets

ckanext/canada/templates/public/package/base.html
fanjinfei: if use  `h.get_translated(pkg, 'title')`, the link http://localhost:5000/en/dataset/new will crash. 
`TypeError: string indices must be integers, not str`
because of pkg=''. How to solve this properly?

fanjinfei:
How do I delete those commit/revert pairs?

wardi:
@fanjinfei I wouldn't worry about it. I don't mind a little noise in the commit history.
But, if it really bothers you take a look at git's interactive rebase feature.

ckanext/canada/templates/internal/user/login.html
```diff
@@ -52,7 +52,7 @@
     {% endif %}
 
     {% block form %}
-      {% snippet 'user/login_form.html', action=c.login_handler, error_summary=error_summary %}
+      {% snippet 'user/snippets/login_form.html', action=c.login_handler, error_summary=error_summary %}
```
wardi: you also changed the variable passed to error_summary. Is the original block in ckanext-wet-boew incorrect? If so fix it there, if not, you don't need to override this block.

fanjinfei: In /usr/lib/ckan/default/src/ckan/ckan/templates/user/login.form, it uses such variable. Do you mean I should follow the one in wet-boew, as `error_summary=errors` ?

wardi: You tell me :-) you're either fixing a bug (error_summary=errors was wrong) or you're introducing a bug. If you're fixing a bug that should be done in ckanext-wet-boew. In either case you should remove this block from ckanext-canada so it doesn't override the one in ckanext-wet-boew.

PR #757, opened 2016-12-05
update resource file format list

fanjinfei:


ckanext/canada/schemas/presets.yaml
```diff
@@ -3166,6 +3178,10 @@ presets:
     - value: RDF
       mimetype: application/rdf+xml
       openness_score: 4
+    - value: RDF/XML
+      mimetype: application/rdf+xml
```
wardi: looks like a duplicate of RDF

ckanext/canada/schemas/presets.yaml
```diff
@@ -3089,13 +3091,18 @@ presets:
     - value: GeoTIF
       mimetype: image/tiff
       replaces: [geotif]
+    - value: GeoJSON
+      mimetype: application/geo+json
+    - value: GeoPackage
```
wardi: I've tried to use the accepted file extensions as values in this list (e.g. `GPKG` for GeoPackage) and use the names of the formats as labels so they can be translated.

ckanext/canada/schemas/presets.yaml
```diff
@@ -3166,6 +3178,10 @@ presets:
     - value: RDF
       mimetype: application/rdf+xml
       openness_score: 4
+    - value: RDF/XML
+      mimetype: application/rdf+xml 
+    - value: RDF Turtle
```
wardi: Value should be `TTL`, use `label: RDF Turtle`

wardi:
Using all-uppercase versions of file extensions is a pattern we're inheriting from CKAN. We should avoid storing format names (use labels instead).

As a future improvement now that we're on a more recent version of CKAN we can also use CKAN's built in mapping of file extensions to mime-types (extending it as necessary) instead of having them embedded here.

ckanext/canada/schemas/presets.yaml
fanjinfei: since XML is also an existing entry, so is this RDF/XML still needed at all?

wardi: IIUC the RDF format above is the same as the RDF/XML you're adding here. XML is a more generic type

fanjinfei:
possible mapinfo file extension: *.TAB, *.MAP, *.DAT, *.ID,  *.IND.  Shall we include them all?
there is no specific extension for sql lite, could be .db, .db3, .SQLite, .sqlite3.  How to handle this?

wardi:
For MapInfo I'd choose `TAB` based on the description in http://www.gdal.org/drv_mitab.html

`SQLITE3` is good.

ckanext/canada/schemas/presets.yaml
```diff
@@ -3186,17 +3204,25 @@ presets:
...
+      value: SQLITE
     - value: SVG
       mimetype: image/svg+xml
       replaces: [svg]
+    - value: TAB
     - value: TIFF
       mimetype: image/tiff
       replaces: [tiff]
+    - value: TRiG
+    - value: TRiX
```
wardi:
```
TRIG
TRIX
```

ckanext/canada/schemas/presets.yaml
```diff
@@ -3186,17 +3204,25 @@ presets:
...
       mimetype: application/sql
       replaces: [sql]
+    - label: SQL Lite3
+      value: SQLITE3
+    - label: SQL Lite
+      value: SQLITE
     - value: SVG
       mimetype: image/svg+xml
       replaces: [svg]
+    - value: TAB
```
wardi: repeated from above

ckanext/canada/schemas/presets.yaml
```diff
@@ -3117,6 +3125,7 @@ presets:
     - value: JSON
       mimetype: application/json
       openness_score: 3
+    - value: JSON-LD
```
wardi: looks like the file extension has no hyphen. so `JSONLD`, optionally with a label.

PR #760, opened 2016-12-12
Change footer to be in line with the Drupal footer

LaurentGoderre:


wardi:
Is this missing updated translations?

PR #761, opened 2016-12-16
different GCPEDIA link according to resource type

fanjinfei:


ckanext/canada/templates/internal/recombinant/resource_edit.html
```diff
@@ -18,7 +18,12 @@
     <div class="panel panel-primary">
       <div class="panel-heading">{{ _("Guide") }}</div>
       <div class="panel-body">
-        <p>{% trans -%}For information on how to prepare, maintain and upload templates to the Registry, refer to the appropriate documentation posted on <a href="http://www.gcpedia.gc.ca/wiki/Open_Data_Inventory_/_l%E2%80%99inventaire_de_donn%C3%A9es_ouvertes">GCPEDIA</a>{% endtrans %}</p>
+        {% if resource['name'] == 'inventory' %}
+          {% set gcpedialink = "http://www.gcpedia.gc.ca/wiki/Open_Data_Inventory_/_l%E2%80%99inventaire_de_donn%C3%A9es_ouvertes" %}
+        {% else %}
+          {% set gcpedialink = _('http://www.gcpedia.gc.ca/wiki/Proactive_Disclosure_on_Open.Canada.ca') %}
+        {% endif %}
+        <p>{% trans gclink=gcpedialink %}For information on how to prepare, maintain and upload templates to the Registry, refer to the appropriate documentation posted on <a href="{{ gclink }}">GCPEDIA</a>{% endtrans %}</p>
```
wardi: instead of starting a if/elif/else chain here, put the override in a new key in the table definition at  https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/tables/inventory.yaml and use something like:

```
{% set gcpedialink = h.recombinant_get_geno(dataset.dataset_type).get(
    'gcpedia_link', 'http://www.gcpedia.gc.ca/wiki/Proactive_Disclosure_on_Open.Canada.ca') %}
```

PR #767, opened 2016-12-28
translation issues

fanjinfei:


ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
```diff
@@ -1727,3 +1739,5 @@ msgstr "Record pour %d ajoutée."
 msgid "Previous Year"
 msgstr "année précédente"
 
+msgid "Treasury Board Secretariat"
+msgstr "Secrétariat du Conseil du Trésor"
```
wardi: This looks like it belongs in the organization metadata, not here.

wardi:
@fanjinfei is this tested and good to go?

PR #769, opened 2017-01-06
open help link to other domain in new tab

fanjinfei:


LaurentGoderre:
Shouldn't the jinja templates handle this?

wardi:
I agree, let's be simple and explicit and add the `target` attribute to each link in the template.

PR #774, opened 2017-01-20
add is_required in url_only resource form

fanjinfei:


wardi:
is this broken in ckanext-scheming too? I don't see it there either.

fanjinfei:
PR 143 in ckanext-scheming

PR #778, opened 2017-01-27
translate html5 validation message and more

fanjinfei:


wardi:
@fanjinfei sorry I merged this prematurely. Would you fix the new user form to use attrs instead of the new parameter you added to the input macro?

PR #781, opened 2017-02-03
modify translation, help text and breadcrumb style

fanjinfei:


ckanext/canada/templates/internal/scheming/package/snippets/package_form.html
```diff
@@ -13,6 +13,11 @@
         <summary>
           {{_('Catalogue Metadata')}}
         </summary>
+        {% if dataset_type == 'info' %}
+          {% set m1 = _('Open Information Metadata Element Set') %}
+        {% else %}
+          {% set m1 = _('Open Data Metadata Element Set') %}
+        {% endif %}
```
wardi: this is ok, but in general I would prefer to put these strings in the schema (so we don't have if/else cases littering templates)

ckanext/canada/schemas/presets.yaml
```diff
@@ -1845,8 +1845,8 @@ presets:
   values:
     field_name: time_period_coverage_start
     help_text:
-      en: The start date of the period covered by the dataset
-      fr: Date de début de la période couverte par le produit (AAAA-MM-JJ)
+      en: The start date of the period covered by the dataset (YYYY-MM-DD)
+      fr: Date de début de la période couverte par le jeu de données (AAAA-MM-JJ)
```
wardi: I know the client wants the (YYY-MM-DD) text but that conflicts with modern browsers that render their own date fields so I tend to remove it

fanjinfei: Shall we remove this later?

PR #785, opened 2017-02-14
fix filter buttons for multi-select facet on organization search page

fanjinfei:


fanjinfei:
ready for review. This is together with PR https://github.com/open-data/ckan/pull/86

PR #789, opened 2017-02-15
Added the Maps Data Viewer

LaurentGoderre:


wardi:
@LaurentGoderre is this ready to merge?

LaurentGoderre:
Well...sort of....the endpoint are probably not the right ones and will need to change

chrismajewski:
Please do, the client needs to see the look and feel even if there is no data.

PR #794, opened 2017-02-16
add BOM support for csv2xlsx

fanjinfei:


fanjinfei:
ready for review

fanjinfei:
ready for review

bin/csv2xlsx.py
```diff
@@ -5,11 +5,20 @@
...
+import codecs
 import sys
 import openpyxl
 
+def peekBOM(f):
+    first3bytes = f.read(3)
+    if first3bytes == codecs.BOM_UTF8:
+        f.seek(3)
+    else:
+        f.seek(0)
```
wardi: can't call seek on a stream AFAIK, did you test this?

bin/csv2xlsx.py
```diff
@@ -5,11 +5,20 @@
...
+    return f
+
 def main():
-    reader = csv.reader(sys.stdin)
+    reader = csv.reader(peekBOM(sys.stdin))
```
wardi: instead of detecting the bom you could read one row and use `row[0].lstrip(codecs.BOM_UTF8)`

bin/csv2xlsx.py
fanjinfei: yes, I did test. It works.

bin/csv2xlsx.py
fanjinfei: Since seek() works, maybe leave as is?

wardi: sys.stdin.seek doesn't work in general, it seems to work only because we've redirected stdin to a file with `<` on the shell. Please use the change I suggested.

Also (minor Python style nit-picking) it's better not to return an object from a function that modifies that object. It makes it look like you're returning a modified version instead changing the thing passed. That's why e.g. list.sort() doesn't return the list, since the sorting is done in-place.

fanjinfei: thanks. I will change it.

wardi:
My favourite kind of PR: adds functionality and shrinks the code at the same time!

PR #795, opened 2017-02-17
add warning text for open data inventory upload page

fanjinfei:


fanjinfei:
@wardi ready for review

wardi:
You can use `{{ super() }}` to insert the previously-defined block content into a block, that's a lot better than copying it.

Is there a reason you can't use `upload_info` like ati, travelq and hospitalityq do?

fanjinfei:
I did change the html5 validation message for French.

wardi:
The language fix that would be better in recombinant

fanjinfei:
The tickets is asking to put the text on top of the submit button and using warning style, otherwise we can use upload_info.

PR #796, opened 2017-02-20
fix sort by title

fanjinfei:
need to rebuild index

### DIFF
--- a/ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
+++ b/ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
@@ -1760,6 +1760,12 @@ msgstr "En attente"
 msgid "Published"
 msgstr "Publié"
 
+msgid "title_string asc"
+msgstr "title_fr asc"
+
+msgid "title_string desc"
+msgstr "title_fr desc"
+
 msgid "login welcome text"
 msgstr ""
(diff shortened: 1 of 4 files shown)


fanjinfei:
@wardi ready for review

ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
```diff
@@ -1760,6 +1760,12 @@ msgstr "En attente"
 msgid "Published"
 msgstr "Publié"
 
+msgid "title_string asc"
+msgstr "title_fr asc"
+
+msgid "title_string desc"
+msgstr "title_fr desc"
```
wardi: it feels pretty strange to use gettext to translate a solr search parameter, maybe just use a if-else in the sort_by snippet instead.

ckanext/canada/plugins.py
```diff
@@ -402,6 +402,9 @@ def before_index(self, data_dict):
         if 'fgp_viewer' in data_dict.get('display_flags', []):
             data_dict['fgp_viewer'] = 'map_view'
 
+        titles = json.loads(data_dict.get('title_translated', '{}'))
+        data_dict['title_fr'] = titles.get('fr','')
+        data_dict['title'] = titles.get('en','')
```
wardi: french is going into `title_fr`, english is going into `title` but it looks like we're sorting english on `title_string`

PR #800, opened 2017-02-23
add default relevance sort_by option in package search page

fanjinfei:


fanjinfei:
ready for review

ckanext/canada/templates/internal/recombinant/snippets/xls_upload.html
```diff
@@ -15,7 +15,7 @@
   {% set geno = h.recombinant_get_geno(pkg.dataset_type) %}
   {% if 'upload_warn' in geno %}
     <div class="alert alert-danger">
-      <p>{{ _(geno.upload_warn) }}</p>
+      <p>{{ _(geno.upload_warn) | safe }}</p>
```
wardi: you want to insert HTML into the warning? That's ok, but seems unrelated to the PR

fanjinfei: So I use a separate PR?

wardi: that's probably better, and make the same change for upload_info (no reason for just one of them to support html)

ckanext/canada/templates/public/snippets/sort_by.html
```diff
@@ -11,14 +11,15 @@
 <div class="form-group">
   <label for="field-order-by" class="control-label">{{ _('Order by') }}</label>
   <select id="field-order-by" name="sort" class="form-control input-sm" onchange="this.form.submit()">
+    <option value="score desc,metadata_modified desc" >{{ _('Relevance') }}</option>
```
wardi: is there a reason you use `value="score desc,metadata_modified desc"` instead of `value=""` here?

fanjinfei: when there is no sort option, ckan will automatic use "score desc,metadata_modified desc" as default. That's why I am using it as a default value.

PR #801, opened 2017-02-23
fix help text display issue, translations

fanjinfei:


wardi:
A better commit message would be "allow HTML to be used in upload_warn messages". Would you make the same change for upload_info as part of this PR too?

fanjinfei:
I might add other changes to help text in the same PR. Should I change the message now?

wardi:
Sure, I'll wait for you to mark this ready for review

fanjinfei:
ready for review

ckanext/canada/tables/contracts.yaml
```diff
@@ -8,7 +8,7 @@ portal_type: info
 collection: pd
 
 resources:
-- title: Proactive Disclosure - Contracts
+- title: Proactive Disclosure - Contracts over 10K (Quarterly)
```
wardi: need to update the .po as well?

wardi: oh, I see this is the same as the title above, nevermind

ckanext/canada/templates/internal/recombinant/snippets/xls_upload.html
```diff
@@ -15,7 +15,7 @@
   {% set geno = h.recombinant_get_geno(pkg.dataset_type) %}
   {% if 'upload_warn' in geno %}
     <div class="alert alert-danger">
-      <p>{{ _(geno.upload_warn) }}</p>
+      <p>{{ _(geno.upload_warn) | safe }}</p>
```
wardi: please update `upload_info` above the same way, so we can add links there too

PR #802, opened 2017-03-01
add PD activity

fanjinfei:


fanjinfei:
In configuration file's plugin section, need to add canada_activity before datastore.  
need to add extendedactivity plugin.

ckanext/canada/plugins.py
```diff
@@ -437,3 +440,55 @@ def after_show(self, context, data_dict):
...
+                continue
+            actions = plugin.get_actions()
+            if 'datastore_upsert' in actions and (
+                    'datastore_delete' in actions):
+                old_upsert = actions['datastore_upsert']
+
+                def new_datastore_upsert(context, data_dict):
+                    act.datastore_activity_create(context, data_dict)
+                    return old_upsert(context, data_dict)
+                actions.update({'datastore_upsert': new_datastore_upsert})
```
wardi: Can we not just return our new implementations from this function instead of monkey-patching the actions in another plugin?

wardi: and does `get_action` not work instead of iterating over the plugins?

fanjinfei: We can not use the same name `datastore_upsert`, otherwise ckan will complain.  
https://github.com/open-data/ckan/blob/master/ckan/logic/__init__.py#L400

ckanext/canada/plugins.py
```diff
@@ -437,3 +440,55 @@ def after_show(self, context, data_dict):
...
+            if plugin == self:
+                continue
+            actions = plugin.get_actions()
+            if 'datastore_upsert' in actions and (
+                    'datastore_delete' in actions):
+                old_upsert = actions['datastore_upsert']
+
+                def new_datastore_upsert(context, data_dict):
+                    act.datastore_activity_create(context, data_dict)
+                    return old_upsert(context, data_dict)
```
wardi: The activity creation should be called after the original upsert/delete calls because those calls might raise a ValidationError.

As a smaller thing: wherever possible don't use (context, data_dict) as parameters to new functions. proper parameters are always easier to understand.

fanjinfei:
ready for review

fanjinfei:
not used

wardi:
tests are failing, would you add the extendedactivities extension to the test environment?

fanjinfei:
Since a related PR in github.com/ckan/ckanext-extendedactivity is not merged yet, should I point it to my fork so the test can pass?

wardi:
No, just make the change here after the extendedactivities one has been merged so the tests can run again

fanjinfei:
So I do not have to add -e git+https://github.com/ckan/ckanext-extendedactivity.git#egg=ckanext-extendedactivity in the requirements.txt? I am not sure if this extension is already included in the test environment.

ckanext/canada/plugins.py
```diff
@@ -440,3 +445,86 @@ def after_show(self, context, data_dict):
...
         return facet_titles
+
+class CanadaActivity(p.SingletonPlugin):
+    p.implements(p.IActions)
+    p.implements(IActivity)
+
+    def get_actions(self):
+        #override datastore_upsert and datastore_delete
+        for plugin in p.PluginImplementations(p.IActions):
+            if plugin == self:
```
TkTech: `if plugin is self` instead - you always want to do an identity test here.

fanjinfei:
need https://github.com/open-data/ckan/pull/93 to finish test.

wardi:
One more thing to try:
http://initd.org/psycopg/docs/connection.html#connection.notices

We can replace the connection.notices object with our own object to capture the number of records inserted or updated instead of using separate queries. This will likely be faster and will let us know when records were skipped because they weren't modified (for the new consultations type and others once we add the triggers for it)

wardi:
actually, this might not be possible until we do https://github.com/ckan/ckan/issues/3518

requirements.txt
```diff
@@ -7,5 +7,6 @@ ckantoolkit
 pytz
 babel
 -e git+https://github.com/ckan/ckanext-dcat.git#egg=ckanext-dcat
+-e git+https://github.com/open-data/ckanext-extendedactivity.git#egg=ckanext-extendedactivity
```
wardi: please remove this, we're deploying extendedactivity from the deploy scripts now so we can track its commit id.

ckanext/canada/plugins.py
```diff
@@ -440,3 +445,75 @@ def after_show(self, context, data_dict):
...
+                'datastore_delete': datastore_delete})
+
+    def string_icons(self, string_icons):
+        string_icons.update({'deleted datastore': 'file',
+                             'changed datastore': 'file'})
+
+    def snippet_functions(self, snippet_functions):
+        snippet_functions.update({'datastore': act.get_snippet_datastore,})
+        snippet_functions.update({'datastore_detail':
+                                  act.get_snippet_datastore_detail,})
```
wardi: looks a little odd. Do you mean to update both with one call? If not
```python
snippet_functions['datastore_detail'] = act.get_snippet_datastore_detail
```
is more idiomatic

ckanext/canada/activity.py
```diff
@@ -0,0 +1,77 @@
...
+            controller='ckanext.recombinant.controller:UploadController')
+        chromo = h.recombinant_get_chromo(resource_name)
+        return ''.join(['<a href="', url, '">', _(chromo['title']), '</a>'])
+    else:
+        return ''
+
+
+def get_snippet_datastore_detail(activity, detail):
+    count = activity['data']['count']
+    return ''.join([' ', str(count), ' ', _('entries')])
```
wardi: I think you want ngettext here.

ckanext/canada/activity.py
```diff
@@ -0,0 +1,77 @@
...
+    user = context['user']
+    model = context['model']
+    user_id = model.User.by_name(user.decode('utf8')).id
+    if activity_type == 'deleted datastore':
+        #  get last deleted activity for this user, if within 2 seconds,
+        #   merge these activities
+        act = model.activity.user_activity_list(user_id, 1, 0)
+        now = datetime.datetime.now()
+        if act and len(act)>0 and (
+                now - act[0].timestamp).total_seconds()<2 and (
```
wardi: use a constant defined at the top of the module like `MERGE_ACTIVITIES_WITHIN_SECONDS = 2`

fanjinfei:
@wardi test environment need to install ckanext-extendedactivity.

wardi:
yes, so add that here https://github.com/open-data/ckanext-canada/blob/master/bin/travis-install#L28

fanjinfei:
@wardi ready now.

PR #807, opened 2017-03-03
add PD data dictionary

fanjinfei:


fanjinfei:
ready for review

ckanext/canada/controller.py
```diff
@@ -262,6 +266,40 @@ def fgpv_vpgf(self, pkg_id):
...
+
+        blob = StringIO()
+        blob.write(codecs.BOM_UTF8)
+        out = unicodecsv.writer(blob)
+        for row in csv_dict:
+            out.writerow([unicode(col).encode('utf-8') for col in row])
+        contents = blob.getvalue()
+        blob.close()
+
+        response.headers['Content-Type'] = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
```
wardi: doesn't look like the right content-type for csv

ckanext/canada/controller.py
```diff
@@ -262,6 +266,40 @@ def fgpv_vpgf(self, pkg_id):
         return render('fgpv_vpgf/index.html', extra_vars={
             'pkg_id': pkg_id,
         })
+    def data_dictionary(self, pd_file):
+        names = pd_file.split('.')
+        if not (len(names)==2 and names[1]=='csv'):
+            return None
+        chromo = h.recombinant_get_chromo(names[0])
+        if not chromo:
+            return None
```
wardi: controllers should abort(404) instead of return None AFAIK.. I'm not even sure what `return None` does

ckanext/canada/controller.py
```diff
@@ -262,6 +266,40 @@ def fgpv_vpgf(self, pkg_id):
...
+        csv_dict += [ (f['datastore_id'], f['label'], _(f['label']))
+                for f in chromo['fields'] ]
+        set_lang(current_lang)
+
+        csv_dict.append(('owner_org', 'owner organization',
+                _('owner organization') ))
+        csv_dict.append(('owner_org_title', 'owner organization title',
+                _('owner organization title') ))
+
+        blob = StringIO()
```
wardi: There's no need for StringIO. We can .write straight to the pylons response object.

ckanext/canada/controller.py
```diff
@@ -262,6 +266,40 @@ def fgpv_vpgf(self, pkg_id):
...
+        if not chromo:
+            return None
+
+        current_lang= get_lang()
+        set_lang('fr')
+        csv_dict = [(u'Field Name / Nom de la zone',
+            u'English Description / Description en anglais',
+            u'French description / Description en français')]
+        csv_dict += [ (f['datastore_id'], f['label'], _(f['label']))
+                for f in chromo['fields'] ]
```
wardi: instead of `set_lang()` and `_()` use `recombinant_language_text(..., lang)` https://github.com/open-data/ckanext-recombinant/blob/master/ckanext/recombinant/helpers.py#L15. That helper will handle cases where we've added both languages directly in to the schema or where we're using gettext.

I've added a bunch of the other fields that TBS would like in the data dictionaries for the new consultations type. so you might want to add those columns too (I think they requested some of them in the ticket) https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/tables/consultations.yaml#L16-L25

ckanext/canada/controller.py
```diff
@@ -262,6 +266,40 @@ def fgpv_vpgf(self, pkg_id):
...
+            return None
+        chromo = h.recombinant_get_chromo(names[0])
+        if not chromo:
+            return None
+
+        current_lang= get_lang()
+        set_lang('fr')
+        csv_dict = [(u'Field Name / Nom de la zone',
+            u'English Description / Description en anglais',
+            u'French description / Description en français')]
```
wardi: instead of having one file with both languages we could generate the language the user requested then add two resources (one pointing to /en/data-dictionary and one pointing to /fr/data-dictionary) That would probably be easier for end users to deal with.

ckanext/canada/controller.py
fanjinfei: How to use the new consultations.yaml? Shall we append these fields to the .csv(such as contracts.csv) we generated? Also It is not loaded by the plugin yet. Do we need to add it to somewhere to get loaded?

wardi: This one isn't final yet, I'm just showing you an example of the new fields I added. We could start adding the same sorts of fields to the other types, but that could happen in a separate PR

fanjinfei: thanks.

ckanext/canada/controller.py
```diff
@@ -263,6 +265,30 @@ def fgpv_vpgf(self, pkg_id):
             'pkg_id': pkg_id,
         })
 
+    def data_dictionary(self, pd_file):
+        names = pd_file.split('.')
+        if not (len(names)==2 and names[1]=='csv'):
```
wardi: pep8: there should be spaces around `==`

Python also has str.partition() instead of split to always split once and not raise an error, e.g.

```python
name, dot, ext = pd_file.partition('.')
if ext != 'csv':
   ...
```

wardi:
I've built this on the similar code in ckanext-recombinant instead, this way improvements to our reference pages will be immediately available to the public data dictionaries too.

PR #811, opened 2017-03-10
fix English text for relate item

fanjinfei:


fanjinfei:
ready for review

wardi:
merged your original PR instead

PR #815, opened 2017-03-16
add pagination for comments

fanjinfei:


fanjinfei:
ready for review

ckanext/canada/templates/public/package/comments.html
```diff
@@ -1,9 +1,27 @@
...
+        <div class="col-md-4"><h4 style="display:inline-block;">{{ _('Comments') }}({{ c.page.item_count }} )</h4></div>
+
+        {% block page_pagination_top %}
+          <div class="col-md-4"><small>{{ c.page.pager(q=c.q) }}</small></div>
+         {% endblock %}
+
+        <div class="text-right"><small>
+          <form class="form-inline mrgn-bttm-sm mrgn-tp-md form-group form-horizontal" role="form" method="get" >
+            <select id="field-page-limit" name="pagelimit" class="form-control input-sm" onchange="this.form.submit()">
+              <option value="20" {% if c.pagelimit==20 %} selected="selected"{% endif %}>20</option>
```
TkTech: Loops work like you'd expect in Jinja, so do this instead of each one by hand. Will make it easier if we have to change the options or if we turn it into a setting.
```
{% for value in [20, 50, 100] %}
{% endfor %}
```

ckanext/canada/wcms.py
```diff
@@ -101,6 +110,21 @@ def wcms_dataset_comments(pkg_id, lang):
...
 
     try:
+        params_nopage = [(k, v) for k, v in request.params.items()
+                         if k != 'page']
+        def pager_url(q=None, page=None):
+            params = list(params_nopage)
+            params.append(('page', page))
+            return search_url(params, pkg_id)
+
+        page = 1 #self._get_page_number(request.params)
```
TkTech:
```
page = int(request.params.get('page', 1))
limit = min(100, int(request.params.get('limit', 50)))
```

ckanext/canada/wcms.py
```diff
@@ -109,7 +133,9 @@ def wcms_dataset_comments(pkg_id, lang):
                 ct.pkg_id == bindparam('pkg_id'),
                 ct.language == bindparam('language')
             ),
-            order_by=[_drupal_db.drupal_comments_table.c.thread]
+            limit = limit,
+            offset = (page -1)*limit,
```
TkTech: `(page - 1) * limit` (missing spaces around operators, extra spaces around limit/offset. Please run a linter like pep8/pyflakes/etc...!

fanjinfei:
ready for review

PR #816, opened 2017-03-17
support sort for resources and related items tables on package_read page

fanjinfei:


fanjinfei:
ready for review

ckanext/canada/templates/public/package/read.html
```diff
@@ -312,6 +312,14 @@
     });
   }
 });
+
+</script>
+<script src="https://cdn.datatables.net/1.10.13/js/jquery.dataTables.min.js"></script>
```
wardi: we're not supposed to be using the cdn for the plain version of datatables. instead enable the one already included with WET: https://wet-boew.github.io/v4.0-ci/demos/tables/tables-en.html

PR #817, opened 2017-03-23
add schema.org markup

fanjinfei:


fanjinfei:
ready for review

ckanext/canada/templates/public/package/snippets/schemaorg.html
```diff
@@ -0,0 +1,91 @@
...
+            {%- for resource in data.resources -%}
+              {%- if not resource.related_type -%}
+                {%- do nonapps.append(resource) -%}
+              {%- endif -%}
+            {%- endfor -%}
+            {%- if nonapps -%}
+              {% for res in nonapps %}
+                <span property="distribution" typeof="DataDownload">
+                    <span property="inLanguage" content="{{ ','.join(res[lang_field.field_name]) }}"> </span>
+                    <span property="name" lang="{{ h.lang() }}">{{ h.scheming_language_text(res[name_field.field_name]) }}</span>
```
wardi: might as well use `h.get_translated(res, 'name')` instead of scheming_field_by_name and scheming_language text. That's what the core ckan and the other templates do.

ckanext/canada/templates/public/package/snippets/schemaorg.html
```diff
@@ -0,0 +1,91 @@
...
+            {%- set nonapps = [] -%}
+            {%- for resource in data.resources -%}
+              {%- if not resource.related_type -%}
+                {%- do nonapps.append(resource) -%}
+              {%- endif -%}
+            {%- endfor -%}
+            {%- if nonapps -%}
+              {% for res in nonapps %}
+                <span property="distribution" typeof="DataDownload">
+                    <span property="inLanguage" content="{{ ','.join(res[lang_field.field_name]) }}"> </span>
```
wardi: there's no need for scheming_field_by_name if all you need is the name. Use `res.language` instead

ckanext/canada/templates/public/package/snippets/schemaorg.html
```diff
@@ -0,0 +1,91 @@
+{% import 'macros/canada_read.html' as cr %}
+
+    <div vocab="http://schema.org/" typeof="DataSet" style="display:none;">
+        <div property="includedInDataCatalog" typeof="DataCatalog">
+            <span property="name" value="{{ _('Government of Canada Open Government Portal') }}">
+            {%- if data.get('collection')== 'geogratis' or data.get('collection')== 'fgp' -%}
```
wardi: it's possible for normal datasets to fill in the spatial field too. How about `if data.get('spatial')` instead

ckanext/canada/helpers.py
```diff
@@ -58,6 +59,12 @@ def user_organizations(user):
     u = User.get(user['name'])
     return u.get_groups(group_type = "organization")
 
+def catalogue_last_update_date():
+    q = model.Session.query(Activity.timestamp).filter(
+        Activity.activity_type.endswith('package')).order_by(
+        Activity.timestamp.desc()).limit(1).one()
```
wardi: instead of `.limit(1).one()`, which can raise an exception, how about `.first()` and return an empty string when the response `is None`. Helpers shouldn't fail because Jinja has no exception handling so the common pattern is to return an empty string instead.

ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
```diff
@@ -1858,3 +1858,21 @@ msgstr "Ce dossier a été supprimé de %s."
...
 msgstr "Dossier supprimé"
+
+msgid "Government of Canada Open Government Portal"
+msgstr "Gouvernement du Canada site gouvernemental ouvert"
+
+msgid "This catalog contains metadata records describing open datasets available from the Government of Canada"
+msgstr "Ce catalogue contient des enregistrements de métadonnées décrivant des ensembles de données ouverts disponibles auprès du gouvernement du Canada"
+
+msgid "en-CA"
+msgstr "fr-CA"
```
wardi: This one doesn't make sense. `fr-CA` isn't a translation of `en-CA`.

ckanext/canada/templates/public/package/snippets/schemaorg.html
```diff
@@ -0,0 +1,91 @@
...
+                  {%- endif -%}
+                    </span>
+                    </span>
+            {%- endif -%}
+            </span>
+
+            <span property="description" value="{{ _('This catalog contains metadata records describing open datasets available from the Government of Canada') }}"></span>
+            <span property="datePublished" value="2013-06-19"></span>
+            <span property="dateModified" value="{{ h.catalogue_last_update_date() }}"></span>
+            <span property="inLanguage" value="{{ _('en-CA') }}"></span>
```
wardi: you can use `value="{{ h.lang() }}-CA"`

ckanext/canada/templates/public/package/snippets/schemaorg.html
fanjinfei: some 'geogratis' data's 'spatial' field is empty. Maybe we add `data.get('spatial')` as another `or`?

wardi: there's no content if the spatial field is empty or missing. I'd just use `spatial` as the test and ignore the other fields.

thriuin:
@fanjinfei Could you resolve the merge conflict in the .po file and let me know when this PR is ready to merge? Thanks

PR #818, opened 2017-03-23
Removing FGP configuration files from this code repository

thriuin:


wardi:
I'm all for removing stuff that we don't need!

PR #823, opened 2017-03-30
wrap resource name

fanjinfei:


fanjinfei:
ready for review

ckanext/canada/validators.py
```diff
@@ -110,6 +110,17 @@ def canada_validate_generate_uuid(value):
         raise Invalid(_("Badly formed hexadecimal UUID string"))
 
 
+email_pattern = re.compile(r"^[A-Za-z0-9\.\+_-]+@[A-Za-z0-9\._-]+\.[a-zA-Z]*$")
```
wardi: I suggest https://html.spec.whatwg.org/#e-mail-state-(type=email) (the regexp is below) and add a comment that links to this doc to say where we got it from.

ckanext/canada/validators.py
```diff
@@ -110,6 +110,17 @@ def canada_validate_generate_uuid(value):
...
+def email_validator(value):
+    if value:
+        try:
+            if not email_pattern.match(value):
+                raise Invalid(_('Please enter a valid email address.'))
+        except:
```
wardi: bare `except:`s make @TkTech :anger:

What exception is this catching? TypeError?

ckanext/canada/schemas/presets.yaml
```diff
@@ -3078,6 +3078,8 @@ presets:
     - value: EXE
       mimetype: application/vnd.microsoft.portable-executable
       replaces: [exe]
+    - value: FEFLOW
+      replaces: [feflow]
```
wardi: replaces is just for formats that were in the previous version of the metadata standard. It doesn't apply to any new types

ckanext/canada/schemas/presets.yaml
```diff
@@ -3142,10 +3148,20 @@ presets:
...
     - value: LYR
+    - value: Leapfrog Mining
+      replaces: [leapfrog mining]
+    - value: Leapfrog Hydro
+      replaces: [leapfrog hydro]
+    - value: Leapfrog Viewer
+      replaces: [leapfrog viewer]
     - label: MapInfo
       value: TAB
+    - value: MapInfo Grid
```
wardi: we use the most common extension as the value for these choices (e.g. see TAB above for MapInfo)

wardi: Might need to go back to the client and ask for the list of file extensions along with these format names

wardi:
@fanjinfei please remove the format changes from this PR so that the rest can be merged while we wait.

PR #825, opened 2017-04-04
add alternative notes for open dialogue

fanjinfei:


fanjinfei:
ready for review. also https://github.com/open-data/ckanext-recombinant/pull/83

wardi:
How about using the translation files to replace the visible text on the search page (save us updating all the datasets for a temporary thing) and add the message to upload_info or upload_warn instead of making recombinant changes?

PR #827, opened 2017-04-05
Add quick and dirty script for modifying analytics file from TBS

thriuin:


thriuin:
Will merge since the only change is to add a utility script to the bin directory.

PR #830, opened 2017-04-07
Updated README with new DDL for opendata_package_count_v

thriuin:


thriuin:
OK, this is tweak to the readme - I'm just mergin' it.

PR #840, opened 2017-04-20
display comments in all languages

fanjinfei:


fanjinfei:
ready for review

ckanext/canada/helpers.py
```diff
@@ -118,7 +118,7 @@ def dataset_rating(package_id):
     return wcms_dataset_rating(package_id)
 
 
-def dataset_comment_count(package_id, lang):
```
wardi: better to remove the lang parameter from the helpers and wcms_dataset_comment functions

PR #844, opened 2017-04-21
convert page item_count to integer

fanjinfei:


fanjinfei:
It causes paginator exception if it is decimal.

fanjinfei:
@wardi ready for review.

ckanext/canada/wcms.py
```diff
@@ -304,4 +304,4 @@ def wcms_dataset_comment_count(package_id):
         # upstream and see if Sentry can tell us why.
         logging.exception('KeyError occured while pulling comment count.')
 
-    return count
+    return int(count)
```
wardi: you can be extra paranoid here. we *really* don't want changes in the drupal db to knock us down
```python
try:
    return int(count)
except ValueError:
    return 0
```

PR #848, opened 2017-04-25
keep url's file basename for cloud resources

fanjinfei:


ckanext/canada/commands.py
```diff
@@ -499,8 +499,8 @@ def _trim_package(pkg):
                 'last_modified', 'position']:
             if k in r:
                 del r[k]
-        if r.get('url_type') == 'upload':
-            r['url'] = 'upload'
```
wardi: os.path isn't the perfect thing to use here because the separator will depend on the operating system. rsplit('/') is fine, or use one of the url splitting functions instead

wardi:
@fanjinfei thank you!

PR #856, opened 2017-05-04
fix pd to solr

fanjinfei:


ckanext/canada/pd.py
```diff
@@ -148,7 +148,8 @@ def rebuild(command_name, csv_files=None):
             geno = get_geno(chromo['dataset_type'])
 
             for org_id, records in csv_data_batch(csv_file, chromo):
-                records = [dict((k, safe_for_solr(v)) for k, v in
+                records = [dict((k, safe_for_solr(v))
+                            if v is not None else (k, u'') for k, v in
```
wardi: maybe add the is None check to `safe_for_solr instead?

PR #859, opened 2017-05-08
organization activity feeds

fanjinfei:


ckanext/canada/controller.py
```diff
@@ -458,11 +458,18 @@ def output_feed(self, results, feed_title, feed_description,
...
         )
 
+        chromos = h.recombinant_get_types()
+
         for pkg in results:
+            title, notes = None, None
+            if pkg['type'] in chromos:
+                chromo = h.recombinant_get_chromo(pkg['type'])
+                title= _(chromo['title'])
+                notes = _(pkg['notes'])
```
wardi: these should have been copied onto the datasets, why is this necessary?

fanjinfei: It is not there. See http://registry.open.canada.ca/en/api/2/rest/package/c3d6f6a9-9558-4376-9237-4573aee5f835

fanjinfei: we translate it here https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/templates/public/snippets/package_item.html#L10

wardi: they get updated by running `paster recombinant create <type>` after the description changes. Hacking it into the templates is something I'd like to avoid

wardi: also it's better to ignore the rest API, it's going away in a future CKAN

fanjinfei: maybe we should add `title_translated` here at https://github.com/open-data/ckanext-recombinant/blob/master/ckanext/recombinant/logic.py#L166 ?

wardi: no we keep our title translations in po files so `paster recombinant create <type>` only needs to update the title field (as it does now)

fanjinfei: With `paster recombinant create -f`, still the title has only English in solr. http://registry.open.canada.ca/api/action/package_show?id=c3d6f6a9-9558-4376-9237-4573aee5f835

wardi: right and `h.get_translated` should use gettext to find the title for display because there is no title_translated field

fanjinfei: So we will fix the h.get_translated() in ckan?

wardi:
@fanjinfei is this complete?

fanjinfei:
need the ckan PR: https://github.com/open-data/ckan/pull/99

wardi:
I can't see a reason that it will be rejected. I'll merge that into ours

PR #865, opened 2017-05-17
remove repetition of search condition

fanjinfei:


wardi:
@thriuin I don't regret merging this 2 hours ago. Is that what you meant?

thriuin:
@wardi smart@ss :-)

PR #866, opened 2017-05-17
check resource link in datasets

fanjinfei:


wardi:
This looks great @fanjinfei 
How long does it take to run? Do you have example output?

fanjinfei:
I leave it run for overnight. Due to poor network on my desktop (link disconnected quick often), I re-run next morning a few times (5 minutes each run) then dump it. Our links count is 312516, broken count is 756.

PR #872, opened 2017-05-25
Changes to ODB schema for version 3 from TBS/OGS

thriuin:
A couple of fields dropped, and some new ones added. We are still missing a number of selection lists which have not yet been provided by the client


thriuin:
@wardi  I'm going to merge for now but we can review next week.

PR #874, opened 2017-05-26
update registry csv file of broken link

fanjinfei:


bin/link_check.py
```diff
@@ -39,14 +43,15 @@ def test_ftp(url):
...
+        response = urlopen(req, timeout=URL_TIMEOUT)
         chunk = response.read(16)
         if len(chunk) == 16:
             res.status_code = 200
         else:
             res.status_code = 404
     except:
-        res.status_code = 404
+        print('ftp exception', url)
+        return Exception()
```
wardi: bare except clauses are a bad idea: they will catch much more than you expect (like pressing CTRL+C while the script is running). returning an empty exception object is also very strange. is there another way you can think of doing this?

fanjinfei: There are a few possible reasons to cause exception such as timeout, NIC link down, proxy rejection, proxy timeout, etc. I am not sure if I can list all of them. For those exceptions, we still can not tell if the link is bad or not ( it might be good from outside network), hence pointless to go further to distinguish them. It might be good to just stop at this level and leave it for further manual check.

wardi: Ok, then at least use `except Exeption as e:` to avoid the KeyboardInterrupt one and how about returning the actual exception instead of a bare one in case the caller wants to do more with the actual error?

PR #875, opened 2017-05-29
obd samples

fanjinfei:


wardi:
this looks great. would you mind adding the french files as additional resources?

fanjinfei:
no problem.

wardi:
in the resource metadata the formats should be set to PDF and the language should be `["en"]` for one and `["fr"]` for the other.

PR #879, opened 2017-06-05
update OBD package search facets

fanjinfei:


README.rst
```diff
@@ -131,6 +131,7 @@ We use a different list of plugins for Open By Default::
         canada_obd canada_package wet_boew_gcweb scheming_datasets
         fluent cloudstorage
 
+   ckan.extra_resource_fields = language
```
wardi: I forgot to ask: Can we apply this setting from our plugin so it's not an extra manual step?

fanjinfei: https://github.com/open-data/ckanext-canada/pull/880/commits/485d3597f6999f377a20a72bf48046f06d2e41cc

PR #883, opened 2017-06-08
sort facet items by locale

fanjinfei:


ckanext/canada/templates/public/snippets/facet_list.html
```diff
@@ -8,14 +8,15 @@
   {% set item_count = items | length %}
   {% set items = items[0:5] if ( not unlimit) and h.get_param_int('_%s_limit' % name) else items %}
 
+  {% for item in items %}
+    {% do item.update({'label': h.scheming_choices_label(scheming_choices, item.name)
+      if scheming_choices else cr.split_bilingual_field(item.display_name, h.lang()) }) %}
+  {% endfor %}
+  {% set items = h.sorted_by_locale(items, locale='en_US', attribute='label') %}
```
wardi: unconditionally sorting by en_US locale seems odd, is that because the sort order will be the same for both en/fr?

If this is just for sorting strings with accents we could call `unicodedata.normalize('NFKD', label)` before sorting and skip the pyicu dependency, right?

fanjinfei: The ticket is about sorting accents. I will close this and use unicodedata.normalize().

PR #884, opened 2017-06-12
fix uncaught exception in link_check, pthread_cancel error

fanjinfei:


bin/link_check.py
```diff
@@ -165,11 +165,17 @@ def test_links(self, new_url, urls):
...
-                    if response.status_code != requests.codes.ok:
-                        res['resources'] = new_url[url]
-                        res['org'] = urls.get(url, None)
+                    try:
+                        res={'timestamp': now,
+                             'status':response.status_code}
+                        if response.status_code != requests.codes.ok:
+                            res['resources'] = new_url[url]
+                            res['org'] = urls.get(url, None)
+                    except:
```
wardi: please add an appropriate exception class here

bin/update_broken_link.sh
```diff
@@ -7,15 +7,16 @@ wget -O /tmp/od-do-canada.jl.gz http://open.canada.ca/static/od-do-canada.jl.gz
 
 rm -rf /tmp/od_linkcheker2.db
 TMPDIR="/home/odatsrv/tmp"
-(time python3 /home/odatsrv/bin/link_check.py --file=/tmp/od-do-canada.jl.gz ) > "${TMPDIR}/link_check.log" 2>&1
+export LD_PRELOAD=/lib64/libgcc_s.so.1
+(time /srv/python/bin/python3 /home/odatsrv/bin/link_check.py --file=/tmp/od-do-canada.jl.gz ) > "${TMPDIR}/link_check.log" 2>&1
 sleep 30
-(time python3 /home/odatsrv/bin/link_check.py --file=/tmp/od-do-canada.jl.gz --quick ) > "${TMPDIR}/link_check.log2" 2>&1
+(time /srv/python/bin/python3 /home/odatsrv/bin/link_check.py --file=/tmp/od-do-canada.jl.gz --quick ) > "${TMPDIR}/link_check.log2" 2>&1
```
wardi: are the retries here in case of failure? If so can you check the error code and only re-run when it's non-zero?

bin/link_check.py
fanjinfei: I can not remember the exact error. It is not an exception type. I believe it is from the test_ftp(), since last commit.

bin/update_broken_link.sh
fanjinfei: some timeout urls will be ok if we re-visit them. I am not sure if it is due to our network or theirs.

bin/link_check.py
wardi: it's not an Exception subclass? Then I'd really like to know what it is!  We don't use bare `except`s (unless you're also using a bare `raise`) because they catch things they shouldn't.

bin/update_broken_link.sh
wardi: oh so it's not an error, just some urls time out. So you combine the report from all runs to filter out the intermittent issues. Sounds good

PR #886, opened 2017-06-14
convert OBD xml to json

fanjinfei:


bin/import_xml2obd.py
```diff
@@ -0,0 +1,179 @@
...
+            return it['values']
+    return None
+
+
+def main():
+    global presets, audience, canada_resource_type,canada_subject
+    global canada_resource_language
+    refs, extras = read_xml(sys.argv[1])
+    read_organizations_from_json(sys.argv[2])
+    presets = read_presets(sys.argv[3])
```
wardi: slightly better than this would be calling the API for `scheming_dataset_schema_show(type='doc')` because some of our fields might get overridden. Not worth the change for now, just wanted to make a note.

wardi:
Looks great @fanjinfei

PR #888, opened 2017-06-19
generate openness report and deleted datasets, import obd docs

fanjinfei:


bin/openness_report.py
```diff
@@ -0,0 +1,138 @@
...
+                                    self.fmt))
+                scores.append(1)
+                score = max(scores)
+
+                for r in record['resources']:
+                    if 'data_includes_uris' in r.get('data_quality', []):
+                        score = max(4, score)
+                        if 'data_includes_links' in r.get('data_quality', []):
+                            score = max(5, score)
+                report[score] += 1
```
wardi: how about importing and calling the original helper instead of duplicating its logic here?

bin/openness_report.py
```diff
@@ -0,0 +1,138 @@
...
+    def __delete__(self):
+        if not self.file:
+            if self.download_file:
+                os.unlink(self.download_file)
+                print('temp file deleted', self.download_file)
+
+    def download(self):
+        if not self.file:
+            # dataset http://open.canada.ca/data/en/dataset/c4c5c7f1-bfa6-4ff6-b4a0-c164cb2060f7
+            url='http://open.canada.ca/static/od-do-canada.jl.gz'
```
wardi: would be better to get this url from the resource because we're moving it to azure soon

bin/openness_report.py
```diff
@@ -0,0 +1,138 @@
...
+            r = requests.get(url, stream=True)
+
+            f = tempfile.NamedTemporaryFile(delete=False)
+            for chunk in r.iter_content(1024 * 64):
+                    f.write(chunk)
+            f.close()
```
wardi: Is there a reason you're downloading the file? you should be able to stream read and yield the records from the request (no temp files to clean up that way)

bin/openness_report.py
wardi: if it's a python3 import issue we could move the logic to a module that is importable from either python2 or 3: this could be the start of a good thing :-)

bin/import_xml2obd.py
```diff
@@ -157,21 +185,105 @@ def get_preset(name):
...
+    if not ds.get('keywords', None):
+        ds['keywords']={'en':['statcan'], 'fr':['statcan']}
+    if not ds.get('subject', None):
+        ds['subject'] = ["information_and_communications"]
+    if not ds.get('notes_translated', None):
+        ds['notes_translated'] = {'en':'n/a', 'fr':'n/a'}
+    if not ds['notes_translated'].get('fr', None):
+        ds['notes_translated']['fr'] = ds['notes_translated']['en']
+    if not ds['title_translated'].get('fr', None):
+        ds['title_translated']['fr'] = ds['title_translated']['en']
```
wardi: I'm going to make the notes optional for this type instead. I don't like filling in "n/a" unless we know that something is actually not applicable

bin/import_xml2obd.py
```diff
@@ -157,21 +185,105 @@ def get_preset(name):
...
 
     res = xml_obd_mapping(refs, xml2resource)
+    res_name = sys.argv[1].split('/')[-1][:-4]
+    res['name_translated']={'en':res_name, 'fr':res_name}
+    res['id'] = str(uuid.uuid5(uuid.NAMESPACE_URL,
+                    'http://dev.obd.ca/resources/' + sys.argv[1]))
+    res['url'] = 'http://dev.obd.ca/' + sys.argv[1].split('/')[-1][:-4]
```
wardi: obd.ca is some domain for sale, please use something like obd.open.canada.ca instead (I know these won't be visible in the resulting datasets, I just don't want them in our source code)

bin/import_xml2obd.py
```diff
@@ -157,21 +185,105 @@ def get_preset(name):
...
 
     ds = xml_obd_mapping(refs, xml2obd)
     ds['collection'] = 'publication'
+    release_date = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    if not ds.get('date_published'):
+        ds['date_published'] = release_date
+    ds['id'] = str(uuid.uuid5(uuid.NAMESPACE_URL,
+                   'http://dev.obd.ca/' + sys.argv[1]))
+    ds['name'] = ds['id']
+    ds['title'] = ds['title_translated']['en']
```
wardi: don't set this, our validator fills in this field for us

bin/import_xml2obd.py
```diff
@@ -157,21 +185,105 @@ def get_preset(name):
...
+    canada_resource_format = None
+
+    # upload the generated json to remote site and upload the file
+    if sys.argv[1] =='upload':
+        return upload_resources(*sys.argv[2:])
+
     refs, extras = read_xml(sys.argv[1])
     read_organizations_from_json(sys.argv[2])
-    presets = read_presets(sys.argv[3])
+    presets = read_presets(sys.argv[3])  # TODO: read from doc.yaml
```
wardi: `site.action.scheming_dataset_schema_show(type='doc')` is all you should need here

bin/import_xml2obd.py
```diff
@@ -157,21 +185,105 @@ def get_preset(name):
...
     audience = get_preset('canada_audience')
     canada_resource_type = get_preset('canada_resource_type')
     canada_subject = get_preset('canada_subject')
     canada_resource_language = get_preset('canada_resource_language')
+    canada_resource_format = map(lambda x:x['value'],
+                                 get_preset('canada_resource_format')['choices'])
```
wardi:
```python
[x['value'] for x in get_preset('canada_resource_format')['choices']]
```
is faster and more idiomatic

bin/openness_report.py
```diff
@@ -0,0 +1,138 @@
...
+        self.download_file = None
+        with open(fmt_file, 'r') as f:
+            presets = yaml.load(f)
+        fmt = None
+        for it in presets['presets']:
+            if it['preset_name']=='canada_resource_format':
+                fmt = it['values']['choices']
+        if not fmt:
+            raise Exception('no resource format loaded')
+        self.fmt = filter(lambda x: 'openness_score' in x, fmt)
```
wardi:
```python
[x for x in fmt if 'openness_score' in x]
```

PR #890, opened 2017-06-27
OBD translations

fanjinfei:


ckanext/canada/templates/obd/snippets/search_form.html
```diff
@@ -0,0 +1,56 @@
+{% ckan_extends %}
+{% import 'macros/canada_read.html' as cr %}
+
+{% block search_input_label %}
+
+<div class="col-md-12">
+  <section class="alert alert-info" >
+    {% if h.lang() =='fr' %}
+    <p>L’initiative « Ouvert par défaut » donne accès aux documents de travail utilisés par le gouvernement du Canada. La plupart de ces renseignements ne sont pas publiés. Les documents peuvent être sous la forme d’ébauche et sont en général disponibles dans la langue dans laquelle ils ont été créés. À moins d’indication contraire, les renseignements sont disponibles en vertu de la <a href="http://ouvert.canada.ca/fr/licence-du-gouvernement-ouvert-canada">licence du gouvernement ouvert</a>.</p>
```
wardi: please use _() here instead

ckanext/canada/templates/obd/snippets/search_form.html
```diff
@@ -0,0 +1,56 @@
...
+	<a id="OGSCartListItems" class="ogscartlistbtn btn btn-primary btn-xs" href="#"><span class="fa fa-list-alt"></span> {{ _('List Cart Items') }}</a>
+	<a id="OGSCartPlotItems" class="ogscartplotbtn btn btn-primary btn-xs" href="#"><span class="fa fa-globe"></span> {{ _('View on Map') }}</a>
+	<a id="OGSCartDumpItems" class="ogscartplotbtn btn btn-danger btn-xs" href="#"><span class="fa fa-trash-o"></span> {{ _('Empty Cart') }}</a>
+</div>
+
+{% endblock %}
+
+
+
+{% block filter_list %}
```
wardi: this block isn't changed at all. you can omit it here.

ckanext/canada/templates/obd/snippets/search_form.html
wardi: maybe add a block to public/snippets/search_form so we don't need to copy the whole form to only override the alert that is displayed

PR #893, opened 2017-07-04
check md5 of resource files

fanjinfei:


bin/import_xml2obd.py
```diff
@@ -41,6 +43,15 @@
...
 
 
+def md5str(fname):
+    return hashlib.md5(open(fname, 'rb').read()).hexdigest().strip()
+
+def base64md5str(val):
+    if not val:
+        return None
+    a = base64.b64decode(val)
+    return ''.join([ '%02x'%(ord(c)) for c in a]).strip()
```
wardi: hexlify is the function you want here

bin/import_xml2obd.py
```diff
@@ -236,16 +302,30 @@ def upload_resources(remote_site, api_key, jsonfile, resource_directory):
...
+            if objmd5 == srcmd5:
+               print('\tsame remote file exists ',  os.path.basename(source))
+               skip = True
+            else:
+                print (objmd5, srcmd5)
+        if not skip:
+            with open(source) as f:
+                try:
+                    rc = site.action.resource_patch(
+                        id=res['id'],
```
wardi: add `hash='md5-%s' % obj.properties.content_settings.content_md5,` here to store the hash in the resource too

bin/import_xml2obd.py
```diff
@@ -236,16 +302,38 @@ def upload_resources(remote_site, api_key, jsonfile, resource_directory):
...
+                    skip = True
+                    break
+        if skip:
+            print('\tsame remote file exists ' + os.path.basename(source))
+        else:
+            with open(source) as f:
+                try:
+                    rc = site.action.resource_patch(
+                        id=res['id'],
+                        hash='md5-%s'%srcmd5,
```
wardi: use the base64-encoded hash here, I'd like to follow the SRI standard for hashing

bin/import_xml2obd.py
```diff
@@ -340,20 +397,35 @@ def pull_docs(conf_file, local_dir):
...
     docs = [x[:-4] for x in xmls]
     inds = [x + '.ind' for x in docs]
     files = xmls + docs
     for fname in files:
         print 'Downloading ',fname
-        src.download_blob(fname, local_dir + '/' + fname.split('/')[-1])
+        obj = src.get_obj(fname)
+        localname = local_dir + '/' + fname.split('/')[-1]
+        objmd5 = base64md5str(obj.properties.content_settings.content_md5)
+        if os.path.isfile(localname) and objmd5 == md5str(localname):
```
wardi: how about storing the files using the hash as the filename so you don't need to compute it again each time?

bin/import_xml2obd.py
```diff
@@ -41,6 +44,20 @@
 canada_resource_format = None
 
 
+def md5str(fname):
+    return hashlib.md5(open(fname, 'rb').read()).hexdigest().strip()
+
+def md5_file(fname):
+    return hashlib.md5(open(fname, 'rb').read()).digest()
```
wardi: *if* you need to compute the hash yourself it's better to do it in chunks of a few MB at a time instead of by loading the whole file into memory. Even better is not needing to compute hash values.

bin/import_xml2obd.py
fanjinfei: It might be more readable. I could save (filename, md5) pairs in a separate local file under the same local directory.

wardi: sure, or have filename.hash files in the same directory. you don't need to know all the hashes at once, just the one of the file you're checking

bin/import_xml2obd.py
```diff
@@ -236,16 +346,45 @@ def upload_resources(remote_site, api_key, jsonfile, resource_directory):
...
+                    skip = True
+                    break
+        if skip:
+            print('\tsame remote file exists ' + os.path.basename(source))
+        else:
+            with open(source) as f:
+                try:
+                    rc = site.action.resource_patch(
+                        id=res['id'],
+                        hash='md5-%s' % base64.b64encode(md5),
```
wardi: One last change: client requested that users should be able to verify that the files they receive are the right files and md5 is a very weak assurance, especially with binary files like PDFs. While looping over the chunks of the files and calculating the md5, please also calculate the SHA384 and store it in this field like `hash='md5-%s sha384-%s' % (b64encode(md5), b64encode(sha384))`

PR #895, opened 2017-07-05
convert rating to integer in template

fanjinfei:


wardi:
should we be rounding instead of truncating? (`rating + 0.5 | int`) do we know how drupal displays their rating?

PR #899, opened 2017-07-17
insert loop11 js on all OG portal pages

fanjinfei:


wardi:
Let's do this as a conditional block like the GA one so that the key can be put in our config file and so that we can turn it on or off on a per-site basis

PR #904, opened 2017-07-25
handle network glitch on pd solr rebuild

fanjinfei:


wardi:
Do the wget -t and --waitretry options not do what we need here?

fanjinfei:
wget will exit on 503 .  It has --retry-on-httperror option on newer version 1.19.

PR #914, opened 2017-11-22
Quick and dirty fix to hardcoded Drupal CSS file

thriuin:
Quick and dirty fix to hardcoded Drupal CSS file in package read temp…late - needs cleanup
TODO: copy relevant portions of CSS to Canada.css and delete this file


wardi:
this should be against registry-private branch

PR #915, opened 2017-11-22
Temporary fix for a harded reference to a Drupal CSS file

thriuin:
TODO: remove duplicated CSS file and copy relevant CSS to Canada.css

### DIFF
--- a/ckanext/canada/templates/public/package/read.html
+++ b/ckanext/canada/templates/public/package/read.html
@@ -233,7 +233,7 @@
 
 {%- block custom_styles %}
 {{ super() }}
-<link rel="stylesheet" href="http://open.canada.ca/sites/all/modules/custom/opendata_blog/css/opendata_blog.css" />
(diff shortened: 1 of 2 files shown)


ckanext/canada/templates/public/package/read.html
```diff
@@ -233,7 +233,7 @@ <h4>{{ _('Contact Information') }}</h4>
 
 {%- block custom_styles %}
 {{ super() }}
-<link rel="stylesheet" href="http://open.canada.ca/sites/all/modules/custom/opendata_blog/css/opendata_blog.css" />
+<link rel="stylesheet" href="/data/static/opendata_blog.css" />
```
wardi: Can you do it like this instead
```
<link rel="stylesheet" href="{{ h.url_for_static('/static/opendata_blog.css" />
```
that way it will work with whatever root path we have set

thriuin:
Leave a comment

PR #934, opened 2018-06-25
As per OGP-430, add the following columns back in:

thriuin:
- external_internal
 - service_type

and remove this column:
 - special_designations

### DIFF
--- a/bin/filter_service.py
+++ b/bin/filter_service.py
@@ -4,8 +4,7 @@ import csv
 import sys
 
 REMOVE_COLUMNS = [
-    'external_internal',
-    'service_type',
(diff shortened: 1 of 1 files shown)


wardi:
there's some redundant (and now incorrect) `visible_to_public` information in https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/tables/service.yaml

@thriuin should we remove the visible_to_public values from the yaml files, or do you think we should use the yaml files' visible_to_public values in these filter scripts?

thriuin:
I agree with the idea of using the yaml files to save this information. Thanks

PR #936, opened 2018-07-12
Update GoC orgs for July 12

thriuin:
Updating the transitional organizations file for July 12th, 2018


transitional_orgs.jsonl
```diff
@@ -1,4 +1,3 @@
-{"extras": [{"key": "shortform_fr", "value": "AB"}, {"key": "shortform", "value": "AB"}], "id": "436939fc-d6f6-43fc-b00e-dc134654469c", "name": "ab", "title": "Government of Alberta | Gouvernement de l'Alberta"}
```
wardi: Should we keep this one (even if it's only on the portal and not the registry)? Or make a separate file for such orgs?

thriuin: It's convenient to have all organizations on file if you need to quickly regenerate the site or a test site. I have no objection to two files, but with the current arrangement it is very fast to generate a new and complete orgs files with the ckanapi command.

wardi: So, if you add this line back in when you `ckanapi load organizations` on the registry there will be a province of alberta there too. Any issues with that?

wardi: also this file seems to be less and less transitional. If we go the two-file route we could name them `gc_orgs.jsonl` and `external_orgs.jsonl` or similar

thriuin: That works too and it is a better naming convention

PR #948, opened 2018-12-20
filter consultation status - for my learning

PatLittle:
To discuss for my own learning purposes, someone will need to fix this up to actually implement.

### DIFF
--- a/bin/filter_consultations.py
+++ b/bin/filter_consultations.py
@@ -39,4 +39,24 @@ def main():
         except ValueError:
             pass
 
+def replace():
+    replacements = (
+        ("CN", "C"),
+        ("CR", "C"),
+        ("CA", "C"),
+        )
+    # I was able to make it run locally as below- 
+    # with open("consultations.csv",'rb') as f, open("consultations_filtered.csv", "w") as outputfile:
+    #   for line in f:
+    #       for pair in replacements:
+    #           line = line.replace(*pair)
+    #       outputfile.write(line)                            
+    # For our environment it would be more like below? (which I don't think works)  
+    with sys.stdin as f, sys.stdout as outputfile:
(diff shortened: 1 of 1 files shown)


bin/filter_consultations.py
```diff
@@ -39,4 +39,24 @@ def main():
...
+        ("CR", "C"),
+        ("CA", "C"),
+        )
+    # I was able to make it run locally as below- 
+    # with open("consultations.csv",'rb') as f, open("consultations_filtered.csv", "w") as outputfile:
+    #   for line in f:
+    #       for pair in replacements:
+    #           line = line.replace(*pair)
+    #       outputfile.write(line)                            
+    # For our environment it would be more like below? (which I don't think works)
```
wardi: you should be able to run the script from the command prompt something like
```
python filter_consultations.py < consultations.csv > consultations_filtered.csv
```
so you don't need to specify the file names in the script

bin/filter_consultations.py
```diff
@@ -39,4 +39,24 @@ def main():
...
+    # with open("consultations.csv",'rb') as f, open("consultations_filtered.csv", "w") as outputfile:
+    #   for line in f:
+    #       for pair in replacements:
+    #           line = line.replace(*pair)
+    #       outputfile.write(line)                            
+    # For our environment it would be more like below? (which I don't think works)  
+    with sys.stdin as f, sys.stdout as outputfile:
+        for line in f:
+            for pair in replacements:
+                line = line.replace(*pair)
```
wardi: this is going to replace *all* instances of "CN" "CR" "CA", e.g. "CN Rail" in a different field would be modified. Instead of this separate loop try modifying `row['columnname']` before line 38 above.

PR #951, opened 2018-12-26
Service changes

wardi:
add tests, change service standard targets from 0-100 integer percentage to decimals on the interval (0.00000, 1.00000)


wardi:
quick fix to accept any values for service_std_target instead of migrating all data for now

PR #956, opened 2019-01-17
add translation for user last modified

PatLittle:


wardi:
this would be fine as a PR against master because it would fix other types immediately. I'll accept it here though because it will get merged to master soon enough

PR #965, opened 2019-03-07
Django search integration

thriuin:
First attempt to add the search integration into the extension.


thriuin:
Will instead push to a feature branch on this repo

PR #982, opened 2019-05-14
add short choice for deputy

PatLittle:
added short choices for deputy heads

### DIFF
--- a/ckanext/canada/tables/briefingt.yaml
+++ b/ckanext/canada/tables/briefingt.yaml
@@ -115,7 +115,7 @@ resources:
       D:
         en: Deputy head (including a person appointed to a position of an equivalent rank)
         fr: Administrateurs généraux (y compris une personne nommée à un poste de rang équivalent)
-        d_short:
+        shortform:
           en: Deputy head or equivalent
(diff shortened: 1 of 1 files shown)


ckanext/canada/tables/briefingt.yaml
```diff
@@ -113,8 +113,11 @@ resources:
         en: Minister
         fr: Ministre
       D:
-        en: deputy head (including a person appointed to a position of an equivalent rank)
-        fr: Administrateurs Généraux (y compris une personne nommée à un poste de rang équivalent)
+        en: Deputy head (including a person appointed to a position of an equivalent rank)
+        fr: Administrateurs généraux (y compris une personne nommée à un poste de rang équivalent)
+        d_short:
```
wardi: how abour `shortform:` instead? The `d` is implied by it being part of the `D:` value above

wardi:
```suggestion
        shortform:
```

wardi: also I thought the lowercase `deputy head` text was specifically requested (part of change ffbe2b691c )

PR #995, opened 2019-07-29
Drupal integration

RabiaSajjad:
Proposing code changes for integration with Drupal external comment and voting


ckanext/canada/helpers.py
```diff
@@ -155,10 +155,17 @@ def remove_duplicates(a_list):
             
     return s
 
-
 def dataset_comments(pkg_id):
     if config.get('ckanext.canada.drupal_url'):
-        return '/' + h.lang() + '/external-entity/ckan-' + pkg_id + '?_wrapper_format=ajax'
+	return '/' + h.lang() + '/external-comment/dataset/' + pkg_id
```
wardi: 4 space indent please :-)

ckanext/canada/templates/public/package/read.html
```diff
@@ -238,6 +238,11 @@ <h4>{{ _('Contact Information') }}</h4>
   {% endif %}
 {% endblock contact_information %}
 
+
+{% block package_rate %}
+{% include "package/rate.html" %}
+{% endblock %}
```
wardi: let's put this one line in directly instead of adding a new `package/rate.html` template to `{% include %}` (we could do the same for the one below too, it used to be a large block of logic but is just an ajax call not)

PR #1000, opened 2019-08-19
update wrongdoing

PatLittle:


wardi:
Might be time to switch these codes to uppercase, at least, to be consistent with other controlled lists. I'll need to add a migration as well to switch the column type

PR #1005, opened 2019-09-24
updated text

PatLittle:


wardi:
Similar text appears in a few other files, (I searched for "You may select multiple")

- ckanext/canada/templates/internal/recombinant/create_pd_record.html
- ckanext/canada/schemas/presets.yaml
- ckanext/canada/schemas/info.yaml
- ckanext/canada/schemas/dataset.yaml

PR #1009, opened 2019-10-04
Enabling nightly build of the Grants Solr core for Django-based search

thriuin:
I was thinking we'd build both Solr cores for now so we can compare


thriuin:
Thanks!

PR #1012, opened 2019-10-17
Update wet theme

HoussamBedja:
Hi Ian,

Could you please review this pull request then deploy it to staging? 

This branch already includes the changes that you previously reviewed in the "add-wet-theme" + the update to WET v5 on both portal and registry sites. 

Before this pull request, I was asked by Ross to first merge commits from the master branch then also from staging branch, which is why you will find lots of changes added that are not made by me, but from what was already existing in master. 

I am trying to have it up on staging as soon as possible for the business team to test both sites (and will work on cleaning up unused code in parallel next week).

Thanks!


wardi:
Yeah we don't want staging merged into this branch. Would you like to reset to the commit before the merge or should I?

wardi:
Also please make the PR against master instead of staging

wardi:
@HoussamBedja I force-pushed to remove the merge from staging, please reset the branch on your repo to the same commit before adding any more changes

wardi:
Testing this locally I downloaded https://github.com/wet-boew/GCWeb/releases/tag/v5.1 and when I run the registry I don't see the registry menu any more.

The README needs to be updated for the new WET installation steps.

wardi:
@HoussamBedja thanks, would you add instructions for installing the necessary files for WET as well? There were some instructions in the ckanext-wet-boew readme you can start from

HoussamBedja:
@wardi I added instructions to install the wet files. Please let me know if you have any feedback

HoussamBedja:
Oh true, the GCWeb version number is change. the --strip-components option is the same.

wardi:
Are you sure? The last time I installed the gcweb package it had no "GCWeb" subdirectory because I used the "cdn" version https://github.com/wet-boew/GCWeb/releases/download/v5.1/themes-cdn-5.1-gcweb.zip should I have used the dist version instead?

HoussamBedja:
Hmm maybe I'm not sure. I just tested these commands and they give the desired output (a GCWeb directory containing the theme files as direct children, with no subdirectory)

ckanext/canada/helpers.py
```diff
@@ -459,3 +460,20 @@ def _add_extra_longitude_points(gjson):
     return {u'coordinates': [out], u'type': u'Polygon'}
 
 
+def recombinant_description_to_markup(text):
```
wardi: 1. it's confusing to have a helper in the canada extension with a `recombinant_...` name
2. why can't we use the markdown helper instead of this custom regexp helper?

wardi: but I realize that I wrote this originally.. hmm.

ckanext/canada/plugins.py
```diff
@@ -51,6 +56,9 @@ def update_config(self, config):
             "ckan.user_list_limit": 4000
         })
 
+    def wet_theme(self):
+        return 'GCWeb'
+
```
wardi: aren't we using the same wet theme for portal and registry now? If so we could remove these methods.

ckanext/canada/templates/internal/header.html
```diff
@@ -1,64 +1,89 @@
...
+                          'travelq',
+                          'travela',
+                          'grants',
+                          'reclassification',
+                          'wrongdoing',
+                          'briefingt',
+                          'consultations',
+                          'inventory',
+                          'dac',
+                          'nap',
```
wardi: did you intend to remove the experiment type as part of this PR?

HoussamBedja: Oh no. I don't know when that disappeared. Thanks for the catch!

ckanext/canada/helpers.py
HoussamBedja: Hhaha yes I'm not sure who wrote this code. Maybe change it later?

PR #1016, opened 2019-11-04
OPEN-194: Audit trail on ATI

RabiaSajjad:


ckanext/canada/tables/ati.yaml
```diff
@@ -52,6 +50,24 @@ resources:
     datastore_type: int
```
wardi: I expect this will cause problems when importing the data. Have you tried importing the existing data? We likely need to clean up the old data so it validates as `int`s here.

wardi: (that's something to add to the migration script)

PR #1018, opened 2019-11-07
OPEN-164: draft template for qpnotes

RabiaSajjad:


bin/pd/Makefile
```diff
@@ -183,6 +184,30 @@ $(workdir)/briefingt.csv:
...
+###  Question Period Notes
+###
+.PHONY: qpnotes
+qpnotes: upload-qpnotes rebuild-qpnotes
+
+.PHONY: upload-qpnotes
+upload-qpnotes: $(workdir)/filtered/qpnotes.csv
+	@echo Uploading Question Period Notes...
+	@$(ckanapi) action resource_patch -c $(registry_ini) \
+	id=ba1284e5-db5f-43e1-854f-2beb7fe8f04b upload@"$(workdir)/filtered/qpnotes.csv"
```
wardi: Where did you get the resource ID from here?

ckanext/canada/tables/choices/qpnotes_minister.yaml
```diff
@@ -0,0 +1,35 @@
...
+M126: Ginette Petitpas Taylor
+M127: Seamus O' Regan
+M128: Pablo Rodriguez
+M129: Bill Blair
+M130: Mary Ng
+M131: Filomena Tassi
+M132: Jonathan Wilkinson
+M133: David Lametti
+M134: Bernadette Jordan
+M135: Joyce Murray
```
wardi: Are these ids from the source CSV? I think we need to store the actual names in our table instead of an opaque minister ID (the keys are what's stored not the values)

ckanext/canada/tables/qpnotes.yaml
```diff
@@ -0,0 +1,357 @@
+dataset_type: qpnotes
+target_dataset: qpnotes
+
+title: Proactive Disclosure - Question Period Notes
+notes: Access, upload and modify Question Period notes for your organization
+
+template_version: 3
+template_updated: 2019-11-18
```
wardi: don't need `template_updated` for new templates

ckanext/canada/tables/qpnotes.yaml
```diff
@@ -0,0 +1,357 @@
...
+
+portal_type: info
+collection: qpnotes
```
wardi: The collection here corresponds to the dataset collection for search purposes https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/schemas/presets.yaml#L33 These have been set to "pd" for the most part. Odd, though that I don't have that in the choice list.

ckanext/canada/tables/qpnotes.yaml
```diff
@@ -0,0 +1,357 @@
...
+      fr: Ce champ affichera le nom du ministre pour qui la note pour la période de questions a été créée.
+    obligation: Mandatory
+    excel_required: true
+    form_required: true
+    format_type: Controlled List
+    validation: This field must not be empty
+    visible_to_public: Yes
+    occurrence: Single
+    datastore_type: text
+    excel_full_text_choices: true
```
wardi: can be set to False when you remove the minister codes

ckanext/canada/tables/qpnotes.yaml
```diff
@@ -0,0 +1,357 @@
...
+    fgColor: FF336B87
+    patternType: solid
+excel_header_style:
+  PatternFill:
+    patternType: solid
+    fgColor: FF90AFC5
+excel_column_heading_style:
+  PatternFill:
+    patternType: solid
+    fgColor: FF90AFC5
```
wardi: Have you chosen some nice header+edge colours for this template? This is where we get to make the template stand out from the others :-)

bin/pd/Makefile
RabiaSajjad: `paster recombinant update -a -c /opt/tbs/ckan/conf/development.ini` gave it to me

wardi: no that's not right. The resources here are resources in datasets on the registry that are used to publish the csv files. Usually we get the business team to create these for us.

bin/download_minister.py
```diff
@@ -0,0 +1,165 @@
...
+    # if crawled choices list does not exist
+    elif existing_list and not choices:
+        return set_status(existing_list)
+    # if both exist, compare both lists for changes
+    elif existing_list and choices:
+        for row in existing_list:
+            if row in choices:
+                # case: minister found in both lists
+                # update minister if existing list is different from choices list
+                if cmp(existing_list[row], choices[row]):
```
wardi: write this as `if existing_list[row] != choices[row]` instead, there's not much use for `cmp` in modern python

bin/download_minister.py
```diff
@@ -0,0 +1,165 @@
...
+    # if existing list does not exist
+    if not existing_list and choices:
+        return set_status(choices)
+    # if crawled choices list does not exist
+    elif existing_list and not choices:
+        return set_status(existing_list)
+    # if both exist, compare both lists for changes
+    elif existing_list and choices:
+        for row in existing_list:
+            if row in choices:
```
wardi: instead of looping over both lists can you turn one into a dict or set and do a lookup?

bin/download_minister.py
```diff
@@ -0,0 +1,165 @@
...
+                        # case: same minister with existing position not in choices then update end date as today
+                        for position in existing_list[row]['positions_' + langCode]:
+                            if not [d for d in choices[row]['positions_' + langCode]
+                                    if d['title_' + langCode] == position['title_' + langCode]]:
+                                position['end_date'] = date.today().strftime("%Y-%m-%dT%H:%M:%S")
+            else:
+                # case: minister found in existing list but now in new list, update end date as today
+                for langCode in lang_codes:
+                    for position in existing_list[row]['positions_' + langCode]:
+                        position['end_date'] = date.today().strftime("%Y-%m-%dT%H:%M:%S")
```
wardi: we're running the script nightly or irregularly so storing the date down to seconds seems like too much precision

bin/download_minister.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    Read XML from external source to generate JSON choices for ministers
+    """
+    choices = {}
+    xml_tree = {}
+
+    # collect responses for both languages
+    for langCode in lang_codes:
+        # url = 'https://www.ourcommons.ca/Parliamentarians/' + langCode + '/ministries/Export?output=XML'
+        url = 'http://localhost/export-' + langCode + '.xml'
```
wardi: use a command line parameter for the url?

bin/download_minister.py
RabiaSajjad: the date from external source comes in this format so I kept it in the same format

bin/download_minister.py
```diff
@@ -0,0 +1,165 @@
...
+        response = requests.get(url, stream=True)
+
+        # if valid response then save as xml tree
+        if response.status_code == 200:
```
wardi: requests will raise an exception on an error, so there's no need to check the status code.

bin/download_minister.py
```diff
@@ -0,0 +1,165 @@
...
+                        + ' (' + row.xpath('PersonShortHonorific/text()')[0] + ')',
+                        'honorific_title_' + langCode: row.xpath('PersonShortHonorific/text()')[0],
+                        'precedence': row.xpath('OrderOfPrecedence/text()')[0],
+                        'positions_' + langCode: [position],
+                    }
+        response.close()
+    return choices
+
+
+def set_status(list):
```
wardi: please don't shadow builtin names like "list" with variable names (It's also confusing to name a dict "list")

bin/download_minister.py
RabiaSajjad: this is a mistake, I was testing using my localhost

wardi:
just some style comments about the download script, this is fine to deploy to staging

bin/pd/Makefile
wardi: need to fix this before merge to master (staging doesn't matter)

bin/download_minister.py
wardi: need to fix before merge to master

bin/download_minister.py
```diff
@@ -0,0 +1,148 @@
...
+    existing_choices = {}
+
+    # fetch existing choices file, if it exists
+    if os.path.isfile(OUTPUT_FILE):
+        with open(OUTPUT_FILE) as json_file:
+            try:
+                existing_choices = json.load(json_file)
+            except ValueError:
+                return False
+        json_file.close()
```
wardi:
```suggestion
```
not necessary when using a context manager

bin/pd/Makefile
```diff
@@ -194,12 +194,13 @@ qpnotes: upload-qpnotes rebuild-qpnotes
...
 	@echo Uploading Question Period Notes...
 	@$(ckanapi) action resource_patch -c $(registry_ini) \
-	id=ba1284e5-db5f-43e1-854f-2beb7fe8f04b upload@"$(workdir)/filtered/qpnotes.csv"
+	id=c55a2862-7ec4-462c-a844-22acab664812 upload@"$(workdir)/filtered/qpnotes.csv"
 
 .PHONY: rebuild-qpnotes
 rebuild-qpnotes: $(workdir)/filtered/qpnotes.csv
-	@echo Downloading ministers list for Question Period Notes...
-	@$(CURDIR)/../download_minister.py
+    # uncomment once we have whitelisting completed on server
```
wardi: have to use hard tab characters or the makefile will refuse to run

bin/download_minister.py
```diff
@@ -97,11 +139,14 @@ def download_xml_from_source():
...
         row_fr = next(iter_fr)
+
         # set name
         name = row_en.xpath('PersonOfficialFirstName/text()')[0] + ' ' + \
                row_en.xpath('PersonOfficialLastName/text()')[0]
-        # set position code with position initials
-        position_code = (''.join([x[0].upper() for x in row_en.xpath('Title/text()')[0].split(' ')]))
+
+        # find position code in ministries dict
+        position_code = ministries_dict.keys()[ministries_dict.values().index(row_en.xpath('Title/text()')[0])]
```
wardi: kinda nasty. Can you write the dict with keys and values swapped so you don't need to do this?

bin/pd/Makefile
```diff
@@ -183,6 +184,31 @@ $(workdir)/briefingt.csv:
...
+###  Question Period Notes
+###
+.PHONY: qpnotes
+qpnotes: upload-qpnotes rebuild-qpnotes
+
+.PHONY: upload-qpnotes
+upload-qpnotes: $(workdir)/filtered/qpnotes.csv
+	@echo Uploading Question Period Notes...
+	@$(ckanapi) action resource_patch -c $(registry_ini) \
+	id=c55a2862-7ec4-462c-a844-22acab664812 upload@"$(workdir)/filtered/qpnotes.csv"
```
wardi: we could copy the ministers.json file to `$(portal_static)` after this line

PR #1020, opened 2019-11-13
adding google analytics to registry site (staging + prod)

HoussamBedja:
Adding google analytics to the registry site (on both staging and production)

### DIFF
--- a/ckanext/canada/plugins.py
+++ b/ckanext/canada/plugins.py
@@ -160,6 +160,7 @@ class DataGCCAInternal(p.SingletonPlugin):
             'parse_release_date_facet',
             'is_ready_to_publish',
             'get_datapreview_recombinant',
+            'googleanalytics_id',
             ])
 
     def configure(self, config):
--- a/ckanext/canada/plugins.py
+++ b/ckanext/canada/plugins.py
@@ -160,7 +160,6 @@ class DataGCCAInternal(p.SingletonPlugin):
             'parse_release_date_facet',
             'is_ready_to_publish',
             'get_datapreview_recombinant',
-            'googleanalytics_id',
             ])
 
     def configure(self, config):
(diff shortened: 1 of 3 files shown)


ckanext/canada/templates/internal/base.html
```diff
@@ -10,3 +10,20 @@
...
+        {% set id = '2' %}
+    {% endif %}
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-152046908-{{id}}"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+
+      gtag('config', 'UA-152046908-{{id}}');
```
wardi: why aren't you using `{{ h.googleanalytics_id() }}` here instead of using an `if`/`else` block and a hard-coded id?

HoussamBedja: Yeah I conpletely missed that variable! Just changed that in a new commit

ckanext/canada/plugins.py
```diff
@@ -160,6 +160,7 @@ def get_helpers(self):
             'parse_release_date_facet',
             'is_ready_to_publish',
             'get_datapreview_recombinant',
+            'googleanalytics_id',
```
wardi: this is already in the public plugin which we use on both portal and registry, so I don't think we need it here

HoussamBedja: True. I just removed it

wardi:
Thanks, let me know if this is good to go and I'll push to prod

HoussamBedja:
Yes. It's good to go now

PR #1022, opened 2019-11-21
Experiment inventory

thriuin:
I'm leaving out the Makefile and test for now. I will add once the template is accepted.

### DIFF
--- a/ckanext/canada/tables/experiment.yaml
+++ b/ckanext/canada/tables/experiment.yaml
@@ -2,7 +2,7 @@
 collection: experiment
 dataset_type: experiment
 title: "Experimentation Inventory"
-notes: "Access, upload and modify the Experimentation Inventory for your organization\n"
+notes: "Access, upload and modify the Experimentation Inventory for your organization"
 portal_type: info
 target_dataset: experiment
 template_version: 3
--- a/ckanext/canada/tables/experiment.yaml
+++ b/ckanext/canada/tables/experiment.yaml
@@ -1,0 +1,396 @@
+
+collection: experiment
+dataset_type: experiment
+title: "Experimentation Inventory"
+notes: "Access, upload and modify the Experimentation Inventory for your organization\n"
+portal_type: info
+target_dataset: experiment
+template_version: 3
+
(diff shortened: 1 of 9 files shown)


ckanext/canada/tables/experiment.yaml
```diff
@@ -0,0 +1,396 @@
...
+          en: "Last Updated"
+          fr: "Dernière mise à jour"
+        obligation: Mandatory
+        occurrence: Single
+        validation: "This field must not be empty"
+        visible_to_public: true
+
+      - datastore_id: experimental_area
+        choices:
+          001:
```
wardi: if these have to be numbers put them in quotes so that they're treated as strings instead of integer keys (allowed in python yaml but not json)

Ideally we should be using letter codes for these instead, 0-prefixed numeric codes are difficult to enter in excel

wardi: same comment about the other choice lists. Otherwise this looks good to me!

thriuin: I agree - these are just the values that were in the DEP. I spoke to AH and got the Okay to change these to alphabetic code values.

thriuin: Actually, I need your help with setting the Excel colors. I thought I was setting it to orange but the template came out blue.

wardi: should be AARRGGBB format

thriuin: Thanks, it worked. I have made the code value changes.

wardi:
This looks great and I see the tests ran on travis. Do you want this on master now or will we be running it in staging for a while?

wardi:
or maybe simpler: looks good to me. Feel free to merge and deploy

thriuin:
OK, thanks. I think it's safe for master. Thanks for your help!

thriuin:
I deleted the branch after merging - I assume we don't need the clutter?

wardi:
Yes, I often forget to delete the branches but it's a good idea

PR #1024, opened 2019-11-22
Adobe analytics portal

HoussamBedja:


ckanext/canada/helpers.py
```diff
@@ -175,6 +175,21 @@ def portal_url():
...
+    return "2" #return 1 if page requires a login and 2 if page is public
+
+def adobe_analytics_lang():
+    if h.lang() == 'en':
+        return 'eng'
+    elif h.lang() == 'fr':
+        return 'fra'
+
+def adobe_analytics_js():
+    if config.get('ckan.site_url', '') == 'https://staging.open.canada.ca':
```
wardi: please use a new configuration option instead of hard-coding the asset locations based on the site_url setting

ckanext/canada/helpers.py
```diff
@@ -175,6 +175,21 @@ def portal_url():
 
 def googleanalytics_id():
     return str(config.get('googleanalytics.id'))
+
+def adobe_analytics_login_required():
+    return "2" #return 1 if page requires a login and 2 if page is public
```
wardi: Should there be a parameter here so we can determine whether a page is public or are you planning on using different helper functions for the portal+registry?

HoussamBedja: Registry does not implement Adobe analytics at all, so it is not related to this function. It is only installed on the portal side. For now this function only returns 2 since all portal pages are public. But the function could be edited in the future if one day we end up having login required on the portal side.

wardi: Can it though? With no parameters how would the helper know which page is being visited?

HoussamBedja: Ok i see what you mean. For now I could add the url path or something, which will then be use in the logic once there is a need to implement it.

ckanext/canada/helpers.py
```diff
@@ -186,7 +186,7 @@ def adobe_analytics_lang():
         return 'fra'
 
 def adobe_analytics_js():
-    if config.get('ckan.site_url', '') == 'https://staging.open.canada.ca':
+    if config.get('ckanext.site_environment', '') == 'staging':
```
wardi: No, what I mean is put the whole "//assets.adobetm.com..." string in a configuration variable instead of hard-coding the two values we're using behind an opaque "site_environment" setting

ckanext/canada/helpers.py
```diff
@@ -176,8 +176,8 @@ def portal_url():
 def googleanalytics_id():
     return str(config.get('googleanalytics.id'))
 
-def adobe_analytics_login_required():
-    return "2" #return 1 if page requires a login and 2 if page is public
+def adobe_analytics_login_required(current_url):
+    return current_url #return 1 if page requires a login and 2 if page is public
```
wardi: :+1:

ckanext/canada/helpers.py
HoussamBedja: Oh okay i see. Yes that would make more sense.

ckanext/canada/templates/public/base.html
```diff
@@ -31,7 +31,7 @@
 {%- endblock -%}
 
 {% block adobe_analytics %}
-  <script src="{{ h.adobe_analytics_js() }}"></script>
+  <script src="{{ h.adobe_analytics_id() }}"></script>
```
wardi: How about
```jinja2
{% if h.adobe_analytics_js() %}
  <script src="{{ h.adobe_analytics_js() }}"></script>
{% endif %}
```
so that we don't end up with an empty tag when this isn't configured.

Also add the new config option to the readme so new users know it exists

ckanext/canada/helpers.py
HoussamBedja: Also I should just return "2" here for now. Then we can use 'current_url' when we need to build a logic inside this function

ckanext/canada/templates/public/base.html
HoussamBedja: Yes that's better.

PR #1033, opened 2020-01-22
Added functionality to allow editing of records in excel template

hodge-jai:
Allows users to download specific records from a recombinant datatype in an excel template


ckanext/canada/controller.py
```diff
@@ -229,6 +230,9 @@ def datatable(self, resource_name, resource_id):
                         )
                     )
                 )
+        # Add empty column for checkboxes #
+        for row in aadata:
+            row.insert(0,u'')
```
wardi: do this in the creation of the original aadata list to save a bunch of unnecessary data copying

ckanext/canada/templates/internal/recombinant/resource_edit.html
```diff
@@ -32,40 +32,15 @@
...
-      });
-     });
-  </script>
 {% endblock %}
 
 {%- block styles %}
   {{ super() }}
-  <link rel="stylesheet"  href="https://cdn.datatables.net/buttons/1.2.2/css/buttons.dataTables.min.css" />
+
 {% endblock %}
```
wardi: please delete this whole block. all it contains now is `{{ super() }}` (the original definition of the block)

ckanext/canada/templates/internal/package/wet_datatable.html
```diff
@@ -44,3 +22,64 @@ <h2 id="prtable">{{_("Preview")}}</h2>
...
+  {% endif %}
+  {% for f in chromo.fields %}
+  {{- "null" if 'preview_class' not in f else
+    '{"className":"' | safe ~ f['preview_class'] ~ '"}' | safe }}
+  {%- if not loop.last %},{% endif %}
+  {% endfor %}],
+  dom: "Blfrtip",
+          buttons: [
+    'colvis',
+    {extend:"selected",text:"Edit in template",action: function ( e, dt, button, config ){
```
wardi: use `"{{ _("Edit in template") }}"` so we can translate this text

ckanext/canada/templates/internal/package/wet_datatable.html
```diff
@@ -44,3 +22,64 @@ <h2 id="prtable">{{_("Preview")}}</h2>
...
+        var row_data = JSON.stringify(dt.rows({selected:true}).data().toArray());
```
wardi: as I mentioned on the other repo, instead of JSON.stringify you can just repeat the <input> fields with the same name to POST multiple primary key values to the recombinant endpoint

ckanext/canada/controller.py
```diff
@@ -214,7 +216,7 @@ def datatable(self, resource_name, resource_id):
             sort=u', '.join(sort_list),
         )
 
-        aadata = [
+        aadata = [[u'']+
```
wardi: put the `[u''] +` on its own line to make the pycharm PEP8 checker happy

ckanext/canada/controller.py
```diff
@@ -223,16 +225,18 @@ def datatable(self, resource_name, resource_id):
             pkg = lc.action.package_show(id=res['package_id'])
             fids = [f['datastore_id'] for f in chromo['fields']]
             pkids = [fids.index(k) for k in aslist(chromo['datastore_primary_key'])]
+            log = getLogger(__name__)
```
wardi: this should be a module (global) variable

ckanext/canada/helpers.py
```diff
@@ -242,11 +242,14 @@ def is_ready_to_publish(package):
...
 
-def get_datapreview_recombinant(resource_name, res_id):
+def get_datapreview_recombinant(resource_name, resource_id, owner_org, dataset_type):
     from ckanext.recombinant.tables import get_chromo
     chromo = get_chromo(resource_name)
+
+
     default_preview_args = {}
 
+
```
wardi: please remove the double-spacing

wardi:
what happened to controller.py, did you save it with linebreaks or something? it's showing as 100% changed

PR #1034, opened 2020-01-22
Removed unnecessary text from the new_resource page

hodge-jai:
Simple typo fix


wardi:
PRs should be made against master, not staging. I'll manually merge to staging as required, but if your work is based on staging I can't merge to master since it includes a bunch of other changes only deployed to staging

PR #1045, opened 2020-02-03
Open 380 minor ei typo

thriuin:


thriuin:
Trivial type fix for EI schema and type from previous commit >-(

PR #1046, opened 2020-02-05
OPEN-486: Add new file formats (for FGP)

RabiaSajjad:


ckanext/canada/schemas/presets.yaml
```diff
@@ -3230,6 +3230,9 @@ presets:
       mimetype: application/x-filegdb
     - value: Flat raster binary
       replaces: [flat raster binary]
+    - value: GDB
+      mimetype: application/x-filegdb
+      label: GeoSoftDatabases
```
wardi: We already have an entry for "FGDB/GDB"

ckanext/canada/schemas/presets.yaml
```diff
@@ -3252,6 +3255,9 @@ presets:
       mimetype: application/gml+xml
       openness_score: 3
       replaces: [gml]
+    - value: GRD
+      mimetype: application/x-filegdb
```
wardi: is this the right mimetype?

ckanext/canada/schemas/presets.yaml
wardi: @PatLittle we have a bunch of these geospatial formats now that have unilingual labels that contain product names. Should we go back to using "file extension (+ description as required)" or is it ok to have unilingual labels for this controlled list?

ckanext/canada/schemas/presets.yaml
RabiaSajjad: fixed

PR #1047, opened 2020-02-12
migration scripts now exit with status code 85 when there is a column…

hodge-jai:
… mismatch


bin/migrate/migrate_cansim_export.py
```diff
@@ -47,81 +47,84 @@
...
+                    else:
+                        res['resource_type']='dataset'
+
+                print simplejson.dumps(rec)
+
+        except IOError:
+            print >> sys.stderr, 'Error while reading line.'
+
+    except KeyError:
+        sys.exit(85)
```
wardi: you don't neet to use two `try:` blocks, just add a second `except` clause

wardi: actually, this script isn't related to PD so you can ignore it (maybe we should move it out of this directory)

bin/warehouse/migrate_all.py
```diff
@@ -0,0 +1,106 @@
...
+
+        except IOError:
+            infile.seek(0)
+            outfile.seek(0)
+            if len(matching_files[1:]) == 0:
+                outfile.write(infile.read())
+                matching_files[:] = []
+            else:
+                matching_files = matching_files[1:]
+
```
wardi:
```suggestion
            matching_files = matching_files[1:]
    else:
        shutil.copyfile(infile, outfile)
```

bin/warehouse/migrate_all.py
```diff
@@ -0,0 +1,106 @@
...
+if not matching_files:
+    outfile.write(infile.read())
+
+else:
+    while matching_files:
+        try:
+            run_scripts(infile, outfile, matching_files)
+            break
+
+        except IOError:
```
wardi: Need to check the exit code of the first process here

bin/warehouse/migrate_all.py
```diff
@@ -0,0 +1,106 @@
...
+else:
+    pd_type = pd_type.replace("-", "_")
+    search_pd = '*_{0}_*'.format(pd_type)
+    matching_files = sorted(glob.glob('../migrate/' + search_pd))
+
+# if there are no migration scripts to run, write the csv file to output file
+if not matching_files:
+    outfile.write(infile.read())
+
+else:
```
wardi:
```suggestion
```

bin/warehouse/migrate_all.py
wardi: oh, copyfile takes the path not a file object, but you get the idea

bin/migrate/migrate_ati_2019_11.py
```diff
@@ -18,8 +18,11 @@
         if 'warehouse' in sys.argv[1:]:
             line['user_modified'] = ''  # special "we don't know" value
         else:
-            raise KeyError('Invalid value')
+            line['user_modified'] = '*'
```
wardi: need to move the comment above to this line, same for other changes

wardi: if you're updating them all anyway I bet you can omit the `line['user_modified'] = ''` block because we don't seem to have it for the other new fields

bin/migrate/migrate_ati_2019_11.py
```diff
@@ -13,7 +13,16 @@
 out_csv = unicodecsv.DictWriter(sys.stdout, fieldnames=FIELDNAMES, encoding='utf-8')
 out_csv.writeheader()
 
+try:
+    for line in in_csv:
+        if 'warehouse' in sys.argv[1:]:
+            line['user_modified'] = ''  # special "we don't know" value
+        else:
+            line['user_modified'] = '*'
```
wardi:
```suggestion
        if 'warehouse' not in sys.argv[1:]:
            line['user_modified'] = '*'  # special "we don't know" value
```

PR #1051, opened 2020-03-09
OPEN-514: add form to travelq template

RabiaSajjad:


ckanext/canada/tables/travelq.yaml
```diff
@@ -5,7 +5,7 @@ title: Proactive Disclosure - Travel Expenses
 notes: Access, upload and modify the monthly travel expense reports for your organization
 
 template_version: 3
-template_updated: 2019-07-10
+template_updated: 2020-03-09
```
wardi: not required for this change

wardi:
otherwise LGTM

PR #1052, opened 2020-03-09
OPEN-510: add form to hospitalityq template

RabiaSajjad:


ckanext/canada/tables/hospitalityq.yaml
```diff
@@ -5,7 +5,7 @@ title: Proactive Disclosure - Hospitality Expenses
 notes: Access, upload and modify the quarterly hospitality expenses for your organization
 
 template_version: 3
-template_updated: 2019-07-10
+template_updated: 2020-03-09
```
wardi:
```suggestion
template_updated: 2019-07-10
```

wardi:
I can't remember, do some of these fields get rendered as markdown?

PR #1053, opened 2020-03-09
added create/edit form to travela pd-type

hodge-jai:


ckanext/canada/tables/travela.yaml
```diff
@@ -388,6 +402,9 @@ resources:
       fr: Expliquez brièvement, en anglais, la ou les raisons significatives de l’augmentation ou de la diminution des dépenses annuelles de voyage pour les fonctionnaires et non-fonctionnaires.
     datastore_type: text
     excel_column_width: 50
+    form_snippet: scheming/form_snippets/textarea.html
+    form_attrs:
+      style: "width: 100%; display: block"
```
wardi: @RabiaSajjad says these explanation of significant variance fields can contain markdown, can you use the markdown snippet instead

RabiaSajjad: Hi @hodge-jai, let's keep it simple and go with the `scheming/form_snippets/textarea.html` snippet. I cannot find markdown implementation on drupal for rendering. Apologies for re-work.

PR #1057, opened 2020-03-19
Ingest suggested dataset submissions

RabiaSajjad:


ckanext/canada/commands.py
```diff
@@ -439,6 +445,72 @@ def changed_datasets(self, since_date):
...
+        results = True
+        counter = 0
+        batch_size = 100
+        existing_suggestions = {}
+        while results:
+            packages = registry.action.package_search(q='type:prop', start=counter, rows=batch_size, include_private=True)['results']
+            if packages:
+                for package in packages:
+                    existing_suggestions[package['id']] = package
+                counter += batch_size
```
wardi: use `counter += len(packages)` to handle package_search returning less than you expect (e.g. because of configuration)

ckanext/canada/commands.py
```diff
@@ -439,6 +445,72 @@ def changed_datasets(self, since_date):
...
+            else:
+                results = False
+
+        # load data from csv
+        csv_file = io.open(filename, "r", encoding='utf-8-sig')
```
wardi: cool, I never new about utf-8-sig

ckanext/canada/commands.py
```diff
@@ -439,6 +445,72 @@ def changed_datasets(self, since_date):
...
+        csv_reader = csv.DictReader((l.encode('utf-8') for l in csv_file))
+        for row in csv_reader:
+            id = row['uuid']
+            if id not in existing_suggestions:
+                # add record
+                record = {
+                        "type": "prop",
```
wardi: only need a single 4-space indent here

wardi:
otherwise looks good!

PR #1067, opened 2020-04-16
Ga events login

RabiaSajjad:


wardi:
looks fine for staging

ckanext/canada/controller.py
```diff
@@ -378,10 +378,11 @@ def logged_in(self):
                 action='home',
                 locale=lang)
         else:
-            h.flash_error(_('Login failed. Bad username or password.'))
+            error_summary = _('Login failed. Bad username or password.')
             return h.redirect_to(
                 controller='user',
-                action='login', locale=lang
+                action='login', locale=lang,
+                error=error_summary,
```
wardi, line 385: how about
```python
    return render('user/login.html', extra_vars={'error_summary':error_summary})
```
instead of this redirect?

ckanext/canada/templates/internal/base.html
```diff
@@ -12,17 +12,9 @@
...
-    <script async src="https://www.googletagmanager.com/gtag/js?id={{ h.googleanalytics_id() }}"></script>
-    <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-
-      gtag('config', '{{ h.googleanalytics_id() }}');
-    </script>
+  {{ super() }}
 {%- endblock -%}
```
wardi, line 16: better to remove the block if you're just calling `{{ super() }}`

ckanext/canada/templates/internal/recombinant/snippets/xls_base.html
```diff
@@ -0,0 +1,30 @@
...
+      </div>
+    </div>
+  </div>
+  {% block upload_submit %}
+  {% endblock %}
+  <div class="col-md-12">
+    {% block messages %}
+    {% endblock %}
+  </div>
+</form>
```
wardi, line 30: this will shadow the complete template in recombinant, hiding all changes there.

wardi: instead you can add a block where you need it in recombinant's template and just override that part

PR #1073, opened 2020-04-29
OPN-109: Update the GA scripts

thriuin:
OPN-109: Update the GA scripts to match what's on production and to use environment variables in place of hard-coded paths.

### DIFF
--- a/bin/og_analytics.py
+++ b/bin/og_analytics.py
@@ -288,8 +288,8 @@ class DatasetDownload():
             for [url, count] in data:
                 id = url.split('/')[-1]
                 id = id.split('&')[0]
-		# added by Rabia 2020-01-29 - removes querystring parameters
-		id = id.split('?')[0]
+                # added by Rabia 2020-01-29 - removes querystring parameters
+                id = id.split('?')[0]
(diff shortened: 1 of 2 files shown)


bin/og_analytics.py
```diff
@@ -267,6 +288,8 @@ def getStats(self, start_date, end_date, og_type):
             for [url, count] in data:
                 id = url.split('/')[-1]
                 id = id.split('&')[0]
+		# added by Rabia 2020-01-29 - removes querystring parameters
+		id = id.split('?')[0]
```
wardi, line 292: tabs instead of spaces?

thriuin:
Hi @wardi , do you want to review this or should I just merge it? Thanks

wardi:
while we're here should we update the openpyxl usage to be compatible with the same version we use in the registry virtualenv?

bin/og_analytics.py
thriuin: Yes, I didn't notice on my local dev since it happens to line up and script ran. Rest of script uses spaces, so I am going to replace tabs with spaces

PR #1075, opened 2020-04-29
OPN-106 Add new OG  licences

thriuin:
Added new open government licenses for provinces and territories of Québec, New Brunswick, and Yukon.


ckanext/canada/public/static/licenses.json
```diff
@@ -163,5 +163,50 @@
...
+    "title_fra": "Licence du gouvernement ouvert – Yukon",
+    "url": "https://open.yukon.ca/data/open-government-licence-yukon",
+    "url_fra": "https://open.yukon.ca/data/fr/gouvernement-ouvert-licence-du-yukon"
+  },
+  {
+    "domain_content": true,
+    "domain_data": true,
+    "domain_software": false,
+    "family": "",
+    "id": "qc-oglqc",
```
wardi, line 202: why not use the standard code for this one instead? Have they modified the terms from cc-by?

thriuin: If this one is cc-by and attributes Quebec, will it work if another province wants the CC with another attribution?

PR #1077, opened 2020-05-05
filter_contractsa.py

maxFredenburgh:


wardi:
thanks. This needs to be called from the the bin/pd/Makefile as well

maxFredenburgh:
> thanks. This needs to be called from the the bin/pd/Makefile as well

Changes made

PR #1079, opened 2020-05-12
broken resource url/file size scripts

maxFredenburgh:


bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[ ]:
```
wardi, line 4: ?

wardi: oh did you paste this in from a jupyter notebook?

maxFredenburgh: Yes I wrote it in a notebook and exported from there

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+import pandas as pd
```
wardi, line 7: pandas isn't the best tool for processing these jsonl/csv files. Pandas loads the complete file into memory instead of streaming it. It's better to use standard python csv/json modules. Streaming lines also lets you divide work evenly across the workers instead of breaking into groups at the beginning and hoping that all groups complete at about the same time.

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+    Manages processes
+    tasks - list of methods to be run
+    """
+    running_tasks = [Process(target=task) for task in tasks]
+    for running_task in running_tasks:
+        running_task.start()
+    for running_task in running_tasks:
+        running_task.join()
+        
+def main(filename,processes=4):
```
wardi, line 28: these jobs are going to be network bound not CPU bound, I bet you can run a lot more than 4 at a time.

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Scan url's of resources from dataset metadata obtained from https://open.canada.ca/data/en
+    CKAN API - ckanapi dump datasets --all -O datasets.jsonl.gz -z -p 4 -r http://localhost
+    
+    filename - destination of jsonl/jl file
```
wardi, line 33: we shouldn't name our jsonl file `something.jl`, that's a mistake I haven't had time to fix (it conflicts with the Julia programming language extension) please just call the format jsonl.

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+    CKAN API - ckanapi dump datasets --all -O datasets.jsonl.gz -z -p 4 -r http://localhost
+    
+    filename - destination of jsonl/jl file 
+    processes - default processes=4. Number of batches and processes to run. Will export 1 file per process.
+    
+    Exports broken links found in metadata
+    """
+    try:
+        df=pd.read_json(filename,lines=True)
+    except:
```
wardi, line 40: bare excepts are a bad idea, they catch things like KeyboardInterrupt which is not what you expect. Always choose an exception class. In this case I'd just let the exception be raised, this is a simple script only used by us, no need to hide the actual error when there is a failure

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+        print('Error: .jsonl/.jl file expected')
+        return False
+    file_length = len(df)
+    batchsize=round(file_length/processes)
```
wardi, line 44: if this rounds down you're going to skip as many as (processes-1) lines from the end of the file

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+# coding: utf-8
+
+# In[ ]:
+
+
+import pandas as pd
+import urllib
+import sys
+from sys import stderr, stdout, argv
+from multiprocessing import Process
```
wardi, line 11: these aren't CPU-bound tasks, an async approach instead of multiprocessing would be a lot better

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+    for i in range(lower,upper):
+        resources = df["resources"][i]
+        error='no-code'
+        for z in range(len(resources)):
+            flag=0
+            if("url" in resources[z]):
+                name=resources[z].get("name")
+                url=resources[z].get("url")
+                try:
+                    c = urllib.request.urlopen(url)
```
wardi, line 79: the requests library handles http more reliably than urllib and is even easier to use

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+                dataset_url = "https://open.canada.ca/data/en/dataset/"+uuid
+                broken_links.append([uuid,title,org,date,dataset_url,name, url,error])
+                
+        percent = float(i) / upper
+        arrow = '-' * int(round(percent * 20)-1) + '>'
+        spaces = ' ' * (20 - len(arrow))
+
+        sys.stdout.write("\rBatch# {4} Percent: [{0}] {1}% -- {2} of {3} checked".format
+                         (arrow + spaces, int(round(percent * 100)),i,upper,batch_number))
+        sys.stdout.flush()
```
wardi, line 104: stderr is preferred for status messages. it's best not to write to stdout/stderr from multiple subprocesses though.

bin/resource_management/resource-broken-links-csv.py
maxFredenburgh: I started out using requests but the info I received from getting headers was inconsistent. Maybe I was implementing it the wrong way?

bin/resource_management/resource-content-info-csv.py
```diff
@@ -0,0 +1,157 @@
...
+                                                   'resource_format','download_format','content_length_kb','date_checked'])
+    export_df.to_csv('file_info_batch{0}.csv'.format(batch_number), index=False)
+    
+
+
+# In[6]:
+
+
+filename,processes = sys.argv[1], sys.argv[2]
+main(filename,processes)
```
wardi, line 150: this won't work, you're passing # processes as a string

bin/resource_management/resource-broken-links-csv.py
```diff
@@ -0,0 +1,128 @@
...
+    export_df = pd.DataFrame(broken_links, columns = ['uuid', 'title','organization','portal_release_date','dataset_url',
+                                                      'resource_name','broken_link','error'])
+    export_df.to_csv('od-broken-links_batch{0}.csv'.format(batch_number), index=False)
+
+
+# In[ ]:
+
+
+filename,processes = sys.argv[1], sys.argv[2]
+main(filename,processes)
```
wardi, line 115: this won't work, you're passing # processes as a string

wardi:
please use vscode, pycharm or some editor that lints your python code for you, there are lots pep8 problems here like spaces missing around `==` and `=`, spaces missing after `,` and unnecessary parenthesis after `if` statements.

Commit messages are important too, just 60 characters or so about what this change does to help make sense of our `git log`s.

bin/resource_management/resource-broken-links-csv.py
wardi: interesting. I thought the requests docs were pretty good. What didn't work about getting the headers?

wardi:
Here's how I think this should work:

1. script that reads urls on stdin and updates a url database (could be a simple csv file) containing:
   - url (primary key)
   - utc timestamp (may be used for removing stale requests and retry in the future)
   - server response
   - file size (if successful)
   - file type (if successful)
   if a url is already present in the database this script won't attempt to fetch it again, this allows repeated runs to refresh the url database and prevents requesting the same url multiple times if present in many resources

   this script can run parallel requests with multiple processes or by using async requests.

2. a script that combines the url database and the od-do-canada json lines file to generate a report on broken links and incorrect file types

3. a script that combines the url database and the od-do-canada json lines file to update file types and sizes in ckan using ckanapi (this should be python2 so it can update directly from the config file instead of over http + api key)

We can feed script #1 with a command like `zcat od-do-canada.jl.gz | jq .resources[].url` there's no need for another script to pull out urls.

bin/resource_management/resource-broken-links-csv.py
maxFredenburgh: Often the http code would be None, and requests also does not work with FTP urls

bin/resource_management/update_resource_format.py
```diff
@@ -0,0 +1,31 @@
...
+argv[1]: site URL
+argv[2]: API Key
+argv[3]: CSV Report ('incorrect_file_types_report.csv')
+
+"""
+
+source = ckanapi.RemoteCKAN(sys.argv[1],apikey=sys.argv[2])
+file_report = sys.argv[3]
+file = open(file_report, "r")
+reader = csv.reader(file)
```
wardi, line 19: try https://docs.python.org/3/library/csv.html#csv.DictReader to make the code below more readable

bin/resource_management/update_resource_format.py
```diff
@@ -0,0 +1,31 @@
+import sys
```
wardi, line 1: this is a script so mark it executable and add a `#!/usr/bin/env python3` line to the top

bin/resource_management/update_resource_format.py
```diff
@@ -0,0 +1,31 @@
+import sys
+import csv
+import ckanapi
+"""
```
wardi, line 4: docstrings need to be the first statement in a module to be accessible with `__doc__`

bin/resource_management/update_resource_format.py
```diff
@@ -0,0 +1,31 @@
...
+reader = csv.reader(file)
+# skip head
+next(reader)
+for line in reader:
+    uuid = line[4]
+    resource_id = line[5]
+    new_format = unicode(line[6].upper())
+    old_format = line[7]
+    print(uuid)
+    source.action.resource_patch(id=resource_id,format=new_format)
```
wardi, line 28: pep8: missing space after comma

bin/resource_management/url_database.py
```diff
@@ -0,0 +1,94 @@
...
+    line = json.loads(dataset)
+    resources = line["resources"]
+    for l in range(len(resources)):
+        url = resources[l]["url"]
+        if url in urls:
+            continue;
+        else:
+            urls.append(url)
+            now = datetime.now()
+            dt_string = now.strftime("%d/%m/%Y %H:%M")
```
wardi, line 40: `YYYY-MM-DD hh:mm:ss` always. see ISO 8601 and our own TBS data policy standards

bin/resource_management/url_database.py
```diff
@@ -0,0 +1,94 @@
...
+            date.append(dt_string)
+            batch_urls.append(url)
+            if len(batch_urls)==batch_size:
```
wardi, line 43: pep8: missing spaces around `==`

bin/resource_management/url_database.py
```diff
@@ -0,0 +1,94 @@
...
+urls = []
+batch_urls = []
+responses = []
+date = []
+
+
+print("Starting...")
+print("Reading and testing URL's")
+
+for dataset in fileinput.input():
```
wardi, line 30: `for i, dataset in enumerate(fileinput.input(), 1):`
then you don't need to initialize and update the line counter yourself

bin/resource_management/url_database.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            urls.append(url)
+            now = datetime.now()
+            dt_string = now.strftime("%d/%m/%Y %H:%M")
+            date.append(dt_string)
+            batch_urls.append(url)
+            if len(batch_urls)==batch_size:
+                sys.stderr.write("\r")
+                sys.stderr.write("Testing Datasets {0} - {1}".format(prev_i, i))
+                prev_i=i
```
wardi, line 46: pep8: missing spaces around `=` when used as an assignment (no spaces when used for keyword args)

bin/resource_management/url_database.py
```diff
@@ -0,0 +1,94 @@
...
+Arguments:
+fileinput - metadata file to be read ('od-do-canada.jsonl')
+batch_size - maximum number of URL's to test in parallel
+"""
+
+batch_size = int(sys.argv[2])
+
+i = 1
+prev_i = 1
+urls = []
```
wardi, line 21: use a `set` to avoid O(N²) performance of `url in urls` below

bin/resource_management/url_database.py
```diff
@@ -0,0 +1,94 @@
...
+
+print("Fetching content data...")
+
+responses=sum(responses,[])
+content_lengths=[]
+content_types=[]
+
+
+for z in range(len(responses)):
+    r=responses[z]
```
wardi, line 66: `for z, x in enumerate(responses)`

bin/resource_management/url_metadata.py
```diff
@@ -0,0 +1,95 @@
...
+            data = broken_links.pop(file_url)
+            broken_links_data.append([file_url,data[0],data[1],
+                                line["organization"]["title"].encode('utf-8'),line["title"].encode('utf-8'),
+                                      line["id"],resources[l]["id"]])
+            if len(broken_links)==0:
+                broken_links_flag=1
+            continue;
+        if file_url in file_types:
+            data = file_types.pop(file_url)
+            if resources[l]["format"].lower() != data[1]:
```
wardi, line 66: content types returned from a web server aren't going to match file extensions, so won't this always fail?

ckanext/canada/commands.py
```diff
@@ -439,6 +443,26 @@ def changed_datasets(self, since_date):
...
+            size = unicode(line[8])
+
+            #portal.action.resource_patch(id=resource_id,format=new_format)
+            resource = portal.call_action('resource_patch',
+                                          {'id':resource_id, 'format':new_format}
+                                          )
+            if 'size' in resource:
+                resource = portal.call_action('resource_patch',
+                                              {'id': resource_id, 'size': size}
+                                              )
```
wardi, line 464: no need for the `if 'size' in resource` check, just add size to the dict above

wardi:
looking good just a few comments. Please assume each comment applies to all files below it (I didn't bother repeating at each location)

maxFredenburgh:
> the content-type returned from a web server is not reliable because it gets ignored by users. Can you collect the file type from the Content-Disposition `filename=...` header instead? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

The headers returned from each request do not contain content-disposition. Here's an example of what is returned:
`{'Content-Length': '968', 'Content-Language': 'en', 'X-AspNet-Version': '4.0.30319', 'Set-Cookie': 'BIGipServer~NRCAN-FGP-PROD~NRCAN-FGP-VIP1.app~NRCAN-FGP-VIP1_pool=rd1o00000000000000000000ffffa72883cco80; path=/; Httponly; Secure', 'X-Powered-By': 'ASP.NET', 'Server': 'Microsoft-IIS/8.5, ', 'Connection': 'close', 'Cache-Control': 'private', 'Date': 'Thu, 11 Jun 2020 16:27:40 GMT', 'Content-Type': 'text/html;charset=utf-8'}`

PR #1080, opened 2020-05-14
OPN-55 changing style of resource format tags

HoussamBedja:


wardi:
this looks the same as the block in `canada/templates/public/snippets/package_item.html` or am I missing something?

HoussamBedja:
Yes but it seems like you (or whoever wrote the code before) did a super.super() to skip `canada/templates/public/snippets/package_item.html` in order to "strip the map cart buttons". But you lost the styling by skipping that file. 

Another way of doing this could be by not skipping the file with super.super(), and instead wrapping the 'add to map cart' logic in a new block, which will be set to being an empty block in `internal/admin/snippets/package_item.html`

wardi:
I mean we could change that line to `{{ super() }}` to avoid copying a bunch of code because the fgp flag would never be set on registry datasets

HoussamBedja:
Oh okay, the original code had a "strip map cart buttons" comment so I assumed it was being affected even on registry. Will apply those changes then.

PR #1084, opened 2020-05-19
foresee: add survey js to portal

RabiaSajjad:


ckanext/canada/templates/internal/base.html
```diff
@@ -13,4 +13,6 @@
 
 {%- block adobe_analytics -%}{%- endblock -%}
 
-{%- block adobe_analytics_meta -%}{%- endblock -%}
\ No newline at end of file
+{%- block adobe_analytics_meta -%}{%- endblock -%}
```
wardi, line 16: why are we adding forsee to the registry?

wardi: oh I get it, you're hiding the block with this

ckanext/canada/templates/public/base.html
```diff
@@ -37,7 +37,7 @@
 {%- endblock -%}
 
 {% block feedback_survey %}
-  <script async defer type="text/javascript" src="{{ h.url_for_static('/static/js/foresee_survey.js') }}" ></script>
+  <script src="/themes/custom/gcweb/js/foresee-survey.min.js" defer="" async=""></script>
```
wardi, line 40: how about adding a comment here like `{# include forsee survey js from drupal hosted on the same domain #}` because we generally don't hard-code url paths into our templates

RabiaSajjad: how about I define a variable in settings ini file and use it here?

wardi: that sounds even better "canada.survey_js_url" or similar would act as documentation and would make it easy to enable/disable from the config file

ckanext/canada/templates/internal/base.html
RabiaSajjad: no longer need to have an empty block for registry

PR #1086, opened 2020-06-02
OPN-190: Update get_deleted_datasets.py to handle MT language codes

thriuin:
The get_deleted_datasets.py script needs to be able to handle compound language codes like en-t-fr and fr-t-en for the machine translated codes used by harvested datasets.
Original code by @iwardi


thriuin:
Thanks @wardi

PR #1089, opened 2020-06-08
view-on-portal button for published datasets

RabiaSajjad:


wardi:
what do you think about making the portal URL configurable so that the staging registry links to the staging portal etc.

RabiaSajjad:
I would really like that, you prefer the configuration localized to only this button? I see other references of the portal hard-coded.

wardi:
sure it could be a separate ticket to fix all portal url references to use an ini setting if you'd like to merge this one first

PR #1095, opened 2020-06-17
Added a schema and solrconfig file compatible with Solr v8

thriuin:
@RabiaSajjad - this should work with latest version of Solr


wardi:
can you make this PR against canada-v2.8?

thriuin:
No problem, makes sense

PR #1103, opened 2020-06-25
add resource preview functionality

RabiaSajjad:


ckanext/canada/templates/public/macros/canada_read.html
```diff
@@ -82,7 +82,7 @@
 Choose the most appropriate display name for a resource
 #}
 {%- macro resource_display_name(res, client_lang) -%}
-  {%- set name = res.name.fr if client_lang == 'fr' else res.name.en -%}
+  {%- set name = res.name_translated.fr if client_lang == 'fr' else res.name_translated.en -%}
```
wardi, line 85: there's a `h.get_translated` we can use to avoid the if/else

wardi:
looks great @RabiaSajjad love all the lines removed :-)

PR #1107, opened 2020-07-15
Opn 333 new date field

thriuin:
Add a new optional federated_date_modified field to open data records to indicate when the federated data partner modified the original dataset on their portal.


thriuin:
Too messy, I am going to branch from staging instead.

PR #1108, opened 2020-07-15
OPN-333 Add a new federated_date_modified field to the open data schema

thriuin:


thriuin:
Sigh - staging changed separately. Will delete and try again

PR #1109, opened 2020-07-15
OPN-333 Add a new federated_date_modified field to the open data schema

thriuin:


ckanext/canada/schemas/dataset.yaml
```diff
@@ -226,6 +226,7 @@ dataset_fields:
 - preset: canada_frequency
 - preset: canada_date_published
 - preset: canada_date_modified
+- preset: canada_federated_date_modified
```
wardi, line 229: better to add this like and with the other "FGP fields" below. Recommend also adding date validation like the default date preset e.g.
```yaml
- field_name: federated_date_modified
  label: ...
  form_snippet: null
  validators: scheming_required isodate convert_to_json_if_date
```

thriuin: Perfect, thanks.

thriuin:
@wardi could you review again? Thanks

wardi:
@thriuin looks fine to me

thriuin:
Thanks @wardi

PR #1110, opened 2020-07-21
remove duplicate jquery et al scripts

RabiaSajjad:


wardi:
testing this I see `Uncaught ReferenceError: jQuery is not defined` on all the pages and the pd pages are mostly blank

ckanext/canada/plugins.py
```diff
@@ -611,17 +611,23 @@ def datastore_upsert(up_func, context, data_dict):
...
-    act.datastore_activity_create(context,
-                                  {'count':res.get('total', 0),
-                                   'activity_type': 'deleted datastore',
-                                   'resource_id': data_dict['resource_id']}
-                                  )
+
+    # no need to log activity for x-loader initiated deleting,
+    # details are recorded in resource's Upload Log
+    # Update Log is available under Data Store tab for resource
+    if len(context['user']) != 0 and 'resource_id' in data_dict.keys():
```
wardi, line 625: we should do the same in our overridden `datastore_upsert` too.

ckanext/canada/templates/public/snippets/wet-boew/headresources.html
```diff
@@ -1,13 +1,8 @@
-<link rel="stylesheet" type="text/css" href="{{ h.url_for_static('base/vendor/font-awesome/css/font-awesome.css') }}" />
-<!--[if ie 7]><link rel="stylesheet" type="text/css" href="{{ h.url_for_static('/fanstatic/vendor/:version:2015-09-01T11:37:20.20/font-awesome/css/font-awesome-ie7.css') }}" /><![endif]-->
 <!--[if gte IE 9 | !IE ]><!-->
-<link href="{{ h.url_for_wet_theme('/assets/favicon.ico') }}" rel="icon" type="image/x-icon" />
 <link rel="stylesheet" href="{{ h.url_for_wet_theme('/css/theme.min.css') }}" />
 <!--<![endif]-->
 <!--[if lt IE 9]>
-<link href="{{ h.url_for_wet_theme('/assets/favicon.ico') }}" rel="shortcut icon" />
 <link rel="stylesheet" href="{{ h.url_for_wet_theme('/css/ie8-theme.min.css') }}" />
-<script src="{% if h.wet_jquery_offline() == false %}http://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js{% else %}{{ h.url_for_wet('/js/jquery/1.11.1/jquery.min.js') }}{% endif %}"></script>
```
wardi, line 10: you said you kept the extra jquery, did you mean to remove this for IE<9?

ckanext/canada/plugins.py
wardi, line 625: `if not context['user']` is a better way to check for empty strings in python

ckanext/canada/templates/public/snippets/wet-boew/headresources.html
RabiaSajjad: i want to continue loading jquery from `footresources`. This can be safely removed. The version is obsolete and does not load anything.

PR #1112, opened 2020-07-24
OPN-341: bugfix for package url in email

RabiaSajjad:


ckanext/canada/templates/public/scheming/display_snippets/email_with_parameters.html
```diff
@@ -1,6 +1,8 @@
-{{ h.mail_to_with_params(email_address=data[field.field_name],
+{% set page_url = h.portal_url() + h.url_for(controller='package', action='read', id=data['id']) %}
```
wardi, line 1: use `qualified=True` instead of `h.portal_url()`

PR #1113, opened 2020-07-28
Opn 207 update registry home page (Quick links)

HoussamBedja:


ckanext/canada/templates/internal/home/quick_links.html
```diff
@@ -25,49 +30,214 @@ <h2 class="h5"><a href="{{
...
+<!--              <header class="modal-header">-->
+<!--                <h2 class="modal-title">{{ _('Suggested Datasets - Update Status') }}</h2>-->
+<!--              </header>-->
+<!--              <div class="modal-body">-->
+<!--                {{ _('Access the suggested datasets that your organization has received in order to provide a status update.') }}-->
+<!--              </div>-->
+<!--            </section>-->
+<!--          </li>-->
+<!--        </ul>-->
+<!--      </section>-->
```
wardi, line 86: please remove this instead of commenting it out in the HTML

HoussamBedja: This is temporary, until the suggested datasets work is complete. You still prefer that it gets removed?

PR #1118, opened 2020-08-14
OPN-345-346: Status display and removal

maxFredenburgh:
1. Detail/Summary display for older statuses.
2. WET Status removal window


wardi:
please resubmit these changes against the suggested-datasets branch

PR #1123, opened 2020-08-28
portal_update: copy datastore tables and views

RabiaSajjad:


ckanext/canada/commands.py
```diff
@@ -313,7 +316,30 @@ def _changed_packages_since(self, registry, since_time):
...
+                    source_table = registry.call_action('datastore_search',
+                                                        {'id': source_resource['id'], 'limit': 0})
+                    if source_table:
+                        from ckanext.datastore.helpers import datastore_dictionary
+                        import simplejson
+
+                        # fetch all records
+                        source_records = registry.call_action('datastore_search',
+                                                              {'id': source_resource['id'],
+                                                               'limit': source_table['total']})
```
wardi, line 331: pulling the records like this won't work for larger tables (need to implement pagination), it's also pretty slow. Was shelling out to pg_dump too complicated?

ckanext/canada/commands.py
```diff
@@ -313,7 +316,30 @@ def _changed_packages_since(self, registry, since_time):
...
+                            record.pop('_id', None)
+
+                        # add views, data dictionary and datastore values for each resource of the source package
+                        source_package[source_resource['id']] = {
+                            "views": registry.call_action('resource_view_list', {'id': source_resource['id']}),
+                            "data_dict": datastore_dictionary(source_resource['id']),
+                            "datastore_records": simplejson.dumps(source_records['records']),
+                        }
+
+                packages.append(json.dumps(source_package))
```
wardi, line 342:
```suggestion
                            "datastore_records": source_records['records'],
                        }

                packages.append(simplejson.dumps(source_package))
```

wardi: you're double-json encoding the records here

ckanext/canada/commands.py
```diff
@@ -408,6 +434,35 @@ def copy_datasets(self, remote, package_ids=None):
...
                 action = 'updated'
                 portal.action.package_update(**source_pkg)
+                # create datastore table and views for each resource
+                for src_res in source_pkg['resources']:
+                    res_id = src_res['id']
+                    if res_id in source_pkg.keys():
+                        # delete resource and recreate table to capture changes made to columns or data
+                        # it would also avoid appending records to an existing resource table by datastore_create
+                        try:
+                            portal.call_action('datastore_delete', {"id": res_id, "force": True})
```
wardi, line 444: can't you check the resource hash to prevent deleting and reloading data every time there's any change to a dataset? the same dataset appears many times in the activity logs as people are making changes so avoiding extra work here is important

ckanext/canada/commands.py
```diff
@@ -408,6 +434,35 @@ def copy_datasets(self, remote, package_ids=None):
...
+                        except NotFound:
+                            # not an issue, resource does not exist in datastore
+                            pass
+
+                        # create table and load data in target datastore
+                        portal.call_action('datastore_create',
+                                           {"resource_id": res_id,
+                                            "fields": source_pkg[res_id]['data_dict'],
+                                            "records": json.loads(source_pkg[res_id]['datastore_records']),
```
wardi, line 453:
```suggestion
                                            "records": source_pkg[res_id]['datastore_records'],
```

ckanext/canada/commands.py
```diff
@@ -313,7 +316,30 @@ def _changed_packages_since(self, registry, since_time):
...
+                    if source_table:
+                        from ckanext.datastore.helpers import datastore_dictionary
+                        import simplejson
+
+                        # fetch all records
+                        source_records = registry.call_action('datastore_search',
+                                                              {'id': source_resource['id'],
+                                                               'limit': source_table['total']})
+                        for record in source_records['records']:
+                            record.pop('_id', None)
```
wardi, line 333: you can pass fields to datastore_search to exclude the _id column from the response instead of doing the work afterwards

ckanext/canada/commands.py
wardi: if we don't want to shell out to pg_dump for copying the data we could run a manual xloader job after running portal_update instead (launch the jobs worker process but tell it to exit on completion)

PR #1124, opened 2020-08-31
Opn 334 + Opn 362: prompt login at timeout

HoussamBedja:
This Pull request includes work from both tickets OPN-334 and OPN-362.


ckanext/canada/templates/internal/page.html
```diff
@@ -32,7 +32,35 @@
...
+          <span class="text">{{ _('Log in') }}</span>
+        </a>
+      </div>
+    </section>
+    <script type="text/javascript" src="{{ h.url_for_static('/session_timeout.js') }}" ></script>
+    <script>
+      $(document).ready(function() {
+        timeoutPop({{ h.get_timeout_length() }});
+      });
+    </script>
```
wardi, line 63: this section needs to be disabled when there is no timeout configured.

HoussamBedja: Oh true. I disabled the whole block instead (including the html portion).

PR #1127, opened 2020-09-01
OPN-346: WET confirmation on status removal

maxFredenburgh:


ckanext/canada/templates/internal/scheming/snippets/form_field.html
```diff
@@ -0,0 +1,23 @@
+{% ckan_extends %}
+
+{% block removal_popup%}
+  {% set locale = h.dump_json({'content': _('Are you sure you want to delete this status?') }) %}
```
wardi, line 4: Need a translation for this string in the .po file

maxFredenburgh: By adding something like this to canada.po?
`#: ckanext/canada/templates/internal/scheming/snippets/form_field.html
msgid "Are you sure you want to delete this status?"
msgstr "Voulez-vous vraiment supprimer ce statut?"`

wardi: yes like that, you can see an example of how @HoussamBedja did it in #1124

wardi:
@maxFredenburgh suggested-datasets has already been merged to master, do you need another merge here?

maxFredenburgh:
@wardi Should I be merging to master instead?

wardi:
@maxFredenburgh if https://github.com/open-data/ckanext-canada/pull/1127/commits/6e2094ffd66496cfe306192693ddfd66cedf4cc2 is the only change go ahead and cherry-pick that commit to master or make a new PR against master

maxFredenburgh:
@wardi PR for these changes to master https://github.com/open-data/ckanext-canada/pull/1133

PR #1130, opened 2020-09-08
Opn 264 more like this

HoussamBedja:


ckanext/canada/helpers.py
```diff
@@ -194,6 +194,8 @@ def portal_url():
     url = PORTAL_URL_DEFAULT_FR if h.lang() == 'fr' else PORTAL_URL_DEFAULT_EN
     return str(config.get(PORTAL_URL_OPTION, url))
 
+def search_url():
```
wardi, line 197: how about adv_search_url as a name, `search_url` is too generic

ckanext/canada/public/static/js/more_like_this.js
```diff
@@ -0,0 +1,13 @@
...
+    $('#related_pkgs li:gt(4)').remove();
+
+    var listItems = $("#related_pkgs a");
+    listItems.each(function(idx, a) {
+        var text = $(a).text().trim();
+        if (text.length > 180) {
+            $(a).text(text.substring(0,175) + '...');
+        }
+    });
+});
```
wardi, line 13: @thriuin can this be done on the advanced search side? I don't like shortening content in JS on the client side like this.

ckanext/canada/helpers.py
```diff
@@ -194,6 +194,8 @@ def portal_url():
...
+    return str(config.get('ckanext.canada.adv_search_url_fr')) if h.lang() == 'fr' else str(config.get('ckanext.canada.adv_search_url_en'))
```
wardi, line 198: Do we really need separate url configuration for en/fr? The language is passed in the url in the template below, does it not work if the domain is wrong?

ckanext/canada/public/static/js/more_like_this.js
thriuin: Sure, you can download the work to search instead of CKAN ;-) It means CKAN formatting logic is in search - why do you not like it in JS?

ckanext/canada/helpers.py
wardi: right, forgot about the redirects

ckanext/canada/public/static/js/more_like_this.js
wardi: It looks like this code is trimming the text length returned from the more like this entries after they are AJAX'ed in to the page. Isn't it simpler to trim them in the advanced search view and avoid the extra JS here?

thriuin: I suppose - it's not a lot of code either way. It's not a lot of extra text transmitted either I don't think, but I don't feel strongly either way. I can make the change - it's too small to open a ticket over.

ckanext/canada/helpers.py
thriuin: Maybe newer browsers can handle this, but I seem to remember IE not being happy when it happened. I haven't tested in a while though.

ckanext/canada/public/static/js/more_like_this.js
wardi: Thanks Ross, every bit of static JS we can keep out of this extension is going to make it simpler when we're migrating to the new way of serving static files in ckan 2.8/2.9

thriuin: Ah, makes sense for sure. No problem, it's just a matter of adding a filter to the Django template. Changes to search are being deployed to staging now. @HoussamBedja, you can remove this and let search truncate titles to a max of 175 chars.

PR #1143, opened 2020-10-01
Opn 454 update contracts validation

HoussamBedja:
Updating the "required" validation rules for solicitation_procedure and limited_tendering_reason, which also overwrides the existing rules under these 2 fields to only apply them after 1-1-2022.

### DIFF
--- a/ckanext/canada/tables/contracts.yaml
+++ b/ckanext/canada/tables/contracts.yaml
@@ -1083,9 +1083,9 @@ iii. Quand un contrat porte sur plus d’un article économique, il est recomman
 
         Si les options « TC » (Concurrentielle - Traditionnelle), « TN » (non concurrentielle) ou « AC » (préavis d’attribution de contrat) sont sélectionnées et qu’une entente de marché ayant une valeur autre que « XX » (nulle) est sélectionnée, le champ Appel d’offres restreint ne peut avoir la valeur « 0 » ou « 00 » (nulle).
 
-    excel_required_formula: '{contract_date}>=DATE(2019,1,1)'
+    excel_required_formula: '{contract_date}>=DATE(2022,1,1)'
     datastore_type: text
-    excel_error_formula: 'AND(OR({cell}="TC",{cell}="TN",{cell}="AC"),{trade_agreement}<>"XX",OR({limited_tendering_reason}="0",{limited_tendering_reason}="00"))'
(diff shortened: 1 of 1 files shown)


ckanext/canada/tables/contracts.yaml
```diff
@@ -1075,17 +1075,17 @@ iii. Quand un contrat porte sur plus d’un article économique, il est recomman
...
-        Obligatoire pour les marchés conclus après le 2019-01-01
+        Obligatoire pour les marchés conclus après le 2022-01-01
 
         Si les options « TC » (Concurrentielle - Traditionnelle), « TN » (non concurrentielle) ou « AC » (préavis d’attribution de contrat) sont sélectionnées et qu’une entente de marché ayant une valeur autre que « XX » (nulle) est sélectionnée, le champ Appel d’offres restreint ne peut avoir la valeur « 0 » ou « 00 » (nulle).
 
-    excel_required_formula: '{contract_date}>=DATE(2019,1,1)'
+    excel_required_formula: '{contract_date}>=DATE(2022,1,1)'
     datastore_type: text
-    excel_error_formula: 'AND(OR({cell}="TC",{cell}="TN",{cell}="AC"),{trade_agreement}<>"XX",OR({limited_tendering_reason}="0",{limited_tendering_reason}="00"))'
+    excel_error_formula: 'AND({contract_date}>=DATE(2022,1,1),AND(OR({cell}="TC",{cell}="TN",{cell}="AC"),{trade_agreement}<>"XX",OR({limited_tendering_reason}="0",{limited_tendering_reason}="00")))'
```
wardi, line 1088: There's two problems here.
1. `{default_formula}` doesn't appear in this line so the field won't be checked against the controlled list
2. You're changing the validation rule without updating the validation text. Something like:
    > If “TC” (Competitive - Traditional), “TN” (Non-Competitive) or “AC” (Advanced Contract Award Notice) is selected and trade agreement with a value other than “XX” (None) is selected, limited tendering cannot have a value of “0” or “00” (None) for contracts after 2022-01-01.

HoussamBedja: 1. The `{default_formula}` wasn't part of the code initially, which is why i didn't change it. So i guess this was a bug in the first place? I can add it now. 
2. Okay I will update the validation text.

wardi: yes it must have been a bug I didn't see earlier. All of our fields with choices and custom errors should include the `{default_formula}` in them

PR #1144, opened 2020-10-01
OPN-490: fix contracts validation bug

HoussamBedja:
Initial validation rules for solicitation_procedure and limited_tendering_reason are in the condition of having agreement_type as "XX". 
This bug fix adds the condition of an empty field, to consider it the same as "XX".


wardi:
This looks good

PR #1149, opened 2020-10-06
OPN-442: replace status removal popup with simple text message

maxFredenburgh:
Replaces the pop-up window when a status is removed with a simple removal text.
Dependent on https://github.com/ckan/ckanext-scheming/pull/260


maxFredenburgh:
No longer relevant

PR #1154, opened 2020-10-28
Opn 146 test pd upload

HoussamBedja:
In this branch, I have created 3 test cases related to the pd template upload:

1. Uploading an empty file
2. Uploading the wrong file (from a different pd type)
3. Uploading the correct file with no errors.

### DIFF
--- a/ckanext/canada/tests/test_xls_upload.py
+++ b/ckanext/canada/tests/test_xls_upload.py
@@ -1,5 +1,6 @@
 # -*- coding: UTF-8 -*-
 import os
+import tempfile
 from nose.tools import assert_equal, assert_raises
 from ckanapi import LocalCKAN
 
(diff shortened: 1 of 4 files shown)


ckanext/canada/tests/test_xls_upload.py
```diff
@@ -0,0 +1,85 @@
...
+from ckan.tests.factories import Organization
+
+from ckanext.recombinant.tables import get_chromo, get_geno
+from ckanext.recombinant.controller import _process_upload_file
+from ckanext.recombinant.write_excel import excel_template, append_data
+from ckanext.recombinant.errors import  BadExcelData
+from openpyxl import load_workbook, Workbook
+
+
+CURRENT_PATH = os.path.dirname(os.path.abspath(__file__))
```
wardi, line 16: use the `tempfile` module to securely create temporary files/directories instead of using the source directory

HoussamBedja: Thanks! I just implemented that in the new commit.

PR #1157, opened 2020-11-06
fix for OPN-533 organization list bug

RabiaSajjad:


wardi:
looks good. love the -145

PR #1158, opened 2020-11-13
OPN-498: Data warehouse

maxFredenburgh:


bin/warehouse/csv_diff.py
```diff
@@ -0,0 +1,107 @@
+import json
```
wardi, line 1: missing a
`#!/usr/bin/env python3`
at the top and a `chmod +x`?

bin/warehouse/csv_diff.py
```diff
@@ -0,0 +1,107 @@
...
+
+def csv_to_dict(csv_file, primary_keys):
+    csv_dict = {}
+    with open(csv_file) as f:
+        hashcsv = csv.DictReader(codecs.EncodedFile(f, 'utf-8', 'utf-8-sig'), delimiter=",")
+        for row in hashcsv:
+            primary_fields = [str(row[t]) for t in primary_keys]
+            uid = '-'.join(primary_fields)
+            csv_dict[uid] = row
+    return csv_dict
```
wardi, line 24: loading the whole csv into memory is a bad idea as these files will keep getting larger

wardi:
I only see two commits, did you squash all of the earlier work?

bin/warehouse/csv_diff.py
```diff
@@ -0,0 +1,107 @@
...
+    return results, result_keys
+
+
+prev_csv, current_csv, endpoint, datestamp, outfile = sys.argv[1], sys.argv[2], sys.argv[3], sys.argv[4], sys.argv[5]
+
+field_info = requests.get(endpoint, timeout=100, verify=False).json()
+
+
+# Grab the primary key fields from the datatype reference endpoint
+if 'nil' in current_csv:
```
wardi, line 73: the fact that our current two-page types always have a `'nil'` in their second page is a detail that shouldn't appear in this script. Match the name to the resource name instead so this doesn't fail on new types.

bin/warehouse/csv_diff.py
```diff
@@ -0,0 +1,107 @@
...
+result_rows, result_keys = add_metadata_fields(old_csv_dict,new_csv_dict,removed_keys,added_keys,modified_keys,datestamp)
+
+exists_flag = 0
+if os.path.isfile(outfile):
+    exists_flag=1
+if result_rows:
+    print("writing")
+    with open(outfile, 'a') as f:
+        warehouse = csv.DictWriter(f, fieldnames=fieldnames, delimiter=',', restval='')
+        if exists_flag == 0:
```
wardi, line 97:
```suggestion
        if not exists_flag:
```

bin/warehouse/csv_diff.py
```diff
@@ -0,0 +1,107 @@
...
+old_csv_dict = csv_to_dict(prev_csv, pk_fields)
+new_csv_dict = csv_to_dict(current_csv, pk_fields)
+
+removed_keys, added_keys, modified_keys = compare_dicts(old_csv_dict, new_csv_dict)
+
+result_rows, result_keys = add_metadata_fields(old_csv_dict,new_csv_dict,removed_keys,added_keys,modified_keys,datestamp)
+
+exists_flag = 0
+if os.path.isfile(outfile):
+    exists_flag=1
```
wardi, line 92:
```suggestion
exists_flag = os.path.isfile(outfile)
```

bin/warehouse/csv_diff.py
```diff
@@ -0,0 +1,107 @@
...
+if result_rows:
+    print("writing")
+    with open(outfile, 'a') as f:
+        warehouse = csv.DictWriter(f, fieldnames=fieldnames, delimiter=',', restval='')
+        if exists_flag == 0:
+            warehouse.writeheader()
+        for row in result_keys:
+            for key in result_rows[row].keys():
+                if not key in fieldnames:
```
wardi, line 101:
```suggestion
            for key in result_rows[row]:
                if key not in fieldnames:
```

bin/warehouse/csv_diff.py
```diff
@@ -0,0 +1,107 @@
...
+                    result_rows[row].pop(key)
+            warehouse.writerow(result_rows[row])
```
wardi, line 103:
```suggestion
            warehouse.writerow({
                k:v for (k, v) in result_rows[row].items() if k in fieldnames
            })
```

bin/warehouse/data/.gitignore
```diff
@@ -0,0 +1 @@
+*.csv
```
wardi, line 1: your other gitignore file ignores this whole directory

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+from datetime import datetime
+
+fname = sys.argv[1]
+
+if len(sys.argv) == 3:
+    operation = sys.argv[2]
+    if operation != '-a' and operation != '-d':
+        sys.exit("Error: invalid operation value (sys.argv[2]), -d or -a expected")
+else:
+    operation = "-d"
```
wardi, line 23: this is a little odd, if you pass more than 2 arguments it will treat it as `-d`. Check out click for handling option parsing and help text display (or argparse if you want to use one in the stdlib)

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+
+def get_base(tfile):
+    base = os.path.basename(tfile)
+    pd_name = os.path.splitext(os.path.splitext(base)[0])[0]
+    return pd_name
+
+def extract(tfile, dest):
+    fpath = './' + dest
+    tar = tarfile.open(fname + tfile)
+    tar.extractall(path=fpath)
```
wardi, line 40: you can operate on tarfiles directly with the tarfile module, there's no need to extract to a temporary directory.

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+    tar.close()
+    return fpath
+
+def run_migrations(fpath):
+    if not os.path.exists('temp'):
+        os.mkdir('temp')
```
wardi, line 46: always use the tempfile module to create temporary directories. This avoids security and concurrency issues.

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+    for csvfile in os.listdir(fpath):
+        print("Migrating {0} from directory {1}".format(csvfile, fpath))
+        proc = subprocess.Popen(['python', 'migrate_all.py', fpath+'/'+csvfile, 'temp/'+fpath+'m_'+csvfile])
+        proc.wait()
```
wardi, line 50: Check the exit code of your subprocesses so you can exit immediately on failures

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+                             dt_string, outfile])
+    proc.wait()
+
+
+if not os.path.exists('warehouse_reports'):
+    os.mkdir('warehouse_reports')
+
+while tar_array:
+    if tar_array == []:
+        break
```
wardi, line 67:
```suggestion
```
dead code

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+    if prev == '':
```
wardi, line 68:
```suggestion
    if not prev:

```

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+    curr_array = [a for a in csv_array if curr_base in a]
+
+    for i in range(len(prev_array)):
+        now = datetime.now()
+        dt_string = now.strftime("%H:%M:%s")
+        print(dt_string,'\n')
+        pdtype = prev_array[i].split('_')[1].split('.')[0]
+        schema = pdtype
+        if 'nil' in pdtype or 'std' in pdtype:
+            schema = schema.split('-')[0]
```
wardi, line 102:
```suggestion
        schema = schema.split('-')[0]
```
instead of listing the types of suffix we currently use, just take everything before the `-`

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+        csv_diff(prev_array[i], curr_array[i],
+            'http://open.canada.ca/data/en/recombinant-schema/{0}.json'.format(schema),
```
wardi, line 104: https?

bin/warehouse/warehouse_reports/.gitignore
```diff
@@ -0,0 +1 @@
+*.csv
```
wardi, line 1: I think you covered the whole directory already

bin/warehouse/generate_warehouse.py
HoussamBedja: Looking at the tarfile module doc, eventually the files would need to be extracted individually to be worked on anyways, whether it's here or later in the migrate_all.py file. Wouldn't it be better to extract_all at this step? It would be simpler for then running the python subprocess command too at line 73.

bin/warehouse/generate_warehouse.py
wardi: @HoussamBedja could you add this check? `wait` should give you the exit status.

bin/warehouse/csv_diff.py
wardi: @HoussamBedja this was never addressed. The original script from Andrew loaded data into a DB to avoid this problem but Max removed that work.  Restoring that code might warrant a separate ticket that we can push to the backlog for now.

bin/warehouse/csv_diff.py
wardi: This wasn't resolved either. @HoussamBedja it's up to you whether to push this to another ticket.

bin/warehouse/generate_warehouse.py
```diff
@@ -0,0 +1,107 @@
...
+    tar.close()
+    return fpath
+
+def run_migrations(fpath, temp_dir):
+
+    for csvfile in os.listdir(fpath):
+        print("Migrating {0} from directory {1}".format(csvfile, fpath))
+        proc = subprocess.Popen(['python', 'migrate_all.py', fpath+'/'+csvfile, temp_dir+'/'+fpath+'m_'+csvfile])
+        if proc.wait() == 1:
+            sys.exit()
```
wardi, line 48:
```suggestion
        if proc.wait():
            sys.exit(1)

```
is better because the exit code could be any non-zero value and we want to return a failure code from our own script.
same comment for below

bin/warehouse/csv_diff.py
maxFredenburgh: @HoussamBedja @wardi Andrew's code can be found here https://github.com/open-data/ckanext-canada/tree/350b3779e1dc35110110c129ea6ad1b6f70529f8

PR #1161, opened 2021-01-15
Opn 711 new agreement type

HoussamBedja:


ckanext/canada/tables/contracts.yaml
```diff
@@ -660,15 +660,27 @@ iii. Quand un contrat porte sur plus d’un article économique, il est recomman
         en: CFTA / CCFTA / CKFTA
         fr: ALEC / ALECC / ALECCS
       D:
-        lookup: [CETA, CFTA]
+        conditional_lookup:
+        - column: contract_date
+          less_than: '2021-01-01'
+          lookup: [CETA, TCA, CFTA]
+        - lookup: [CETA, CFTA]
```
wardi, line 667: I think this is inverted. TCA is an agreement that applies after 2021-01-01, right?

ckanext/canada/tables/contracts.yaml
```diff
@@ -660,15 +660,27 @@ iii. Quand un contrat porte sur plus d’un article économique, il est recomman
...
         en: CETA / CFTA
```
wardi, line 668: add the new agreement to the list that will be shown in the drop-down here too

wardi:
You need to add TCA to the `choices_lookup` below as well

ckanext/canada/tables/contracts.yaml
HoussamBedja: You're right. I just swapped them.

wardi:
Looks good to me. Is there a reason the PR is still marked as a draft?

HoussamBedja:
> Looks good to me. Is there a reason the PR is still marked as a draft?

Business team wants to test in on staging first + they're not sure yet when the client wants this live.

PR #1173, opened 2021-02-11
Opn 794 ati nil edit form remove

PatLittle:


PatLittle:
- will submit based on the ATI changes branch

PR #1176, opened 2021-02-16
Update ati.yaml

PatLittle:
OK think I got it this time..

### DIFF
--- a/ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
+++ b/ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
@@ -788,8 +788,8 @@ msgstr "Veuillez entrer une année valide"
 msgid "Please enter a month number from 1-12"
 msgstr "Veuillez entrer une valeur comprise entre 1 et 12"
 
-msgid "The number of pages released must not be negative"
-msgtr "Le nombre de pages publiées ne doit pas être négatif"
+msgid "This value must not be negative"
+msgstr "Cette valeur ne doit pas être négatif"
 
 msgid ""
 "IMPORTANT: New or updated inventories / rows will be published on open."
--- a/ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
+++ b/ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
@@ -788,6 +788,9 @@ msgstr "Veuillez entrer une année valide"
 msgid "Please enter a month number from 1-12"
 msgstr "Veuillez entrer une valeur comprise entre 1 et 12"
 
+msgid "The number of pages released must not be negative"
+msgtr "Le nombre de pages publiées ne doit pas être négatif"
+
 msgid ""
 "IMPORTANT: New or updated inventories / rows will be published on open."
 "canada.ca daily. IMSO approval must be obtained and sent to <a href=\"mailto:"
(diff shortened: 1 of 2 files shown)


ckanext/canada/tables/ati.yaml
```diff
@@ -219,6 +222,9 @@ resources:
         IF NEW.month < 1 OR NEW.month > 12 THEN
           errors := errors || ARRAY[['month', 'Please enter a month number from 1-12']];
         END IF;
+        IF NEW.pages < 0 THEN
+          errors := errors || ARRAY[['pages', 'The number of pages released must not be negative']];
```
wardi, line 226: we need a translation for this text in the .po file

ckanext/canada/tables/ati.yaml
```diff
@@ -134,6 +134,9 @@ resources:
     description:
       en: The number of pages released in response to the completed Access to Information request. If no records were released, enter a value of “0”.
       fr: Le nombre de pages publiées en réponse à la demande d’accès à l’information. Si aucun enregistrement n’a été publié, entrez la valeur « 0 ».
+    validation:
+      en: The number of pages released must not be negative
+      fr: Le nombre de pages publiées ne doit pas être négatif
```
wardi, line 139: need to define an `excel_error_formula` so that errors are highlighted in the template. See the month column for an example.

ckanext/canada/tables/ati.yaml
wardi: Bonus points for a unit test covering this custom validator in test_ati.py too

ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
```diff
@@ -788,6 +788,9 @@ msgstr "Veuillez entrer une année valide"
 msgid "Please enter a month number from 1-12"
 msgstr "Veuillez entrer une valeur comprise entre 1 et 12"
 
+msgid "The number of pages released must not be negative"
+msgtr "Le nombre de pages publiées ne doit pas être négatif"
```
wardi, line 792: misspelling here (even github picked it up)

ckanext/canada/tables/ati.yaml
```diff
@@ -134,12 +134,16 @@ resources:
...
     datastore_type: int
     form_required: true
     excel_required: true
     form_attrs:
       size: 10
     excel_column_width: 34
+    excel_error_formula: 'OR({default_formula},{cell}>=0)'
```
wardi, line 146: this says it's an error to have a value >= 0

ckanext/canada/i18n/fr/LC_MESSAGES/canada.po
wardi: Another thought: if you make this error message "This value must not be negative" then we can apply the same message and translation to other fields with the same validation check.

The UI and API will show the field name next to the error, e.g. "Number of Pages: The number of pages released must not be negative" is a little redundant.

ckanext/canada/tables/ati.yaml
```diff
@@ -135,15 +135,15 @@ resources:
...
+      en: This value must not be negative
+      fr: Cette valeur ne doit pas être négatif
     datastore_type: int
     form_required: true
     excel_required: true
     form_attrs:
       size: 10
     excel_column_width: 34
-    excel_error_formula: 'OR({default_formula},{cell}>=0)'
+    excel_error_formula: 'OR({default_formula},{cell}<=0)'
```
wardi, line 146: 0 is a valid number of pages

PR #1179, opened 2021-03-04
format error page based on error code

RabiaSajjad:


ckanext/canada/templates/public/error_document_template.html
```diff
@@ -2,20 +2,27 @@
 
 {% block page_heading %}
   {% block subtitle %}
-    {% if c.code[0] == '404' %}
-      {{ _('We couldn\'t find that Web page') }}
-    {% else %}
-      {{ _('We encountered an error') }}
-    {% endif %}
+    {{ _("Open Government Registry") }}
```
wardi, line 5: this will appear on the portal too, what is this change for?

wardi: If you just want to remove the custom title removing the `{% block page_heading %}` would be better, right?

PR #1181, opened 2021-03-11
Bugfixes 2.8

RabiaSajjad:
OPN-748, OPN-537 and some other bugfixes


ckanext/canada/templates/internal/package/read.html
```diff
@@ -1,6 +1,21 @@
 {% ckan_extends %}
 
-{% block package_comments %}{% endblock %}
+{% block internal_actions %}
+  <section class="resources module-content indent-large">
+    {% if pkg.portal_release_date %}
```
wardi, line 5: technically this should check if portal_release_date is in the past, but this is close enough

wardi:
looks good @RabiaSajjad

PR #1183, opened 2021-03-29
extend portal-update command to include views

RabiaSajjad:


ckanext/canada/commands.py
```diff
@@ -61,6 +61,7 @@ class CanadaCommand(CkanCommand):
         paster canada portal-update <portal.ini>
```
wardi, line 61:
```suggestion
        paster canada portal-update <portal.ini> -u <user>
```
I would put it here instead and not use `[` `]` because it's not an optional parameter

ckanext/canada/logic.py
```diff
@@ -123,19 +124,27 @@ def activity_list_from_user_since(context, data_dict):
...
+    'deleted resource view' activities for the whole site.
 
     '''
-    q = model.activity._changed_packages_activity_query()
-    q = q.order_by(model.Activity.timestamp)
+    q = model.Session.query(model.Activity.object_id.label('object_id'),
+                            func.max(model.Activity.timestamp).label(
+                                'timestamp'))
+    q = q.filter(or_(model.Activity.activity_type.endswith('package'),
+                     model.Activity.activity_type.endswith('package')))
```
wardi, line 141: endswith(package) or endswith(package)?

ckanext/canada/logic.py
```diff
@@ -123,19 +124,27 @@ def activity_list_from_user_since(context, data_dict):
...
-    'deleted package' activities for the whole site.
+    This activity stream includes recent 'new package', 'changed package',
+    'deleted package', 'new resource view', 'changed resource view' and
+    'deleted resource view' activities for the whole site.
 
     '''
-    q = model.activity._changed_packages_activity_query()
-    q = q.order_by(model.Activity.timestamp)
+    q = model.Session.query(model.Activity.object_id.label('object_id'),
+                            func.max(model.Activity.timestamp).label(
```
wardi, line 138: you're taking the max timestamp but you need the min in order to paginate, right?

ckanext/canada/logic.py
```diff
@@ -70,11 +72,11 @@ def wrapper(context, data_dict,
 
 @side_effect_free
 def changed_packages_activity_list_since(context, data_dict):
-    '''Return the activity stream of all recently added or changed packages.
+    '''Return the package_id and timestamp of all recently added or changed packages.
 
     :param since_time: starting date/time
 
-    Limited to 31 records (configurable via the
+    Limited to 63 records (configurable via the
```
wardi, line 79: this query is only returning unique package ids and timestamps now. I think we can make this limit much, much larger (I'd guess in the thousands). Would be interesting to experiment with our live site and large time ranges to choose a better values.

ckanext/canada/logic.py
```diff
@@ -70,11 +72,11 @@ def wrapper(context, data_dict,
 
 @side_effect_free
 def changed_packages_activity_list_since(context, data_dict):
```
wardi, line 74: better to rename this since it's not returning an activity list anymore

ckanext/canada/logic.py
```diff
@@ -90,9 +92,8 @@ def changed_packages_activity_list_since(context, data_dict):
     # hard limit this api to reduce opportunity for abuse
     limit = int(config.get('ckan.activity_list_hard_limit', 63))
 
-    activity_objects = _changed_packages_activity_list_since(
-        since_time, limit)
-    return model_dictize.activity_list_dictize(activity_objects, context)
+    return _changed_packages_activity_list_since(since_time, limit)
```
wardi, line 95: actions need to return values that can be serialized to json, e.g. lists

ckanext/canada/logic.py
```diff
@@ -90,9 +89,14 @@ def changed_packages_activity_list_since(context, data_dict):
         raise ValidationError({'since_time':e.error})
 
     # hard limit this api to reduce opportunity for abuse
-    limit = int(config.get('ckan.activity_list_hard_limit', 63))
+    limit = int(config.get('ckan.activity_list_hard_limit', 10000))
```
wardi, line 92:
```suggestion
    limit = int(config.get('ckan.activity_timestamp_since_limit', 10000))
```

ckanext/canada/templates/internal/package/read.html
```diff
@@ -10,8 +10,11 @@
       </a>
     {% endif %}
     {% if pkg.state == "deleted" %}
-    <div class="module-alert alert alert-danger">
+    <div class="module-alert alert alert-danger mrgn-tp-sm">
       <p>{{ _('This dataset has been deleted and is no longer accessible to non-admin users') }}.</p>
+      {% if h.check_access('package_update', {'id': pkg.id}) %}
+        {% link_for _('Undelete'), controller='ckanext.canada.controller:CanadaController', action='package_undelete', pkg_id=pkg.name, class_='btn btn-default' %}
+      {% endif %}
```
wardi, line 17: this feature could be confusing, the data and views attached to a deleted record could have been permanently removed. Is there a reason you're adding UI for it in this PR?

wardi:
LGTM

PR #1194, opened 2021-06-22
hide 2fa from user_edit form if totp is disabled

RabiaSajjad:


wardi:
don't we need to call this helper in the form template?

RabiaSajjad:
> don't we need to call this helper in the form template?

Yes, we do https://github.com/open-data/ckanext-security/commit/6b667fa819f0704896d3e282e979fbcd2910e172

PR #1206, opened 2021-07-09
Update ga_script.sh

PatLittle:
fixing the TMPFILE location


bin/ga_script.sh
```diff
@@ -120,7 +120,7 @@ python $GA_OG_ANALYTICS \
     $STARTYR-$CURMTH-01 $PRVYR-$PRVMTH-$LASTDY visit \
     >> $GA_LOG_FILE 2>&1
 mv $GA_TMP_DIR/od_ga_downloads.xls $GA_ARCHIVE_DIR/visits-$PRVMTH$STARTYR-$PRVMTH$PRVYR.xls
-TMPFILE=$GA_TMP_DIR/visits-$PRVMTH$STARTYR-$PRVMTH$PRVYR.xls
```
wardi, line 123: if we're making a change here, might as well change the variable name from TMPFILE to something more appropriate. UPLOADFILE maybe?

wardi: I ask for this change because otherwise I look at this line and think "why on earth are we using our archive directory for a temporary file??"

PatLittle: updated with that change.

PR #1207, opened 2021-07-09
add archive

PatLittle:
adding the archive zip.
I tested this on the server by running this section line by line - worked

### DIFF
--- a/bin/ga_script.sh
+++ b/bin/ga_script.sh
@@ -142,3 +142,12 @@ $GA_PORTAL_CKANAPI action resource_patch -c $GA_PORTAL_INI \
 	id=d22d2aca-155b-4978-b5c1-1d39837e1993 upload@"${UPLOADFILE}"
 echo 'deleted datasets done' >> "$GA_LOG_FILE"
 
+#upload archive
+cd $GA_ARCHIVE_ROOT
+cd ..
+zip -r $GA_TMP_DIR/archive.zip analytics
+UPLOADFILE=$GA_TMP_DIR/archive.zip
+$GA_PORTAL_CKANAPI action resource_patch -c $GA_PORTAL_INI \
+	id=8debb421-e9cb-49de-98b0-6ce0f421597b upload@"${UPLOADFILE}"
+echo 'archived updates done' >> "$GA_LOG_FILE"
+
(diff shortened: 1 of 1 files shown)


PatLittle:
https://open.canada.ca/data/en/dataset/2916fad5-ebcc-4c86-b0f3-4f619b29f412/resource/8debb421-e9cb-49de-98b0-6ce0f421597b

bin/ga_script.sh
```diff
@@ -142,3 +142,12 @@ $GA_PORTAL_CKANAPI action resource_patch -c $GA_PORTAL_INI \
 	id=d22d2aca-155b-4978-b5c1-1d39837e1993 upload@"${UPLOADFILE}"
 echo 'deleted datasets done' >> "$GA_LOG_FILE"
 
+#upload archive
+cd $GA_ARCHIVE_ROOT
+cd ..
+zip -r $GA_TMP_DIR/archive.zip analytics
```
wardi, line 148: for safety we can add a pattern to the files that we're adding to the archive. don't want to accidentally publish something that was left in this directory that we're not meaning to
```suggestion
zip -r $GA_TMP_DIR/archive.zip analytics/????-??-??/openDataPortal.siteAnalytics.*.{csv,xls}
```

bin/ga_script.sh
```diff
@@ -129,8 +129,8 @@ python $GA_OG_ANALYTICS \
     $GA_CLIENT_SECRET 68455797 $GA_PORTAL_INI \
     $STARTYR-$CURMTH-01 $PRVYR-$PRVMTH-$LASTDY download \
     >> $GA_LOG_FILE 2>&1
-mv $GA_TMP_DIR/od_ga_downloads.xls $GA_ARCHIVE_DIR/downloads-$PRVMTH$STARTYR-$PRVMTH$PRVYR.xls
-UPLOADFILE=$GA_ARCHIVE_DIR/downloads-$PRVMTH$STARTYR-$PRVMTH$PRVYR.xls
+mv $GA_TMP_DIR/od_ga_downloads.xls $GA_ARCHIVE_DIR/openDataPortal.siteAnalytics.downloads-$PRVMTH$STARTYR-$PRVMTH$PRVYR.xls
```
PatLittle, line 132: @wardi I think if I add that prefix pattern to the 2 xls files your suggestion would work

wardi: this won't work unless we rename all the old files too, right?

PR #1217, opened 2021-09-09
bugfix for csrf authentication error on delete org

RabiaSajjad:


wardi:
Is a translation required for this new string?

PR #1220, opened 2021-10-15
add schema for organization

RabiaSajjad:


ckanext/canada/schemas/organization.yaml
```diff
@@ -0,0 +1,145 @@
...
+    fr:
+      en: The French name by which the organization is known
+      fr: pending translation
+  required: true
+  # copied from fluent_text preset
+  form_snippet: fluent_text.html
+  display_snippet: fluent_text.html
+  error_snippet: fluent_text.html
+  validators: fluent_text unicode scheming_required
+  output_validators: fluent_core_translated_output
```
wardi, line 42: we need a new version of this validator. `fluent_core_translated_output` copies the system default language into the `title` field but we want the `En | Fr` copied there instead

ckanext/canada/schemas/organization.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+# Field = Description.
+# {The organization's description}
+- field_name: notes_translated
```
wardi, line 46: I believe the core organization field for description is called `description` not `notes`. Also, do we need descriptions for our organizations at all?

ckanext/canada/schemas/organization.yaml
```diff
@@ -0,0 +1,145 @@
...
+  # copied from fluent_markdown preset
+  form_snippet: fluent_markdown.html
+  display_snippet: fluent_markdown.html
+  error_snippet: fluent_text.html
+  validators: fluent_text
+  output_validators: fluent_core_translated_output
+
+# Field = Short Form.
+# {The organization's short form}
+- field_name: shortform_translated
```
wardi, line 73: this field can be called `shortform`. the `_translated` field names were added as a way to add fluent fields that override core fields, and there's no core field called `shortform`

ckanext/canada/schemas/organization.yaml
```diff
@@ -0,0 +1,145 @@
...
+    fr:
+      en: The French abbreviation by which the organization is known
+      fr: pending translation
+  required: true
+  # copied from fluent_text preset
+  form_snippet: fluent_text.html
+  display_snippet: fluent_text.html
+  error_snippet: fluent_text.html
+  validators: fluent_text scheming_required
+  output_validators: fluent_core_translated_output
```
wardi, line 97: `fluent_text_output` can be used here. `shortform` isn't a core field

ckanext/canada/schemas/organization.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+# Field = URL.
+# {The organization's URL}
+- field_name: name
+  label: URL
+  help_text:
+    en: The department identifier used for linking to the department e.g. "tbs-sct" is used to create URL registry.open.canada.ca/en/organization/tbs-sct
```
wardi, line 104: please add `statcan` as an example for a department using a common shortform for both languages

wardi:
looks good!

bin/migrate-org-to-scheming.py
```diff
@@ -0,0 +1,64 @@
...
+                                     {'all_fields': True})
+    for org in orgs:
+        if org.get('title'):
+            title_translated = org[u'title'].split(' | ')
+            org[u'title_translated'] = {
+                u'en': title_translated[0],
+                u'fr': title_translated[1]
+                if len(title_translated) > 1 else '',
+            }
+        if org.get('extras') and \
```
wardi, line 52: usually we would want to loop through the extras in case there's more than one. Do we only ever have a single extra?

bin/migrate-org-to-scheming.py
```diff
@@ -0,0 +1,64 @@
...
+              '-o/--output outfile.jsonl'
+        sys.exit(1)
+
+    output = codecs.open(outfile, 'w', encoding='utf-8')
+    ckan_instance = RemoteCKAN(server)
+    orgs = ckan_instance.call_action('organization_list',
+                                     {'all_fields': True})
+    for org in orgs:
+        if org.get('title'):
+            title_translated = org[u'title'].split(' | ')
```
wardi, line 46: I wonder if we have some that weren't entered exactly like this. If so maybe splitting on `|` and stripping whitespace would be safer. Are there any orgs with no translations?

bin/migrate-org-to-scheming.py
```diff
@@ -0,0 +1,64 @@
...
+            org[u'title_translated'] = {
+                u'en': title_translated[0],
+                u'fr': title_translated[1]
+                if len(title_translated) > 1 else '',
```
wardi, line 50: if there's no French translation it's better to not have the `fr` key at all otherwise it will appear as blank on the French side of the site

bin/migrate-org-to-scheming.py
RabiaSajjad: https://github.com/open-data/ckanext-canada/blob/68e41fd6af164632b53f67cd4cf6109b879e0f39/transitional_orgs.jsonl#L110

ckanext/canada/validators.py
```diff
@@ -270,19 +270,19 @@ def canada_org_title_translated_output(key, data, errors, context):
...
-def protect_is_federal(key, data, errors, context):
+def protect_reporting_requirements(key, data, errors, context):
     """
-    Ensure the is_federal flag is not changed by an unauthorized user.
+    Ensure the reporting_requirements field is not changed by an unauthorized user.
     """
     if not is_sysadmin(context['user']):
-        errors[key].append("is_federal field can only be changed by sysadmin")
+        errors[key].append("reporting_requirements field can only be changed by sysadmin")
         raise StopOnError
```
wardi, line 279: I think you need to replace the value passed with the existing value when a user isn't sysadmin instead of returning an error, otherwise regular users won't be able to edit their own info (full name etc.)

wardi: or maybe only return an error if the value submitted doesn't match the existing one for non-sysadmins

ckanext/canada/validators.py
```diff
@@ -274,8 +274,24 @@ def protect_reporting_requirements(key, data, errors, context):
...
     """
-    if not is_sysadmin(context['user']):
-        errors[key].append("reporting_requirements field can only be changed by sysadmin")
+    if is_sysadmin(context['user']):
+        return
+
+    original = ''
+    org = context.get('group')
+    if org:
+        original = org.extras.get('reporting_requirements', '')
```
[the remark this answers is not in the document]
RabiaSajjad: yeah

PR #1225, opened 2021-11-04
Open 1685 supplier vax

PatLittle:
- I got this working to the point where I was able to generate the template, and in registry-staging it gives the "get started" screen, but when I click get started I'm getting the error page.

### DIFF
--- a/ckanext/canada/plugins.py
+++ b/ckanext/canada/plugins.py
@@ -309,6 +309,7 @@ ckanext.canada:tables/service.yaml
 ckanext.canada:tables/dac.yaml
 ckanext.canada:tables/nap.yaml
 ckanext.canada:tables/experiment.yaml
+ckanext.canada:tables/suppliervax.yaml
 """
         config['ckan.search.show_all_types'] = True
         config['search.facets.limit'] = 200  # because org list
@@ -322,6 +323,7 @@ ckanext.canada:schemas/dataset.yaml
 ckanext.canada:schemas/info.yaml
 ckanext.canada:schemas/prop.yaml
 """
+        config['scheming.organization_schemas'] = 'ckanext.canada:schemas/organization.yaml'
 
         # Enable our custom DCAT profile.
         config['ckanext.dcat.rdf.profile'] = 'canada_dcat'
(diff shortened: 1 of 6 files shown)


ckanext/canada/plugins.py
```diff
@@ -531,6 +533,10 @@ def get_validators(self):
                 validators.user_read_only,
             'user_read_only_json':
                 validators.user_read_only_json,
+            'no_future_date':
+                validators.no_future_date,
+            'canada_sort_prop_status':
+                validators.canada_sort_prop_status,
```
wardi, line 539: these seem to be unrelated changes. You should always create branches from master, even if you intend to merge to staging, to avoid pulling in things like this.

ckanext/canada/tables/suppliervax.yaml
```diff
@@ -0,0 +1,424 @@
...
+
+resources:
+- title: Contract Data - Reporting on COVID-19 Vaccination Requirement for Supplier Personnel
+  resource_name: contract-data
+  create_form: true
+  edit_form: true
+  fields:
+
+  # 3.1
+  - datastore_id: c_report_date
```
wardi, line 27: oh, `c_` is for "contract", I thought you were naming all your columns starting with `c_`

ckanext/canada/tables/suppliervax.yaml
```diff
@@ -0,0 +1,424 @@
...
+        END IF;
+        RAISE EXCEPTION E'TAB-DELIMITED\t%', array_to_string(errors, E'\t');
+      END;
+  - update_record_modified_created_trigger
+
+  examples:
+    record:
+      v_report_date: 2022-02-14
+      v_total: 19
+      v_terminations: 3
```
wardi, line 404: some of the required fields seem to be missing from your examples

ckanext/canada/tables/suppliervax.yaml
```diff
@@ -0,0 +1,424 @@
...
+      additional_information_en: optional additional information goes here
+      additional_information_fr: text
+    filters:
+      c_report_date: 2022-02-14
+    filter_one:
+      c_report_date: 2022-02-14
```
wardi, line 410: this filter should return a single record, how about filtering on v_report_date instead?

PR #1238, opened 2021-12-03
exclude xloader initiated datastore_delete from chained action

RabiaSajjad:


ckanext/canada/plugins.py
```diff
@@ -717,17 +717,20 @@ def datastore_upsert(up_func, context, data_dict):
...
-    lc = ckanapi.LocalCKAN(username=c.user)
-    res_id = data_dict['id'] if 'id' in data_dict.keys() else data_dict['resource_id']
-    res = lc.action.datastore_search(
-        resource_id=res_id,
-        filters=data_dict.get('filters'),
-        limit=1,
-    )
-    result = up_func(context, data_dict)
+    if context.get('agent') == 'xloader':
+        result = up_func(context, data_dict)
```
wardi, line 721: I like to
```suggestion
        return up_func(context, data_dict)
```
then you don't need the whole rest of the function in an else block

wardi:
LGTM

PR #1248, opened 2022-01-27
remove outdated resource from datastore

RabiaSajjad:


ckanext/canada/plugins.py
```diff
@@ -896,3 +904,31 @@ def build_nav_main(*args):
...
+    if h.is_url(res['url']) and h.validation_status(res['id']) != 'unknown':
+        try:
+            p.toolkit.get_action(u'resource_validation_delete')(
+                context, {'resource_id': res['id']})
+            log.info('Validation report deleted for resource %s' % res['id'])
+        except ObjectNotFound:
+            log.error('Validation report for resource %s does not exist' % res['id'])
+
+    # remove outdated datastore table for linked resources
+    if h.is_url(res['url']) and 'datastore_active' in res and res['datastore_active']:
```
wardi, line 928: same here

PR #1249, opened 2022-01-27
Open 1763 admin aircraft

PatLittle:


ckanext/canada/tables/admin_aircraft.yaml
```diff
@@ -0,0 +1,330 @@
...
+      style: "width: 100%; display: block"
+
+# 3.10
+  - datastore_id: passengers
+    label:
+      en: Passenger names
+      fr: Nom des passagers
+    description:
+      en: Full list of flight passengers.
+      fr: Liste complète des passagers du vol.
```
wardi, line 199: You're not describing how the passenger list should be separated or sorted (if at all)

ckanext/canada/tables/admin_aircraft.yaml
```diff
@@ -0,0 +1,330 @@
...
+    datastore_type: date
+    form_snippet: scheming/form_snippets/date.html
+
+  # 3.7
+  - datastore_id: locations_en
+    label:
+      en: Location(s) visited (English)
+      fr: Endroit(s) visité (anglais)
+    description:
+      en: To include names of all places visited to conduct government business during the trip in question, in English.
```
wardi, line 138: ... separated by semicolons (;)

wardi:
Is this minister list different than the other list we're generating? If not we should be using the same file.

ckanext/canada/tables/admin_aircraft.yaml
```diff
@@ -0,0 +1,330 @@
+dataset_type: admin_aircraft
+target_dataset: admin_aircraft
+
+title: Proactive Disclosure - Government Administrative Aircraft Use
```
wardi, line 4: A translation for this string needs to appear in the .po file

ckanext/canada/tables/admin_aircraft.yaml
```diff
@@ -0,0 +1,330 @@
+dataset_type: admin_aircraft
+target_dataset: admin_aircraft
```
wardi, line 2: this would be the first type with an underscore in the name. I've been trying to keep these names brief and without underscores to avoid mixing `-` and `_` in a bunch of places which can be confusing.

ckanext/canada/tables/admin_aircraft.yaml
wardi: Is the order of place names significant? Do we want to validate that the same number of places are provided in english and french?

PatLittle:
messed up the branches and ruined the PR, will submit fresh one.

PR #1251, opened 2022-02-09
added https update command

habibayoub:


ckanext/canada/commands.py
```diff
@@ -498,6 +502,37 @@ def resource_size_update(self, size_report):
...
+        for org in data:
+            for res in data[org]:
+                if check_https(res['url'], org, alt_data) and res['url_type'] == 'http':
+                    try:
+                        https_url = res['url'].replace('http://', 'https://')
+                        local_ckan.call_action('resource_show', {'id': res['id']})
+                        local_ckan.call_action('resource_patch',
+                                               {'id': res['id'],
+                                                'url': res['url'].replace(
+                                                    'http://', 'https://')})
```
wardi, line 531:
```suggestion
                                                'url': https_url})
```

ckanext/canada/commands.py
```diff
@@ -498,6 +502,37 @@ def resource_size_update(self, size_report):
...
+                    for url in organization['urls']:
+                        if check_url == url:
+                            return True
+            return False
+
+        local_ckan = LocalCKAN()
+
+        for org in data:
+            for res in data[org]:
+                if check_https(res['url'], org, alt_data) and res['url_type'] == 'http':
```
wardi, line 524: is `url_type` really set to `'http'`?

wardi: oh I thought this referred to the ckan resource `url_type` field. nevermind

wardi:
Some documentation on the formats of the files required by this command should be provided.

Otherwise this looks good

PR #1253, opened 2022-02-17
fresh commit based on master

PatLittle:
Incorporated suggestions from review on #1249 

this is deployed and working on staging.

### DIFF
--- a/bin/filter/filter_adminaircraft.py
+++ b/bin/filter/filter_adminaircraft.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+"filter for adminaircraft.csv"
+
+import csv
+import sys
+
+REMOVE_COLUMNS = [
+    'record_created',
+    'record_modified',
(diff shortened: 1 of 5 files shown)


wardi:
Looks good except that our minister list is still duplicated. @RabiaSajjad what do you think? Does it make sense to have a separate copy of the minister list for administrative aircraft use?

PatLittle:
> Looks good except that our minister list is still duplicated. @RabiaSajjad what do you think? Does it make sense to have a separate copy of the minister list for administrative aircraft use?

The only reason I went with a copy of the list is they had a business requirement to add some additional names that aren't on the main minister list. Right now it is just the Governor General, but it's likely they will need other names in the future, such as former prime ministers. I was thinking of setting up a small script that would take the main ministers list, then add the custom names to it, that way it wouldn't need to be manually kept in sync with the main list. Was thinking a GitHub action.

PR #1258, opened 2022-02-28
add new pd admin aircraft to quick links

RabiaSajjad:


wardi:
We don't want a quick link for this type. Only one org is using it

PR #1259, opened 2022-02-28
Admin aircraft

RabiaSajjad:


ckanext/canada/schemas/presets.yaml
```diff
@@ -97,6 +97,9 @@ presets:
         en: Briefing packages for Parliamentary Committee appearances for deputy heads
         fr: Documents d'information à l'intention des administrateurs généraux pour les comparutions devant un comité parlementaire
       value: parliament_committee_deputy
+    - label:
+        en: Use of Administrative Aircraft
+        fr: Utilisation des avions d'affaires
```
wardi, line 102: Does adminaircraft use a custom collection type? If so it needs a value too:
```suggestion
        fr: Utilisation des avions d'affaires
      value: adminaircraft
```

wardi:
one change looks unrelated/incomplete otherwise this looks good

PR #1261, opened 2022-03-04
Update resource_read.html

habibayoub:


ckanext/canada/templates/public/package/resource_read.html
```diff
@@ -10,3 +10,29 @@
...
+            {% if res.url and h.is_url(res.url) %}
+              <p class="text-muted ellipsis">{{ _('URL:') }} <a class="resource-url-analytics" href="{{ res.url }}" title="{{ res.url }}">{{ res.url }}</a></p>
+            {% elif res.url %}
+              <p class="text-muted break-word">{{ _('URL:') }} {{ res.url }}</p>
+            {% endif %}
+          {% endblock %}
+{% endblock %}
+{% block resource_data_dictionary %}
+       <div class="module-content">
+        <a href="#data-dictionary"><h2 id="data-dictionary">{{ _('Data Dictionary') }}</h2></a>
```
wardi, line 25: If this is where the change is, why not only override the resource_data_dictionary block instead of the whole resource_content block?

wardi: oh, I see. you're removing some of the text above.
in that case one way to simplify the change would be to use `{{ super() }}` inside the sub-blocks that haven't changed from the original

ckanext/canada/templates/public/package/resource_read.html
```diff
@@ -10,3 +10,29 @@
...
 {% block secondary_content %}{% endblock %}
 
+{% block resource_content %}
+          {% block resource_read_title %}<h1 class="page-heading">{{ h.resource_display_name(res) | truncate(50) }}</h1>{% endblock %}
+          {% block resource_read_url %}
+            {% if res.url and h.is_url(res.url) %}
+              <p class="text-muted ellipsis">{{ _('URL:') }} <a class="resource-url-analytics" href="{{ res.url }}" title="{{ res.url }}">{{ res.url }}</a></p>
+            {% elif res.url %}
+              <p class="text-muted break-word">{{ _('URL:') }} {{ res.url }}</p>
+            {% endif %}
```
wardi, line 20:
```suggestion
          {% block resource_read_title %}{{ super() }}{% endblock %}
          {% block resource_read_url %}
            {{ super() }}```

PR #1262, opened 2022-03-15
Added save_button_text block override to the new package form interna…

JVickery-TBS:
…l template.

### DIFF
--- a/ckanext/canada/templates/internal/package/new_package_form.html
+++ b/ckanext/canada/templates/internal/package/new_package_form.html
@@ -1,7 +1,17 @@
 {% extends 'package/snippets/package_form.html' %}
+
 {% set form_style = c.form_style or c.action %}
+
 {% block delete_button %}
   {% if c.pkg_dict and h.check_access('package_delete', {'id': c.pkg_dict.id}) %}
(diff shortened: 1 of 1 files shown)


wardi:
isn't this the same as the block it's overriding?

JVickery-TBS:
@wardi It is overriding this block https://github.com/open-data/ckan/blob/canada-v2.8/ckan/templates/package/snippets/package_form.html#L43

From what I could tell from debugging, if the template exists in the ckanext directory, it will load that one and not the one in the ckan core install. So it seem that `ckanext-canada/ckanext/canada/templates/internal/package/new_package_form.html` is loaded, not `ckan/ckan/templates/package/new_package_form.html`

wardi:
could the problem be that line 1 is extending the wrong template and we should use `{% ckan_extends %}` instead of `{% extends 'package/snippets/package_form.html' %}`

wardi:
or even better can we fix the problem by deleting this template? I don't see the point of the only other block override here

JVickery-TBS:
@wardi Template is deleted and works fine locally.

wardi:
LGTM

PR #1263, opened 2022-03-29
Feature/more api error messages

JVickery-TBS:
Overrode the isodate validator from ckan in ckanext-canada. It is the same method, just has a different error message output because of the issues with the english strings not working in the po files as discussed.

New error message string has been added to the ckanext-canada po files. Official translations will be added in soon.


wardi:
did you branch from staging or something?

PR #1264, opened 2022-03-29
Feature/date format error message

JVickery-TBS:
Overrode the ckan/validators.isodate method in ckanext-canada to add a more helpful error message to ill-formatted dates.

Added english and french strings into the po files.


ckanext/canada/i18n/ckanext-canada.pot
```diff
@@ -2233,3 +2233,5 @@ msgstr ""
 msgid "Remember me"
 msgstr ""
 
+msgid "Date format incorrect. Expecting YYYY-MM-DD"
+msgstr "Date format incorrect. Expecting YYYY-MM-DD"
```
wardi, line 2237: pot files don't include translations. they're translation templates for new languages and for updating existing translations only. Also, they should only be updated by using the `setup.py extract_messages` command

ckanext/canada/i18n/en/LC_MESSAGES/ckanext-canada.po
```diff
@@ -6133,3 +6133,5 @@ msgstr "Keep me logged in"
 #~ msgid "time_descend"
 #~ msgstr "Newest first"
 
+msgid "Date format incorrect. Expecting YYYY-MM-DD"
+msgstr "Date format incorrect. Expecting YYYY-MM-DD"
```
wardi, line 6137: These changes won't do anything. `msgstr ""` or a missing message means "leave it as it is"

JVickery-TBS:
@wardi I have tested the webform and the api a lot and have not run into any issues. So I am fairly happy with the string validation now. I did add in some json string validation, which will later get replaced by scheming subfields.

I also noticed some css/layout issues with the read template. So there is some new css and html there.

ckanext/canada/schemas/dataset.yaml
```diff
@@ -78,21 +78,25 @@ dataset_fields:
     en: Parent identifier
     fr: Identifiant parent
   form_snippet: null
+  validators: string_safe
```
wardi, line 81: I like to use `scheming_required` https://github.com/ckan/ckanext-scheming/blob/3aa6ed7f134d14b882c5cea474d6519eed36bac9/ckanext/scheming/validation.py#L165 as the first validator most of the time because if `required:true` is not set it acts as `ignore_missing` which saves the `if value is None or value is missing: return` dance at the beginning of all the new validators you added.

`scheming_required` also actually enforces requiredness when `required:true` is added to the schema

JVickery-TBS: @wardi okay I have removed those code checks for empty values, and added some scheming_required/ignore_missing on a bunch of the fields.

I also added in the new validator for the licence_choices.

wardi:
this looks good. I also think it's worth making a small PR against ckan for the `licence_choices` validator. It can be added to the core schema here https://github.com/ckan/ckan/blob/844f9b46431076bcf43f5b5bf72c5e0f9281d01b/ckan/logic/schema.py#L136

PR #1270, opened 2022-04-20
extract member list for DAC recombinant type

RabiaSajjad:


bin/download_dac.py
```diff
@@ -0,0 +1,59 @@
...
+                outfile = arg
+        if not outfile:
+            raise ValueError
+
+    except ValueError:
+        print 'USAGE: python download_dac.py -o/--output outfile.yaml'
+        sys.exit(1)
+    except getopt.GetoptError:
+        print 'USAGE: python download_dac.py -o/--output outfile.yaml'
+        sys.exit(1)
```
wardi, line 36: this should work too
```suggestion
    except (ValueError, getopt.GetoptError):
        print 'USAGE: python download_dac.py -o/--output outfile.yaml'
        sys.exit(1)
```

bin/download_dac.py
```diff
@@ -0,0 +1,59 @@
...
+          '=SELECT distinct "CONTACT_NAME" from ' \
+          '"59c86d81-fd29-495f-9a7a-7ba3d2c70cc2" '
+    response = urlopen(url)
+    results = json.loads(response.read().decode('utf-8'))
+    records = sorted(results['result']['records'],
+                     key=lambda k: k['CONTACT_NAME'].lower())
+
+    for record in records:
+        output.write('\"' + record['CONTACT_NAME'] + '\": ' +
+                     '\"' + record['CONTACT_NAME'] + '\"\n')
```
wardi, line 53: why not use pyyaml to generate this output file? Then we don't have to worry about things like `"` in the contact name values which would break the output here

wardi:
that works

PR #1273, opened 2022-04-27
add repeating field credit to open info schema

RabiaSajjad:


ckanext/canada/schemas/presets.yaml
```diff
@@ -307,6 +313,52 @@ presets:
...
+      fr:
+        en: Credit (French)
+        fr: Crédit (français)
+    fluent_help_text:
+      en:
+        en: The English names of entities (persons, groups, GC Departments or Agencies) that are credited for the dataset
+        fr: placeholder
+      fr:
+        en: The French names of entities (persons, groups, GC Departments or Agencies) that are credited for the dataset
+        fr: placeholder
```
wardi, line 339: for repeating subfields these settings don't do anything do they?
```suggestion
```

wardi:
that looks better

wardi:
looks good!

PR #1275, opened 2022-05-03
Update Makefile

PatLittle:


wardi:
What would the client think about changing the name of the last-two-years file instead of calling the new file `-full`?

PR #1283, opened 2022-05-13
Update presets.yaml

PatLittle:
add gzip as an option for format type

### DIFF
--- a/ckanext/canada/schemas/presets.yaml
+++ b/ckanext/canada/schemas/presets.yaml
@@ -3336,6 +3336,9 @@ presets:
       label: GeoSoftGrids
     - value: GRIB1
     - value: GRIB2
+    - value: Gzip
+      mimetype: application/gzip
+      label: Gzip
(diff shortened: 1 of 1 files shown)


ckanext/canada/schemas/presets.yaml
```diff
@@ -3336,6 +3336,9 @@ presets:
       label: GeoSoftGrids
     - value: GRIB1
     - value: GRIB2
+    - value: Gzip
```
wardi, line 3339: I think this is the correct extension
```suggestion
    - value: GZ
```

wardi:
Do we want to add GZ or should users give the format that has been compressed (e.g. `.txt.gx` would be better stored as `TXT` wouldn't it?)

PatLittle:
I agree in principle - I just worry that a user will try to download it and won't know how to extract it. Especially since our tablets don't have an included way to extract it besides `tar -xf ` on the commandline

PR #1291, opened 2022-06-14
add timestamp to commands in phony targets

RabiaSajjad:


bin/pd/Makefile
```diff
@@ -145,7 +145,7 @@ ati: upload-ati rebuild-ati
 
 .PHONY: upload-ati
 upload-ati: $(workdir)/filtered/ati.csv $(workdir)/filtered/ati-nil.csv
-	@echo Uploading ATI Summaries...
+	@echo $(shell date): Uploading ATI Summaries...
```
wardi, line 148: How about something like
```suggestion
	@$(echo_date) Uploading ATI Summaries...
```
so that we can define `echo_date` above to include `echo` and whatever date format prefix we choose. That would let us use the One True Date format without repeating code all over the Makefile.

bin/pd/Makefile
```diff
@@ -38,6 +38,7 @@ yyyy_mmdd := $(shell date "+%Y %m%d")
 yyyy := $(word 1,$(yyyy_mmdd))
 yyyymmdd := $(word 1,$(yyyy_mmdd))$(word 2,$(yyyy_mmdd))
 isodate := $(shell TZ=UTC date "+%FT%T")
+echo_date := echo $(isodate)
```
wardi, line 41:
```suggestion
echo_date := echo $(isodate):
```

PR #1292, opened 2022-06-15
Update filter_ati.py

PatLittle:
This should make all requests from Jan 2020 or later available in the search and dataset.

This is just an interim step. There is still a desire to release the full historical data.

### DIFF
--- a/bin/filter/filter_ati.py
+++ b/bin/filter/filter_ati.py
@@ -10,7 +10,7 @@ import csv
 import sys
 import codecs
 
-WINDOW_YEARS = 2
+#WINDOW_YEARS = 2
 
 REMOVE_COLUMNS = [
     'record_created',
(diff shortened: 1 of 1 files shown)


bin/filter/filter_ati.py
```diff
@@ -10,7 +10,7 @@
 import sys
 import codecs
 
-WINDOW_YEARS = 2
+#WINDOW_YEARS = 2
```
wardi, line 13:
```suggestion
start_year_month = 2020, 1  # publicly accessible records
```

bin/filter/filter_ati.py
```diff
@@ -26,8 +26,8 @@ def main():
     assert bom == codecs.BOM_UTF8
     sys.stdout.write(codecs.BOM_UTF8)
 
-    today = datetime.datetime.today()
-    start_year_month = (today.year - WINDOW_YEARS, today.month)
+    #today = datetime.datetime.today()
+    start_year_month = (int('2020'), int('01'))
```
wardi, line 30:
```suggestion
```

bin/filter/filter_ati.py
wardi: better to define constants at the top of a module

PR #1294, opened 2022-06-20
show user email if same org

RabiaSajjad:


ckanext/canada/helpers.py
```diff
@@ -538,3 +538,10 @@ def canada_check_access(package_id):
         return h.check_access('package_update', {'id': package_id})
     except NotFound:
         return False
+
+
+def user_in_organization(user):
+    u = User.get(user['name'])
+    group_ids = u.get_group_ids(group_type="organization")
+
+    return u.is_in_groups(group_ids)
```
wardi, line 547: this checks if the user is in their own organizations?

wardi:
I think we can use the `h.organizations_available` helper instead of writing one of our own

ckanext/canada/helpers.py
RabiaSajjad: no, it checks groups for a user in user list against logged in user (self's) groups

wardi: I'm reading the definition of `is_in_groups` and `get_group_ids` and I don't see any reference to the logged in user. `self` means the object the method was called against (`u` in this case)

ckanext/canada/helpers.py
```diff
@@ -538,3 +538,28 @@ def canada_check_access(package_id):
...
+def get_user_email(user_id):
+    '''
+    Return user email address if belong to the same organization
+    '''
+    u = User.get(user_id)
+    orgs = h.organizations_available()
+    org_ids = [o['id'] for o in orgs]
+
+    if not u.is_in_groups(org_ids):
+        return False
```
wardi, line 552: better to return `None` or `""` because on success we return an email string

ckanext/canada/helpers.py
```diff
@@ -538,3 +538,28 @@ def canada_check_access(package_id):
...
+               'keep_email': True}
+
+    try:
+        data_dict = {'id': user_id}
+        user_dict = ckan.logic.get_action('user_show')(context, data_dict)
+
+        return user_dict['email']
+
+    except NotFound as e:
+        return False
```
wardi, line 565: here too. `""` is easier to deal with in jinja

wardi:
LGTM

PR #1295, opened 2022-06-20
Added background color for modal headers. Added z-index for resource …

JVickery-TBS:
…modules action panels.


wardi:
manual url cache busting can be found in ckanext/canada/templates/public/base.html

wardi:
LGTM (wait for @PatLittle 's confirmation on staging though)

PR #1297, opened 2022-06-27
fix for OPEN-1952

RabiaSajjad:


wardi:
nice. was this all that was required?

PR #1300, opened 2022-07-14
Fix/user list

JVickery-TBS:
- Removed the mapping for the user list (allowing the admin.index template to be loaded by default).
- Modified the user toolbar snippet to have the "Admin" button link go to the admin.index url.
- Modified the admin base template tabs to have the "User" tab go to the user.index url.
- Modified the user list template to have search and paging.
- Modified the user list template to display similarly to the other admin templates.
- Removed the wet html attributes from the user list table as it was not loading the datatable (and the datatable would not work fully due to ckan pagination, although we could have increased the users per page above 4000. At least this way, we stick closer to how ckan core works).
- Added super call to the user list scripts as it was missing and preventing the font-awesome scripts from loading.
- Modified the config set for the users per page from 4000 to 250.
- Added a modified user search snippet template.
- Fixed the wet pager to work with Pylon request.
- Added translations.


ckanext/canada/templates/internal/user/list.html
```diff
@@ -1,43 +1,95 @@
-{% extends 'admin/base.html' %}
+{% if c.userobj and c.userobj.sysadmin %}
```
wardi, line 1: we still need this large template in our extension? Was hoping we could remove some code as part of this change so that we have an easier time upgrading. I'm assuming this is mostly copied from core, right?

JVickery-TBS: @wardi Core does not display the user full name or user role. In our template, I am keeping the "User" tab in the ckan-admin structure. Without it, the user list does not display anywhere in the ckan-admin in core as far as I can tell.

wardi: Do you think this could be submitted as a core feature so we don't have to maintain it in our extension and fix it as core changes?

PR #1304, opened 2022-07-19
add cascading dropdown for select fields in nap5

RabiaSajjad:


wardi:
the sorting I mentioned is here: https://github.com/open-data/ckanext-canada/blob/a92856fa7e772f9e95393e10a55a6e6d68ff4185/ckanext/canada/templates/internal/recombinant/create_pd_record.html#L65 if we were to prefix the text choices with the values then the lexicographic ordering *should* put all the fields in the right order.

I was thinking we could add an option to enable the prefix on drop-downs to let this code do its work.

If you think it's easier to keep the sorting in the JS code you've added though, I'm fine with that too

ckanext/canada/tables/nap5.yaml
```diff
@@ -0,0 +1,1333 @@
...
+      en: Commitments
+      fr: Engagements
+    description:
+      en: The 5 commitments in the 2022-2024 National Action Plan on Open Government
+      fr: Les 5 engagements dans le Plan d’action national du Canada pour un gouvernement ouvert de 2022-2024
+    obligation: Mandatory
+    excel_required: true
+    form_required: true
+    datastore_type: text
+    excel_full_text_choices: true
```
wardi, line 56: you could add an option like
```yaml
form_choices_include_codes: true
```
that will add the codes to the select text similar to how `excel_full_text_choices` adds the text to the options in excel

ckanext/canada/controller.py
```diff
@@ -817,10 +817,18 @@ def create_pd_record(self, owner_org, resource_name):
...
+            for r in chromo['fields']:
+                if r['datastore_id'] == f['datastore_id'] and \
+                        r.get('form_full_text_choices', False):
+                    choice_fields[f['datastore_id']] = [{'value': k,
+                                                         'label': k + ': ' + v}
+                                                        for (k, v) in f['choices']]
+            if f['datastore_id'] not in choice_fields:
+                choice_fields[f['datastore_id']] = [{'value': k, 'label': v}
+                                                    for (k, v) in f['choices']]
+
```
wardi, line 831: Let's take the loop that doesn't depend on `f` out front to avoid unnecessary O(n**2)

```suggestion
        for f in chromo['fields']:
            if f['datastore_id'] == r['datastore_id']:
                ft_choices = f.get('form_full_text_choices', False)
                break

        choice_fields = {
            f['datastore_id']: [
                {
                    'value': k, 
                    'label': k + ': ' + v if ft_choices else: v,
                } for (k, v) in f['choices']
            ] for f in h.recombinant_choice_fields(resource_name)
        }
```

ckanext/canada/internal/static/cascading_select.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+      } else {
+        while (element.options.length) {
+          element.remove(0);
+        }
+      }
+
+      // add new select choices based on selected value of parent field
+      let options = all_options[element.name][selected_option];
+      let sorted_options = Object.entries(options).sort();
```
wardi, line 59: these might be sorted already now with the keys in the select values right?

ckanext/canada/controller.py
RabiaSajjad: unfortunately, `if f['datastore_id'] == r['datastore_id']` won't work as I need to find the value of `form_full_text_choices` for each chromo field that I find in the `recombinant_choice_fields(resource_name)` so I have to double loop. However, I can optimize the code to improve time complexity.

PR #1309, opened 2022-08-04
Feature/gcnotify

JVickery-TBS:
Added simple if-else statement along with an inline comment regarding the custom function in the ckan.lib.mailer that is added by the ckanext-gcnotify

### DIFF
--- a/ckanext/canada/controller.py
+++ b/ckanext/canada/controller.py
@@ -431,12 +431,21 @@ class CanadaUserController(UserController):
                 return self._save_new(context)
             except HTTPFound:
                 # redirected after successful user create
-                notify_ckan_user_create(
+                import ckan.lib.mailer
+                if hasattr( ckan.lib.mailer, "notify_ckan_user_create" ):
+                  ckan.lib.mailer.notify_ckan_user_create(
                     email=request.params.get('email', ''),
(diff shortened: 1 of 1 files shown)


ckanext/canada/controller.py
```diff
@@ -431,12 +431,22 @@ def register(self, data=None, errors=None, error_summary=None):
...
                     username=request.params.get('name', ''),
                     phoneno=request.params.get('phoneno', ''),
                     dept=request.params.get('department', ''))
+                else:
+                  notify_ckan_user_create(
+                      email=request.params.get('email', ''),
+                      fullname=request.params.get('fullname', ''),
+                      username=request.params.get('name', ''),
+                      phoneno=request.params.get('phoneno', ''),
+                      dept=request.params.get('department', ''))
```
wardi, line 449:
```suggestion
                getattr(
                    ckan.lib.mailer,
                    "notify_ckan_user_create",
                    notify_ckan_user_create
                )(
                    email=request.params.get('email', ''),
                    fullname=request.params.get('fullname', ''),
                    username=request.params.get('name', ''),
                    phoneno=request.params.get('phoneno', ''),
                    dept=request.params.get('department', ''),
                )
```

PR #1312, opened 2022-08-18
avoid creating a temp user table for non-recombinant datastore upsert

RabiaSajjad:


ckanext/canada/logic.py
```diff
@@ -169,6 +169,9 @@ def datastore_create_temp_user_table(context):
     Create a table to pass the current username and sysadmin
     state to our triggers for marking modified rows
     '''
+    if 'user' not in context and context['ignore_auth']:
```
wardi, line 172:
```suggestion
    if 'user' not in context:
```

wardi: I think this change is good enough

PR #1313, opened 2022-08-18
Brought back canada_activity plugin and simplified it. Added dataset …

JVickery-TBS:
…facets to group_facets plugin implementation.

This contains the simplifying of the canada_activity plugin, not needing the extendedactivity extension anymore. (yay!)

This also contains the fixes for the organization facets not loading. So just adding those checks to the `group_facets`. ckan/ckan has labelled this already for back port: https://github.com/ckan/ckan/issues/7017 so we might see this soon in a patch and be able to remove those `group_type` checks.

### DIFF
--- a/ckanext/canada/plugins.py
+++ b/ckanext/canada/plugins.py
@@ -406,11 +406,15 @@ ckanext.canada:schemas/prop.yaml
 
     def group_facets(self, facets_dict, group_type, package_type):
         ''' Update the facets_dict and return it. '''
+        if group_type == 'organization':
+            return self.dataset_facets(facets_dict, package_type)
         return facets_dict
 
     def organization_facets(self, facets_dict, organization_type,
                             package_type):
         return self.dataset_facets(facets_dict, package_type)
+
+    
 
     def get_helpers(self):
         return dict((h, getattr(helpers, h)) for h in [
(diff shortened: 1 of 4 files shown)


ckanext/canada/plugins.py
```diff
@@ -406,12 +406,16 @@ def dataset_facets(self, facets_dict, package_type):
 
     def group_facets(self, facets_dict, group_type, package_type):
         ''' Update the facets_dict and return it. '''
+        if group_type == 'organization':
```
wardi, line 409: add a comment here with a FIXME linking to the upstream issue, please

ckanext/canada/plugins.py
```diff
@@ -789,6 +793,21 @@ def datastore_delete(up_func, context, data_dict):
...
+
+    def get_actions(self):
+        return ({'datastore_upsert':datastore_upsert,
+                'datastore_delete': datastore_delete})
+
+    def update_config(self, config):
+        logic_validators.object_id_validators.update({
+            'changed datastore': logic_validators.package_id_exists,
+            'deleted datastore': logic_validators.package_id_exists,
+        })
```
wardi, line 808: While we're cleaning this up, can this be merged with one of our other plugins? I don't think we ever want this disabled on one of our sites, right?

ckanext/canada/plugins.py
JVickery-TBS: @wardi comments added: https://github.com/open-data/ckanext-canada/pull/1313/commits/79a72e41f00d2e071b47fdb8d687bdbc9234a2cc

We can just fully remove the `group_facets` methods once it gets patched.

ckanext/canada/plugins.py
JVickery-TBS: @wardi okay `canada_activity` is removed and it's plugin code has been migrated into `canada_public`: https://github.com/open-data/ckanext-canada/pull/1313/commits/9d44b352ac70fff6f8345d43dccf2f9683879365

wardi:
I like it, just also update the plugins in the README and be careful about the config change required on deployment

PR #1318, opened 2022-08-25
override datatable_column_label block 

RabiaSajjad:
override `datatable_column_label` block in datatables ckan extension to display language specific column labels


wardi:
LGTM

PR #1319, opened 2022-08-25
Added `NotFound` exception handling to `portal_update` processing to …

JVickery-TBS:
…not fail on purged packages.

The dataset_purge action does not delete related activities. Our `portal_update` command gathers the `object_id`s from the activities table, and uses those ids to gather the package objects. However if you have purged a package, this cases and error.

This new try catch will handle the `NotFound` exception that is raised from the `package_show` action call.

### DIFF
--- a/ckanext/canada/commands.py
+++ b/ckanext/canada/commands.py
@@ -353,11 +353,12 @@ class CanadaCommand(CkanCommand):
         for result in data:
             package_id = result['package_id']
             try:
-              source_package = registry.action.package_show(id=package_id)
-              source_package = get_datastore_and_views(source_package, registry)
-              packages.append(json.dumps(source_package))
+                source_package = registry.action.package_show(id=package_id)
             except NotFound:
-              print package_id + " not found in database."
+                print package_id + " not found in database."
+            else:
+                source_package = get_datastore_and_views(source_package, registry)
+                packages.append(json.dumps(source_package))
 
         if data:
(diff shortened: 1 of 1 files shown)


ckanext/canada/commands.py
```diff
@@ -352,9 +352,12 @@ def _changed_packages_since(self, registry, since_time):
...
             package_id = result['package_id']
-            source_package = registry.action.package_show(id=package_id)
-            source_package = get_datastore_and_views(source_package, registry)
-            packages.append(json.dumps(source_package))
+            try:
+              source_package = registry.action.package_show(id=package_id)
+              source_package = get_datastore_and_views(source_package, registry)
+              packages.append(json.dumps(source_package))
+            except NotFound:
+              print package_id + " not found in database."
```
wardi, line 360:
```suggestion
            try:
                source_package = registry.action.package_show(id=package_id)
            except NotFound:
                print package_id + " not found in database."
            else:
                source_package = get_datastore_and_views(source_package, registry)
                packages.append(json.dumps(source_package))
```

wardi: *always* 4 space indents for python files and using `else` here lets us be more precise with where the exception is coming from

JVickery-TBS: @wardi oops, my editorconfig in that project was set to 2 indents, that is fixed now. And I added the `else` statement as well.

wardi:
LGTM

PR #1320, opened 2022-09-01
Fix/jquery on loads

JVickery-TBS:
Added window on loads to a bunch of the JS scripts to catch miscellaneous `$` or `jQuery` or `wb` not defined errors. This makes sure that jquery and wet scripts will be loaded by the time the scripts execute.

Added `load_voting_scripts` js function to the package read template. This is a temporary solution to the wet ajax-fetch plugin stripping script, style, and link tags in the returned html content. Sadly due to how the wet script work (along with arbitrary jquery loading), a simple `$.on('ajax-fetched.wb')` binding does not work 100% of the time due to the asynchronous nature of js. Also a simple `addEventListener('ajax-fetched.wb')` will not work either as wet does a jquery trigger and jquery triggers are not picked up by native event listeners.

So I had to create a mutation observer and observe the class list to look for the class the wet script adds once completed. I then check if the voting scripts have been loaded yet, and if the ajax actually put a form into the page. Then load the scripts and styles.


wardi:
Pretty big changes. Should we wait to hear from the wet folk on their recommended fix?

Also don't love the massive static JS blob in the read.html template, that bit would be better in a separate script.

@RabiaSajjad do you have any thoughts on these change?

wardi:
yes we can always revert later

PR #1324, opened 2022-09-12
Update nap5.yaml

PatLittle:


wardi:
LGTM. I don't think the quotes will cause a problem with yaml parsing. (one of the things that would be picked up if we had CI working again)

PR #1326, opened 2022-09-22
integrate invitation manager service to portal

RabiaSajjad:


wardi:
do we need this loaded on every page? what about adding it only on the required dataset pages instead?

RabiaSajjad:
> do we need this loaded on every page? what about adding it only on the required dataset pages instead?

I understand you are concerned about performance. The assets (css/javascripts) won't load for any pages on the registry, I am overwriting the blocks on both base.html and page.html for `internal` templates to avoid loading invitation manager assets. On the portal, we only have dataset and resource pages which would load these assets. If you prefer, I can optimize it to load only for dataset pages.

wardi:
IIRC the requirement was to have a survey on only specific departments' dataset pages and I don't think that can be done by applying the invitation manager to all pages on the site.
(Was also concerned it might appear in places like the frames of datastore views or other pages we're not expecting)

RabiaSajjad:
Replaced by https://github.com/open-data/ckanext-canada/pull/1495

PR #1327, opened 2022-09-23
Feature/ati informal requests portal data

JVickery-TBS:
Shell script to update the resource `e664cf3d-6cb7-4aaa-adfa-e459c2552e3e` with the `generated_file` (generated from Drupal cron https://github.com/open-data/og/pull/19)

### DIFF
--- a/bin/update_ati_analytics.sh
+++ b/bin/update_ati_analytics.sh
@@ -47,7 +47,13 @@ generated_file='/opt/tbs/ckan/smb_portal/portal/public/ati-informal-requests-ana
 
 if [[ ! -f "${generated_file}" ]]; then
 
-    error_message '/opt/tbs/ckan/smb_portal/portal/public/ati-informal-requests-analytics.csv does not exist.';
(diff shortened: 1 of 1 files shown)


bin/update_ati_analytics.sh
```diff
@@ -0,0 +1,72 @@
...
+
+    if [[ ! -f "${generated_file}" ]]; then
+
+        error_message '/opt/tbs/ckan/[smb_portal|smb]/portal/public/ati-informal-requests-analytics.csv does not exist.';
+
+    fi
+
+fi
+
+tmp_file='/opt/tbs/tmp/ati-informal-requests-analytics.csv';
```
wardi, line 60: `mktemp` can securely create unique file names so that you don't have to worry about cases where the script is run twice and one truncates the file from the other.

JVickery-TBS: @wardi  https://github.com/open-data/ckanext-canada/pull/1327/commits/5aa63272e754bd33fa7339742eef67c293d9dbda

I used temp directories instead of files so that the resource in ckan would keep the file name `ati-informal-requests-analytics.csv`

PR #1328, opened 2022-10-03
Added flask view override for Resource Create. Added flash message fo…

JVickery-TBS:
…r resource creation.

Just adding another Flask View override to add flash messages when a resource is created.


wardi:
Sure this looks good.

Did you test with creating open information records? Does this catch the resource create for them too?

wardi:
thanks, looks good

PR #1330, opened 2022-10-06
Removed `license_title`, `license_title_fra`, `license_url_fra`, and …

JVickery-TBS:
…`license_url` from the `_trim_package` method.


wardi:
Would you test to see if this makes a difference publishing a dataset on staging? I don't see how these values would be ever stored in our datasets: there are no extra fields defined for them and they're not in the base ckan model. I'm puzzled by this one.

PR #1331, opened 2022-10-11
Template fixes for resource views. Using flask views instead of pylon…

JVickery-TBS:
… controllers now.


wardi:
This looks good to me.

PR #1332, opened 2022-10-13
Added missing `resource.edit` and `resource.new` flask view overrides.

JVickery-TBS:
Set the `title` and `title_fr` validators for resource default views.

### DIFF
--- a/ckanext/canada/plugins.py
+++ b/ckanext/canada/plugins.py
@@ -299,8 +299,9 @@ def resource_view_create_bilingual(up_func, context, data_dict):
             schema=dict(
                 s,
                 title=[get_validator('default')('View'), get_validator('unicode_safe')],
-                title_fr=[get_validator('default')('Voir'), get_validator('unicode_safe')],
-                description_fr=list(s['description']),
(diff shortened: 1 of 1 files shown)


JVickery-TBS:
@wardi Final issue here is that at this time the `title` is already set in the `data_dict` as `view_info.get('default_title', _('View'))`

So as it is now, it does not solve the issue with the default view English name. E.g. making the view in French would result in the `title` being saved as "Voir" in the database.

I don't think we can just unset the `title` from the `data_dict` here as that would override the webform submission data as well.

wardi:
Hmm. Maybe removing the default_title in core is the least bad option here (won't be hard to merge with upstream but worth adding a comment about why we're doing it)

Another possibility is having a `title_en` field, that could be done without any core changes but we'd have to handle existing views that don't have these values set in the templates so it might be kind of messy.

JVickery-TBS:
@wardi the schemas for the resource view fields have been updated now with default values for `title` and `description`

wardi:
@JVickery-TBS looks great

PR #1333, opened 2022-10-19
allow registry to run without validation extension (for tests)

wardi:


wardi:
@JVickery-TBS `git cherry-pick` this onto your branch or I can merge this to master

PR #1334, opened 2022-10-27
Feature/GitHub tests

JVickery-TBS:
github workflow tests should work now.

I had to add a subclass for the Organization factory `Canada_Organization` as we implemented the org schema.

`test-requirements.txt` file has been added.

I left the lint job commented out for now.


ckanext/canada/tests/factories.py
```diff
@@ -0,0 +1,12 @@
+from ckan.tests.factories import Organization
+
+class Canada_Organization(Organization):
```
wardi, line 3:
```suggestion
class CanadaOrganization(Organization):
```
python style issue (don't use _ in class names)

wardi:
only the class name issue, everything else looks good

PR #1336, opened 2022-11-04
Feature/GitHub tests

JVickery-TBS:
- Modified the github tests workflow to function with our ckan 2.9 setup (with python2.7 and not 3)
- Removed `pylons` usages throughout the code
- Switch all the tests from using `FunctionalTestBase` to now using `pytest` with `@pytest.mark.usefixtures('clean_db')`
- Changed test classes `setup` methods to just `__init__` as they are no longer children of `FunctionalTestBase`
- Removed the `logged_in` method from the user controller and replaced it with a Flask View override of the user `login` and `logged_in` view functions.

TODO: replace all of the `nose` assertion usages with basic asserts.


JVickery-TBS:
@wardi well, apparently, I did it. 

I have some TODOs in there to assert the raises to expected error messages. Let me know if you want me to do that as part of this PR or later, as it used to just use `nose.tools.assert_raises` which just checked if the code raised the given exception class.

README.md
```diff
@@ -1,6 +1,6 @@
 # ckanext-canada
 
-[![Tests](https://github.com/open-data/ckanext-canada/workflows/Tests/badge.svg?branch=master)](https://github.com/open-data/ckanext-canada/actions?query=name%3ATests)
+[![Tests](https://github.com/open-data/ckanext-canada/workflows/Tests/badge.svg?branch=canada-v2.9&event=push)](https://github.com/open-data/ckanext-canada/actions?query=name%3ATests)
```
wardi, line 3: looks like something we'll forget to change :thinking:

.github/workflows/test.yml
```diff
@@ -17,13 +17,13 @@ jobs:
...
       matrix:
-        ckan-version: [2.8]
+        ckan-version: [2.9]
       fail-fast: false
 
     name: CKAN ${{ matrix.ckan-version }}
     runs-on: ubuntu-20.04
     container:
-      image: openknowledge/ckan-dev:${{ matrix.ckan-version }}
+      image: openknowledge/ckan-dev:2.8
```
wardi, line 26: add a comment here about why we need to use the 2.8 version of the image?

.github/workflows/test.yml
```diff
@@ -17,13 +17,13 @@ jobs:
     # needs: lint
     strategy:
       matrix:
-        ckan-version: [2.8]
+        ckan-version: [2.9]
```
wardi, line 20: version is hard-coded in a number of places, should we remove this setting?

ckanext/canada/tests/fixtures.py
```diff
@@ -0,0 +1,65 @@
...
+from ckanapi import LocalCKAN
+from factory import Sequence
+from ckanext.canada.tests.factories import CanadaOrganization as Organization
+
+import logging
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def prepare_dataset_type_with_resources(request):
```
wardi, line 13:
```suggestion
def prepare_recombinant_dataset_type_with_resources(request):
```

ckanext/canada/tests/test_briefingt.py
```diff
@@ -1,31 +1,33 @@
 # -*- coding: UTF-8 -*-
-from nose.tools import assert_equal, assert_raises
 from ckanapi import LocalCKAN, ValidationError
 
-from ckan.tests.helpers import FunctionalTestBase
-from ckanext.canada.tests.factories import CanadaOrganization as Organization
+import pytest
+from ckanext.canada.tests.fixtures import prepare_dataset_type_with_resources
```
wardi, line 5: do you need this import? I thought pytest had some magic to collect fixtures

wardi: (like the clean_db one used below)

ckanext/canada/tests/test_pd.py
```diff
@@ -1,37 +1,32 @@
...
+
 
     def test_top_bucket(self):
-        assert_equal(
-            dollar_range_facet('foo', [100, 500, 1000], 5000), {
-                'foo_range': u'2',
-                'foo_en': u'A: $1,000.00+',
-                'foo_fr': u'A: 1\xa0000,00\xa0$ +'})
+        assert dollar_range_facet('foo', [100, 500, 1000], 5000) == {
+           'foo_range': u'2',
```
wardi, line 16:
```suggestion
            'foo_range': u'2',
```

wardi:
that was a lot, and this method of passing parameters to fixtures isn't pretty, but tests are green :-)

.github/workflows/test.yml
JVickery-TBS: Removed matrix stuff.

ckanext/canada/tests/fixtures.py
JVickery-TBS: n/a removed custom fixtures

ckanext/canada/tests/test_briefingt.py
JVickery-TBS: n/a removed custom fixtures

ckanext/canada/tests/test_pd.py
JVickery-TBS: fixed.

.github/workflows/test.yml
JVickery-TBS: Comment added.

README.md
JVickery-TBS: Changed to master branch. People will get emails for tests failing in canada-v2.9 anyways, so should not need visual badge for this specific branch.

JVickery-TBS:
@wardi I have cleaned everything up. Namely, instead of the custom strange fixtures, I am now only using the setup class methods and directly calling `reset_db`.

So it all looks a lot nicer now and makes way more sense.

Other comments have been addressed.

As discussed in slack, the `test_example` and `test_multiple_errors` in the consultations tests do not work in that order. Which is currently a mystery as the other tests do very similar things with `datastore_upsert`. I will try your suggestion of putting `None` as the values to the expected erroring fields.

ckanext/canada/tests/test_ati.py
```diff
@@ -1,55 +1,97 @@
...
+        """
+        reset_db()
+
+        log.info('Running setup for {}'.format(self.__name__))
+
         org = Organization()
         lc = LocalCKAN()
+
+        log.info('Creating organization with id: {}'.format(org['name']))
+        log.info('Setting organization dataset type to {}'.format('ati'))
```
wardi, line 28:
```suggestion

        org = Organization()
        lc = LocalCKAN()

```

wardi: debug logging probably not required here

JVickery-TBS: Logging removed.

JVickery-TBS:
@wardi Okay I think this is good to go now.

ckanext/canada/tests/test_consultations.py
```diff
@@ -82,14 +82,6 @@ def test_multiple_errors(self):
...
 
 
-    def test_example(self):
-        lc = LocalCKAN()
-        record = get_chromo('consultations')['examples']['record']
-        lc.action.datastore_upsert(
-            resource_id=self.resource_id,
-            records=[record])
-
-
```
wardi, line 67: just to confirm, moving this method above isn't required for the tests to pass now, right?

JVickery-TBS: @wardi Correct, the order of the methods does not matter now with the change to `setup_method`

wardi: ok, then this all looks good

PR #1337, opened 2022-11-21
Added `readonly` and `size` form attributes to the organization `name…

JVickery-TBS:
…` (`url`) field.

I also added size 45 to the field to make the generated url more visible to the biz team. The generated url can get long if the English and French abbreviations are different from each other.


wardi:
this doesn't interfere with js updating the value?

JVickery-TBS:
@wardi Nope, the JS can still modify the value of the node.

PR #1338, opened 2022-11-21
Restructure of solr config files.

JVickery-TBS:
Restructured the directories and files for the solr configs. Namely, separating them into `solr6` and `solr8` dirs.

This will prevent any extra steps in renaming files during deployments.


wardi:
If we don't need the solr6 files for anything I'd rather remove them and keep the directories simpler

JVickery-TBS:
@wardi The solr 6 files have now been removed. So we just have the solr config files for the ckan portal and registry in the root, and the base config files for the various ati cores in the `ati` directory.

PR #1339, opened 2022-12-15
Updated download uri for country codes.

JVickery-TBS:
Old download URI 503 errors: https://raw.github.com/datasets/country-codes/master/data/country-codes.csv


JVickery-TBS:
So there was a github incident regarding the 503 error on raw.github.com. However the incident was not occurring on the githubuser.com.

I will bring this change in now, because github is using this new domain name for raw files, so I feel like we should use it.

PR #1341, opened 2022-12-16
Set datatable `edit_form` value to the value of `resource_update` acc…

JVickery-TBS:
…ess for Recombinant preview table.

### DIFF
--- a/ckanext/canada/helpers.py
+++ b/ckanext/canada/helpers.py
@@ -298,8 +298,6 @@ def get_datapreview_recombinant(resource_name, resource_id, owner_org, dataset_t
             out['priority'] = priority
             priority += 1
         fields.append(out)
-
-    chromo['edit_form'] = h.check_access('resource_update', {'id': resource_id})
     
     fids = [f['datastore_id'] for f in chromo['fields']]
     pkids = [fids.index(k) for k in aslist(chromo['datastore_primary_key'])]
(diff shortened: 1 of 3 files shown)


ckanext/canada/helpers.py
```diff
@@ -299,6 +299,8 @@ def get_datapreview_recombinant(resource_name, resource_id, owner_org, dataset_t
             priority += 1
         fields.append(out)
 
+    chromo['edit_form'] = h.check_access('resource_update', {'id': resource_id})
```
wardi, line 302: this doesn't look right. `chromo` contains the schema for this recombinant type that is shared by all threads and requests, but `check_access` is returns a value based on the current user's permissions.

JVickery-TBS: No longer modifying the `edit_form` chromo value.

JVickery-TBS:
@RabiaSajjad okay, as Ian pointed out, I should not have been modifying the `edit_form` value in the chromos as that is used in a lot of places, namely the NIL types.

I have now instead just used the `check_access` on `resource_update` throughout the WET datatable template and the datatable Controller method.

Basically, wherever it checks for the `chromo.edit_form` I also check if the user is able to edit the resource. I store the helper `check_access` value in a variable in the template and controller method as it is used more than once.

RabiaSajjad:
looks like a good solution @JVickery-TBS

PR #1342, opened 2022-12-19
collapse pco sub-orgs' briefingt/qpnotes records under pco

RabiaSajjad:


bin/migrate/migrate_briefingt_2022_12.py
```diff
@@ -0,0 +1,36 @@
...
+assert sys.stdin.read(3) == codecs.BOM_UTF8
+
+in_csv = unicodecsv.DictReader(sys.stdin, encoding='utf-8')
+out_csv = unicodecsv.DictWriter(sys.stdout, fieldnames=in_csv.fieldnames, encoding='utf-8')
+out_csv.writeheader()
+
+for line in in_csv:
+    if line['owner_org'] in sub_orgs:
+        line['owner_org'] = PCO['owner_org']
+        line['owner_org_title'] = PCO['owner_org_title']
```
wardi, line 35: I think this will do the same thing
```suggestion
        line.update(PCO)
```

RabiaSajjad:
@wardi what would be the optimal strategy to migrate nil reports? Typically each organization has only 1 nil report for a reporting period if there are no records to add. With merging, PCO would have multiple nil reports from its sub-organizations if they exist.

wardi:
@RabiaSajjad I guess the smartest thing would be to remove any nil reports if any of the new pco records overlap that same time period

PR #1343, opened 2022-12-22
More / Better Tests

JVickery-TBS:
Figured it might just be easier to communicate this in a PR.

So the TODOs are all where the skip tests are happening currently. I am assuming that all of these feature are in production now as all of these SkipTests were added 8 years ago?

Let me know if any of these tests are not actually in production, and can just be fully removed @wardi @RabiaSajjad


wardi:
@JVickery-TBS these tests would be nice to finish, but maybe there are some more important places to cover?

We could start from our internal user guide and verify that everything there is at least partially covered by tests. E.g. we had one regression recently with the `dry_run` parameter to `datastore_upsert` used when users want to check for errors with PD data they're submitting. It would have been nice if our tests had caught that issue.

wardi:
Looks like a good start.
Maybe we need to move these test from nose to pytest-style to get them working?

JVickery-TBS:
@wardi Okay tests should be good to go again now. Had to add in the `beaker.session.timeout`;

And there was an issue with the Revisions (activity streams): `DetachedInstanceError: Instance <Revision> is not bound to a Session; attribute refresh operation cannot proceed` <- I solved this for now by just having `ckan.activity_streams_enabled = false` in the `test-core.ini`. I would really like to try to get the activity streams working in the test though.

We no longer have any skipped tests now. And I made the `LocalCKAN()` usages throughout all the test classes better. And our custom test factories have been improved now too.

I did have to add in `create_datastore_views = 'false'` to the `CanadaDataset` factory class, as I was getting validation errors on the creation of default views that CKAN kept trying to do. I could not figure out how to make it use the `CanadaResourceView` factory class. <- OKAY just checked our production value for `ckan.views.default_views` and it is an empty list, which solves the issues in the tests. There is now a test called `test_resource_view` in `test_validators.py` which asserts the validation error on the missing `title_fr` field for the resource view. So the resource view is good to go for sure in the tests.

JVickery-TBS:
@wardi Okay we did it! The `ckan.tests.factories.User` did not add the user object to the session. So if you did any actions that invoked any session stuff with that user, it would throw that `DetachedInstanceError` exception.

So I have made a `CanadaUser` factory sub class which does add the user object to the session (along with some inline comments detailing why the new sub class is doing it). So we can invoke actions with non sysadmin users during the tests.

So NOW all of these tests should be good to go, and some of the class `setup` methods and tests have been slightly modified to work with our custom factories.

JVickery-TBS:
Once this is merged into master, we can merge master into `canada-v2.9` and then change anymore nose stuff to pytests.

ckanext/canada/tests/test_ati.py
```diff
@@ -1,32 +1,33 @@
...
 
+
 class TestAti(FunctionalTestBase):
     def setup(self):
         super(TestAti, self).setup()
         org = Organization()
-        lc = LocalCKAN()
-        lc.action.recombinant_create(dataset_type='ati', owner_org=org['name'])
-        rval = lc.action.recombinant_show(dataset_type='ati', owner_org=org['name'])
+        self.lc = LocalCKAN()
```
wardi, line 15: LocalCKAN is essentially free. There's no need to share it between tests (more sharing between tests is more potential for state to leak and then we might accidentally have tests dependent on run order)

ckanext/canada/tests/factories.py
```diff
@@ -1,12 +1,76 @@
...
+    @classmethod
+    def _create(self, target_class, *args, **kwargs):
+        if args:
+            assert False, "Positional args aren't supported, use keyword args."
+
+        # The ckan.tests.factories.User does not add the user to a session
+        # so we need to override the _create method to add the user to the 
+        # session to prevent DetachedInstanceError execptions from being raised
+        user = target_class(**dict(kwargs, sysadmin=False))
+        Session.add(user)
```
wardi, line 18: This is really odd, but we can revisit it in the 2.9 branch after switching to pytest

ckanext/canada/tests/factories.py
```diff
@@ -1,12 +1,76 @@
...
+        Session.commit()
+        Session.remove()
```
wardi, line 20: does this `remove` do anything?

wardi:
Nice work, just a couple comments above. Let's remember to revisit the CanadaUser hack and remove it if we can later.

ckanext/canada/tests/test_ati.py
JVickery-TBS: @wardi The `setup` method is run before every test in the class, so would `self.lc = LocalCKAN()` just create a new instance of `LocalCKAN` at the beginning of each test?

Or is there a better way to clear that variable to prevent state leakage? Such as in the `teardown` method?

ckanext/canada/tests/factories.py
JVickery-TBS: I just copied what the `Sysadmin` factory had.

ckanext/canada/tests/factories.py
JVickery-TBS: Yes it is. Looks like the factories are all different in 2.9+ so hopefully it will just work normally in there.

ckanext/canada/tests/test_ati.py
wardi: you're right, this is the other setup method than the one I was thinking of. Still there's nothing valuable in these LocalCKAN instances and no time is spent creating them so the extra complexity of keeping them on the instance isn't really worthwhile for the simple cases IMO. When setup is creating a bunch of LocalCKAN instances I think it's okay to pass them on `self` because it saves lines of code in the tests, on the other hand using fixtures would be even cleaner, right?

ckanext/canada/tests/factories.py
wardi: ¯\\\_(ツ)\_/¯

ckanext/canada/tests/test_ati.py
JVickery-TBS: @wardi yeah, fixtures would be best, but as these ones are not pytest yet we cannot use fixtures to my knowledge?

I can change the `self.lc = LocalCKAN()` stuff to fixtures when modifying them for the `canada-v2.9` branch.

wardi: sure, that's fine.

PR #1345, opened 2023-01-23
Atom Feeds Fixes

JVickery-TBS:
Added dataset feed; 
Fixed links in templates; 
Fixed dataset feed endpoint mapping;

### DIFF
--- a/ckanext/canada/controller.py
+++ b/ckanext/canada/controller.py
@@ -31,7 +31,7 @@ from ckan.controllers.feed import (
 from ckan.lib import i18n
 import ckan.lib.jsonp as jsonp
 from ckan.controllers.package import PackageController
-from ckan.logic import parse_params
+from ckan.logic import parse_params, NotFound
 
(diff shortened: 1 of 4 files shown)


wardi:
any new code added as pylons controllers is going to slow down our migration to python 3. Consider holding off on this or implementing as a flask view instead?

JVickery-TBS:
@wardi looks like a lot of the Feed view's functions (even in CKAN 2.9) call url_for with a controller. So would that mean that the controller functions need to exist for the feed? So would I need to keep the custom Controller, and then just not map it and only register the new Blueprint for the feeds?

Or should I just write my own helper functions for my own feeds stuff so it does not call any of the Controller stuff? Seems like a fair amount of stuff to do to get this to work.

JVickery-TBS:
@wardi As discussed in Slack, we are not going to be doing a Flask Blueprint & View for this PR. We will work on it within the CKAN 2.9 upgrade development.

So I think this fix is fine for now, and I will get it onto staging to test there.

PR #1346, opened 2023-01-23
Flask Blueprint and View methods for Atom Feeds

JVickery-TBS:
Removed `CanadaFeedController` controller and mapping; Added flask blueprint and methods for custom Atom feeds;

TODO: Review differences between generated XML from Core and these view methods. Is there a difference? If so, is it major or can we just go with the Core feeds?


wardi:
What differences did you see between our code and core?  Would be helpful to leave code comments around them in our view

PR #1350, opened 2023-01-30
Even More 2.9 Compatibility

JVickery-TBS:
Removed test_functional; Added css overflow behavior to DataTables wrappers; Added package type to private read template; Updated all links to new format;

@wardi nothing to major in this initial commit. Namely the `pkg.type` in the `internal/package/read_base.html` template, which solves issue with PD type links not working on that tab. And another one to point out would be the CSS for the DataTables wrappers, just makes it so the overflow scroll is there, so that the tables do not expand the entire page's X axis to a crazy amount.

All the other changes so far is just changing all of the `controller= ... action=....` to the new format of `blueprint.view_method`.


ckanext/canada/plugins.py
```diff
@@ -37,6 +44,54 @@
...
+    #IDatasetForm
+    def is_fallback(self):
+        # type: () -> bool
+        return False
+
+
+    #IDatasetForm
+    def package_types(self):
+        # type: () -> list[str]
+        return ['dataset','info','prop']
```
wardi, line 62: Do we need to implement `is_fallback` and `package_types`? won't the parent implementation work?

ckanext/canada/templates/internal/organization/edit_base.html
```diff
@@ -7,23 +7,23 @@
...
             </li>#}
             <li class="{% if c.action == 'members' %}active{% endif %}">
-              <a href="{{h.url(controller='organization', action='members', id=c.group_dict.name)}}">{{_('Edit Members')}}</a>
+              <a href="{{h.url('organization.members', id=c.group_dict.name)}}">{{_('Edit Members')}}</a>
             </li>
           {% endif %}
           {% if h.check_access('group_update', {'id': c.group_dict.id}) %}
             <li class="{% if c.action == 'edit' %}active{% endif %}">
-              <a href="{% url_for controller='organization' if c.group_dict.type == 'organization' else 'group', action='edit', id=c.group_dict.id %}">{{ _('Edit') }}</a>
+              <a href="{% url_for 'organization' if c.group_dict.type == 'organization' else 'group' + '.edit', id=c.group_dict.id %}">{{ _('Edit') }}</a>
```
wardi, line 26: we should be using `group_dict.is_organization` here to be more technically correct

ckanext/canada/templates/internal/package/snippets/stages.html
```diff
@@ -36,7 +36,7 @@
         {# stage 1 #}
         <a><button class="btn-link highlight" name="save" value="go-resources" type="submit">{{ _('Add Resource Metadata') }}</button></a>
       {% else %}
-        {% link_for _('Add data'), controller='package', action='new', class_="highlight" %}
+        {% link_for _('Add data'), 'package.new', class_="highlight" %}
```
wardi, line 39: does this work? shouldn't it be one of the dataset ones?

ckanext/canada/templates/public/home/index.html
```diff
@@ -1,2 +1,2 @@
 {# hide CKAN front page on public site #}
-{{ h.redirect_to(controller='package', action='search') }}
+{{ h.redirect_to('package.search') }}
```
wardi, line 2: same question about these `package.search` urls: do they work? should they be dataset?

wardi:
removed test_functional so we have no front end test coverage anymore?

JVickery-TBS:
> removed test_functional so we have no front end test coverage anymore?

https://github.com/open-data/ckanext-canada/blob/canada-v2.9/ckanext/canada/tests/test_webforms.py test_functional was renamed test_webform. But test_functional was brought back in by a merge from canada-v2.8 -> canada-v2.9

ckanext/canada/plugins.py
JVickery-TBS: @wardi Yeah my bad, I had those in before I added the inherit on the implementation of DatasetForm.

ckanext/canada/templates/internal/organization/edit_base.html
JVickery-TBS: @wardi fixed!

ckanext/canada/templates/internal/package/snippets/stages.html
JVickery-TBS: @wardi is does work, but perhaps it is loading the old Controller? There is some stuff in the link_for and url_for that breaks down whether it is a Pylons or Flask thing. I have changed them all to be `dataset` or `pkg.type` now though, as that will set us up for the future. And perhaps make it better for the link_for and url_for to resolve stuff.

ckanext/canada/templates/public/home/index.html
JVickery-TBS: All `package` stuff has been replaced with `dataset` blueprint or the `pkg.type` blueprint.

wardi:
Thanks, looks good

PR #1352, opened 2023-02-03
Controllers to Views (+ OBD Removal)

JVickery-TBS:
Started restructuring the plugins into their own scripts;

Also removed OBD;


wardi:
-1 This kind of change creates conflicts when merging changes and doesn't really add anything.

JVickery-TBS:
@wardi Okay, all the plugins are back into the one `plugins.py` file. Added some inline comments and cleaned up the file a bit. And I will be moving the `DataGCCAPackageController` code to the `DataGCCDatasets` class. AND will be getting rid of the `controller.py` script and all of the before and after mappings, and getting those into views.

So more to come in this PR.

wardi:
same comment about re-indenting a bunch of code without changes: It only makes merging changes harder and can be done after we're on 2.9.

Also if we do decide to make a bunch of indentation changes it should be because we're implementing some kind of auto-formatter to keep our style consistent

JVickery-TBS:
@wardi okay, so should I just keep all the indents to 0 for the config update lines then? And then we will do the auto syntax stuff much later after the big merge?

JVickery-TBS:
Going to reset the head of this, and just have this be to move controllers to views, along with the OBD removal.

wardi:
Not just the config line updates, many of the plugins.py changes in this PR seem to re-indent, shift around braces, write imports differently, remove trailing commas etc. without changing the behaviour of the code. Please don't reformat code that's unrelated to your actual changes because it obscures the important part of the PR and makes merging harder later.

PR #1353, opened 2023-02-06
Controller to Views (+ OBD Removal)

JVickery-TBS:
Refactor: Parody: `CanadaDatasetsPlugin` -> `DataGCCADatasets`;
Removed `CanadaOpenByDefault` entry plugin class;
Added more comments for plugin implements;
Added TODOs and FIXME comments;


JVickery-TBS:
@wardi I moved the code from the `DataGCCAPackageController` class into the `DataGCCADatasets` class.

I also was able to just do `p.implements(p.IPackageController, inherit=True)` and not have all of those passing methods anymore as well. Let me know if that is fine.

There was also a FIXME in there to move a flash message into a Controller. But as we are trashing Controllers, I added it to the view functions. And it seems that the `allow_state_change` is not really available, and it does not seem like the view functions use them too much in 2.9? Not 100% sure on the use of `allow_state_change` in the flash message here, so let me know thanks!

ckanext/canada/plugins.py
```diff
@@ -578,7 +697,7 @@ def get_actions(self):
             ])
         actions.update({k: disabled_anon_action for k in [
             'current_package_list_with_resources',
-            'user_list',
+            'user_list', #FIXME: required for password reset w/ email
```
wardi, line 700: password resets with email address work now if you merge from master

wardi:
Yes, fewer plugins is better. Remember to update the example configs in the readme.

ckanext/canada/plugins.py
```diff
@@ -44,11 +44,12 @@
 log = logging.getLogger(__name__)
 
 
-class CanadaDatasetsPlugin(SchemingDatasetsPlugin):
+class DataGCCADatasets(SchemingDatasetsPlugin):
```
wardi, line 47: data.gc.ca was an old domain for the site, so `CanadaDatasetsPlugin` might be the better name to try to be uncoupled from future site name changes :-)

ckanext/canada/plugins.py
JVickery-TBS: @wardi https://github.com/open-data/ckanext-canada/commit/567a35aefabcd61bfffbf14b6c7bc8cb80518a9e <- this commit?

I will remove the comment for now and merge later on and see if it works for the 2.9 setup.

ckanext/canada/plugins.py
JVickery-TBS: Yay! `CanadaDatasetsPlugin` it will be!

ckanext/canada/tests/test_contracts.py
```diff
@@ -88,26 +82,22 @@ def test_2022_fields(self):
...
+        err = ve.value.error_dict
+        expected = ['vendor_postal_code',
+                    'buyer_name',
+                    'trade_agreement',
+                    'agreement_type_code',
+                    'land_claims',
+                    'indigenous_business']
+        assert 'records' in err
+        for k in set(err['records'][0]):
+            assert k in expected
```
wardi, line 97: if we already have code testing the error messages themselves then let's keep it so that we know we're hitting the correct validation rules

wardi: We could use just part of the error message and do `in` tests instead of equality if we want to make it easier to update the text of the messages, but by only checking the key we don't know that the thing we're testing is really working

ckanext/canada/tests/test_trim_package.py
```diff
@@ -1,48 +1,44 @@
...
+    def setup_method(self, method):
+        """Method is called at class level before EACH test methods of the class are called.
+        Setup any state specific to the execution of the given class methods.
+        """
+        reset_db()
+        clear_all()
+
+        self.example_pkg = Dataset()
+        resources = []
+        for _ in range(6):
```
wardi, line 24: don't use `_` as a variable name, it shadows the gettext function `_` in common use

ckanext/canada/tests/test_trim_package.py
```diff
@@ -1,48 +1,44 @@
...
-"""
+    def setup_method(self, method):
+        """Method is called at class level before EACH test methods of the class are called.
+        Setup any state specific to the execution of the given class methods.
+        """
+        reset_db()
+        clear_all()
+
+        self.example_pkg = Dataset()
```
wardi, line 22: +1 on switching to factory for the example data

ckanext/canada/tests/test_validators.py
```diff
@@ -24,19 +29,17 @@ def test_simple(self):
     def test_too_short(self):
         with pytest.raises(Invalid) as ie:
             canada_tags(u'h', {})
-        err = ie.value
-        expected = {}
-        #TODO: assert the expected error
-        assert ie is not None
+        err = str(ie.value)
+        expected = 'length is less than minimum'
+        assert expected in err
```
wardi, line 34:
```suggestion
        assert 'length is less than minimum' in err
```
writing tests this way makes it easier to see exactly what failed directly in the traceback

ckanext/canada/views.py
```diff
@@ -0,0 +1,960 @@
+import json
```
wardi, line 1: Would you leave this file name the same so that the changes made are more clear in this PR?

ckanext/canada/tests/test_trim_package.py
JVickery-TBS: variable has been renamed.

ckanext/canada/tests/test_validators.py
JVickery-TBS: all singles string `expected` variables have been removed, and the string has been moved directly into the `assert` statements.

ckanext/canada/views.py
JVickery-TBS: reverted refactor of `view.py` -> `views.py`

ckanext/canada/tests/test_contracts.py
JVickery-TBS: reverted simplification of error dict assertions. also updated `test_validators.py` error dict asserts to look the same for code parody.

PR #1354, opened 2023-02-14
remove search integration

wardi:
@thriuin how does this look?


thriuin:
Hi @wardi , this should cover it. I can't recall anything else that would need to be removed. Thanks! 
There are some settings that can be removed from actual .ini files related to search, but they should not cause any harm if they are not removed right away.

PR #1355, opened 2023-02-17
Admin Toolbar Button Hovers

JVickery-TBS:
fix(templates): admin menu button states;

- Added compliant contrast to admin menu button hover.

### DIFF
--- a/ckanext/canada/internal/static/internal.css
+++ b/ckanext/canada/internal/static/internal.css
@@ -652,14 +652,3 @@ ul#brf-pkg { padding-left: 0; }
 
 }
 
-/*
-Hover states for the #wb-so (Wet Boew Sign On)
-section in templates/internal/snippets/user.html
-*/
-body a.btn.og-btn-hover-light:hover{
-    color: white;
-}
-body a.btn.og-btn-hover-light:hover .badge{
-    background-color: white;
-    color: #335075;
-}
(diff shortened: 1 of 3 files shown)


ckanext/canada/internal/static/internal.css
```diff
@@ -650,4 +650,16 @@ ul#brf-pkg { padding-left: 0; }
...
+Hover states for the #wb-so (Wet Boew Sign On)
+section in templates/internal/snippets/user.html
+*/
+body a.btn.og-btn-hover-light:hover{
+    color: white;
+}
+body a.btn.og-btn-hover-light:hover .badge{
+    background-color: white;
+    color: #335075;
+}
```
wardi, line 665: Can we use an existing wet class for the effect? Otherwise we need to get business team's approval and maintain this customization in our repo across wet releases.

JVickery-TBS: The section uses the id `wb-so` which has all of the text alignment, padding, and margins. But it also borks button hovers, which seems to just be an issue with the wet theme itself.

We could remove the `wb-so` id and then add in the padding and margin and text align ourselves. The buttons would then have some rounded borders though.

The `wb-so` stands for "Sign On" section, which this section technically isn't, it is more of a toolbar. So not sure how they want it to look, the hover states have just been bothering me for a long time.

JVickery-TBS: I can try using these classes to get the correct alignments:
https://wet-boew.github.io/wet-boew-styleguide/design/proximity-en.html
https://wet-boew.github.io/wet-boew-styleguide/design/alignment-en.html

JVickery-TBS: @wardi okay, I was able to use some other wet classes in the templates.

wardi:
Thank you @JVickery-TBS

PR #1356, opened 2023-02-17
Even, Even More CKAN 2.9 Compatibilities

JVickery-TBS:
- new `ckan.auth.public_user_details` option prevents public user routes from being shown to non-authed users. This does NOT have any negative effects on password resets.
- better labels for some of the password fields for edit user and new user forms.
- replaced deprecated `ckan.plugin.toolkit.c` with `ckan.plugin.toolkit.g`. Should we replace all usages of `c` with `g` throughout the repo?
- started working on the views feeds again as we will need those, more to come with those...


ckanext/canada/templates/internal/user/edit_user_form.html
```diff
@@ -33,9 +33,9 @@
                   attrs={'autocomplete': 'off', 'class': 'form-control'}
                   ) }}
 
-    {{ form.input('password1', type='password', label=_('Password'), id='field-password', value=data.password1, error=errors.password1, classes=['control-medium'], attrs={'autocomplete': 'off', 'class': 'form-control'} ) }}
+    {{ form.input('password1', type='password', label=_('New Password'), id='field-password', value=data.password1, error=errors.password1, classes=['control-medium'], attrs={'autocomplete': 'off', 'class': 'form-control'} ) }}
 
-    {{ form.input('password2', type='password', label=_('Confirm Password'), id='field-password-confirm', value=data.password2, error=errors.password2, classes=['control-medium'], attrs={'autocomplete': 'off', 'class': 'form-control'}) }}
+    {{ form.input('password2', type='password', label=_('Confirm New Password'), id='field-password-confirm', value=data.password2, error=errors.password2, classes=['control-medium'], attrs={'autocomplete': 'off', 'class': 'form-control'}) }}
```
wardi, line 38: Are these aligning with upstream strings or were they requested by our side?

JVickery-TBS: @wardi neither, just semantic labels. Or do you think the legend of the field set is fine enough? Or did you want this suggested upstream?

wardi: I ask because if it isn't upstream then we need to provide translations in our extension. Usually I leave the strings alone unless the business team asks for a change, IIRC they are in the process of reviewing the site now so it could be left to them.

JVickery-TBS: @wardi template changes have been reverted for the password labels.

wardi:
@JVickery-TBS looks good to me

PR #1359, opened 2023-03-03
OPEN-2353: Update the Similar Records section of the Package read tem…

thriuin:
…plate to use Search v.2

### DIFF
--- a/ckanext/canada/helpers.py
+++ b/ckanext/canada/helpers.py
@@ -227,6 +227,9 @@ def portal_url():
 def adv_search_url():
     return config.get('ckanext.canada.adv_search_url_fr') if h.lang() == 'fr' else config.get('ckanext.canada.adv_search_url_en')
 
+def adv_search_mlt_root():
+    return "{0}/donneesouvertes/similaire/".format(config.get('ckanext.canada.adv_search_url_fr')) if h.lang() == 'fr' else "{0}/opendata/similar/".format(config.get('ckanext.canada.adv_search_url_en'))
+
 def googleanalytics_id():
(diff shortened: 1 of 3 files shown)


wardi:
If the settings aren't defined we'll get an error rendering the page, want to add the production urls as defaults to the `get` calls?

wardi:
oh, not an error, just a broken url with "None" in it

thriuin:
Oops, already pulled the trigger on  this. If the setting is not defined, I thought the test on line 233 would prevent the block from rendering. If that isn't the case, we could add production as a default.

wardi:
233 will always be true because a string is always returned from the helper (it would start with "None/..." if the appropriate setting isn't defined)

This makes me wonder about the best way to manage multilingual urls for external services. We could use gettext on them like other things instead of if/else in the code. Still need multiple settings in the ini file for different domains though.

If we follow your lead on multilingual urls in the ckan portion of the site I can see making the `url_for` logic in ckan language-aware and provide a bunch of "translations" for url components and have them automatically registered as alternate urls. That way most of ckan would just work for multilingual links and redirects, and it would even be possible to customize core urls with translations instead of code overriding.

thriuin:
Ah, looking more closely at it now, I realize this won't work. For now, a default to the portal is probably best. I could also extend Search to provide a meaningful message when the UUID can't be found in Solr. Right now it will just return an empty list.
I definitely like your idea for translations for the urls. That definitely seems like a better long-term approach than hard-coding EN and FR toggles. Having different domains is an irritating complication however - I think there is a reason you don't see it very often.

PR #1361, opened 2023-03-16
Use `body_contains` Helper for Webform Tests

JVickery-TBS:
fix(tests): webform helper usage;

- Made use of core `body_contains` helper.


wardi:
-1 on this for our code for these reasons:

- the called function isn't part of ckan's official plugin interface
- it does exactly the same thing we were doing before
- calling `body_contains` is longer to type and less obvious
- test failures will show as being inside this function instead of inside the actual test next to the thing being tested

JVickery-TBS:
Good enough for me. Chimi-chucking it in the bin.

PR #1362, opened 2023-03-16
Removed Old API Key Generation Ability

JVickery-TBS:
feat(users): api keys to tokens;

- Removed links to generate old API keys.
- Added button links to the new tokens page.

As discussed, we are going to prevent the generation of the old user API Keys, and try to move users to create the new API Tokens. We need to keep the old API Keys still to allow users to use the API with them. And we will keep the "Show API Key" popup link still for users that already have a generated old API Key.

To discuss:
* What will the communication to users be with the move from API Keys to Tokens?
* Will we need to add more messaging in the user pages to help with API Token generation and usage?
* Will we want to eventually fully discontinue the old API Keys?

### DIFF
--- a/ckanext/canada/templates/internal/user/edit_user_form.html
+++ b/ckanext/canada/templates/internal/user/edit_user_form.html
@@ -39,17 +39,17 @@
   </fieldset>
 
   <div class="form-actions">
+    {{ form.required_message() }}
     {% block delete_button %}
       {% if h.check_access('user_delete', {'id': data.id})  %}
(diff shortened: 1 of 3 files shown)


ckanext/canada/templates/internal/user/edit_user_form.html
```diff
@@ -39,17 +39,17 @@
   </fieldset>
 
   <div class="form-actions">
+    {{ form.required_message() }}
```
wardi, line 42: why is a button a required field?

wardi: or you're just moving the location of this text up in the form?

JVickery-TBS: @wardi Yeah just moving it up as it was in between buttons before.

PR #1366, opened 2023-03-23
Tests for package_search on Registry and Portal

JVickery-TBS:
feat(tests): added package_search tests;

- Added test class for package_search on Registry.
- Added test class for package_search on Portal.

### DIFF
--- a/ckanext/canada/tests/test_search.py
+++ b/ckanext/canada/tests/test_search.py
@@ -1,0 +1,213 @@
+# -*- coding: UTF-8 -*-
+from ckanapi import LocalCKAN
+import ckan.plugins as p
+
+from ckan.tests.helpers import FunctionalTestBase
+from ckan.tests.factories import Sysadmin
+from ckanext.canada.tests.factories import (
+    CanadaOrganization as Organization,
+    CanadaUser as User,
+    CanadaDataset as Dataset
+)
+
+
+class TestRegistrySearch(FunctionalTestBase):
+    """
+    Class to test the package_search functionality for the Registry.
+    """
+    def setup(self):
+        super(TestRegistrySearch, self).setup()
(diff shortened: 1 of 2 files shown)


ckanext/canada/tests/test_search.py
```diff
@@ -0,0 +1,213 @@
...
+        self.user_lc = LocalCKAN(username=user['name'])
+        self.editor_lc = LocalCKAN(username=editor['name'])
+        self.system_lc = LocalCKAN(username=sysadmin['name'])
+        #
+        # Datasets 0, 1, 2, and 3
+        # 1 & 3 will be ready to publish
+        # 3 will not have a portal release date
+        # datasets belong to self.org
+        #
+        for _i in range(0, 4):  # 0-3
```
wardi, line 39: `i` is a common loop variable, why `_i`?

JVickery-TBS: @wardi I just personally use `_<var name>` for local scope. And the `_` before the variable in the loop will ignore any typing for the basic loop.

wardi:
Looks good. These will be great to have.

Next bits of core functionality we don't have covered are requesting user accounts and working with the PD UI. Make some tickets for the backlog?

ckanext/canada/tests/test_search.py
wardi: in Python `_<var name>` means a name that shouldn't be imported from outside modules, so it's kind of funny looking. What do you mean about ignoring typing?

wardi: All variables in python are local unless you don't declare them or use `nonlocal foo` or `global foo` in the function, and there is a convention of *not* prefixing variable names for scope, type or anything else. There are some conventions around capitalization, however.

PR #1368, opened 2023-04-12
use plugin_loaded for is_registry

wardi:
see https://github.com/ckan/ckan/pull/7526#pullrequestreview-1381582534

### DIFF
--- a/ckanext/canada/helpers.py
+++ b/ckanext/canada/helpers.py
@@ -1,5 +1,6 @@
 import json
 import re
+from ckan.plugins import plugin_loaded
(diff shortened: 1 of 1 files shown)


wardi:
@JVickery-TBS same test is failing on canada-v2.9 so it looks unrelated

JVickery-TBS:
@wardi Looks good. Yeah I don't think we need to worry about these test failing in this PR for now. Will fix in the `canada-v2.9` branch when doing the other tests.

JVickery-TBS (approved):

PR #1369, opened 2023-04-18
Resource Format Auto Detect

JVickery-TBS:
feat(schemas): use core guess_format for resource format schema;

- Made resource format field read only.
- Used core `if_empty_guess_format` validator for the field.
- Modified choices into `openness_scores` for the Openness Score mappings on the format values.

I think we would like to have a custom validator, as the core one is `if_empty_guess_format`. I would like to keep this field read only as it seems like so many users do not select or type in the correct file formats.

So there are the scenarios now that if a user cleared an upload or change it to a link, then the `if_empty_guess_format` would not fire off again hence the `if_empty`. So doing a `guess_format` which would just always set the format correctly.

If the resource is a link and not upload however, what should we set the format to?


wardi:
Need business team support for this kind of a change, maybe you could demo it at dev chat?

JVickery-TBS:
@wardi okay I worked on this concept a bit more and I think it is in a good place to show.

The format field is now a select2 with the normal endpoint to the format autocomplete. However, I override the callback to get the choice values from our schema and match those without case sensitivity.

And from my initial test, I was uploading a JPG image, and the `if_empty_guess_format` validator was setting it to `JPEG` which we do not have in our choices. So I used the existing `replaces` we have in our schema, and made a validator which matches against those. E.g. `if_empty_guess_format` would fire off and return `JPEG`, that would go through the new validator and match the `replaces: [jpeg]` and return `JPG`. We would probably have to keep adding to the replaces lists over time.

I have also removed the EXE format from the choices during this.

JVickery-TBS:
As discussed, we are going to plan to get some front-end stuff happening. Which includes:
* localizing scheming options to JS
* using JS callback for select2
* not allowing select2 custom input
* clearing format field when clearing upload or change of upload/url field
* guessing format during upload/url field update
* run guess format through scheming replacement matches
* run scheming choices validation
* simplifying front-end error message for format as it will have a select2 now

We will also have to confirm that the guess format works in the API. We should not change/simplify the API error message however, as that will NOT have a select2/type-ahead. We will also want to do something about the clear_upload in the back-end as well.

PR #1371, opened 2023-04-24
More WebForm Tests

JVickery-TBS:
Adding tests for new user form.

Adding tests for all of the Recombinant related UI and forms.

### DIFF
--- a/ckanext/canada/templates/internal/package/wet_datatable.html
+++ b/ckanext/canada/templates/internal/package/wet_datatable.html
@@ -78,7 +78,8 @@
       'colvis',
       {extend:"selected",text:"{{_('Edit in Excel')}}",action: function ( e, dt, button, config ){
 
-          var form = $('<form>',{
+          // closing form tag needed for py webtest lib
+          var form = $('<form></form>',{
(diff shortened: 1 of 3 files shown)


JVickery-TBS:
@wardi I have some placeholder test methods in there now for the Recombinant test class. I think those should be all the ones we need?

But my main question is if we need those tests for each PD type? As the create single record and the validation tests would be different for each PD type? That way we would test to make sure that all of our table/schema stuff for the PD types are correct?

Let me know if you think we should have these tests for each PD type. I could try abstracting the class to subclass it for each PD type.

wardi:
There's no need to abstract pd tests. Better to keep them as simple (and repetitive) as possible or we'll have to start writing tests for our tests.

wardi:
Having tests of the forms for one PD type is sufficient because we have API tests for all the types that should cover all the specific validation rules.

JVickery-TBS:
@wardi okay all of these tests should be good now. These should now cover the user registration UI and the PD UIs (per role type depending on the test e.g. bulk-delete and xls_upload).

wardi:
Looks great @JVickery-TBS really happy to have some coverage on this functionality

PR #1374, opened 2023-05-08
Turn Off DAC Member Download CRON Workflow

JVickery-TBS:
Commented out the CRON schedule for the dac member download github workflow. Also refactored from `main`.yml to `dac.yml`.

Added ssl verify option to the dac member list download python script. This was because the python ssl lib could not verify our custom cert from github servers.


JVickery-TBS:
@wardi commented code has been removed.

PR #1376, opened 2023-05-11
Remove Old Org Title Validators

JVickery-TBS:
feat(dev): removed old org title validators;

- Removed org title save validator.
- Removed org title output validator.


JVickery-TBS:
Closing this, as it does not hurt the site at all, and there may be other systems and client systems which rely on the format of the org titles (`English | French`)

PR #1379, opened 2023-05-23
Redirecting "contact us" footer lc messages

DRH-TBS:
Hoping I updated all the "Contact Us" links directly to state "All Contacts"


wardi:
better to not use special characters like `"` in branch names.

wardi:
@DRH-TBS please change the branch name and squash these little changes into a single commit with a good commit message.

PR #1380, opened 2023-05-23
Redirect contact

DRH-TBS:


wardi:
By "with a good commit message" I mean create a commit with a better commit message than "update footer.html" and "update ckanext-canada.po". I know it's a small thing but it makes it much easier when we need to go back through the commit history when rebasing or debugging.

wardi (changes requested):
By "with a good commit message" I mean create a commit with a better commit message than "update footer.html" and "update ckanext-canada.po". I know it's a small thing but it makes it much easier when we need to go back through the commit history when rebasing or debugging.

Also "redirect" has a specific technical meaning in web applications. How about "Rename contact link to All contacts" instead.

PR #1382, opened 2023-06-16
Jobs to Remove Validation and Datastore Tables for Unsupported Resource Types

JVickery-TBS:
feat(dev): implemented IDomainObjectModification;

- Implemented `IDomainObjectModification` instead of `IResourceUrlChange`.
- Queued new job callback for unsupported datastore/validation resources.

### DIFF
--- a/ckanext/canada/plugins.py
+++ b/ckanext/canada/plugins.py
@@ -10,6 +10,9 @@ import ckan.lib.helpers as hlp
 from ckan.logic import validators as logic_validators
 from routes.mapper import SubMapper
 from paste.reloader import watch_file
+from ckan.model.domain_object import DomainObjectOperation
+from ckan.model.resource import Resource
+
 
 from ckantoolkit import h, chained_action, side_effect_free, ValidationError, ObjectNotFound
 import ckanapi
@@ -21,6 +24,7 @@ from ckanext.canada import auth
 from ckanext.canada import helpers
 from ckanext.canada import activity as act
 from ckanext.extendedactivity.plugins import IActivity
+from ckanext.validation.settings import SUPPORTED_FORMATS as SUPPORTED_XLOADER_FORMATS
 
 import json
 
@@ -48,7 +52,7 @@ class DataGCCAInternal(p.SingletonPlugin):
     p.implements(p.IRoutes, inherit=True)
     p.implements(p.IPackageController, inherit=True)
     p.implements(p.IActions)
-    p.implements(p.IResourceUrlChange)
+    p.implements(p.IDomainObjectModification)
 
     def update_config(self, config):
         p.toolkit.add_template_directory(config, 'templates/internal')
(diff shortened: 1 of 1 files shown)


wardi:
Need to be really careful with this one. This plugin endpoint is really low level and gets called in lots of different places. Are there performance impacts of adding this here? Does this fill the queue with jobs that aren't needed?

Is this better implemented as a cli command that can report on or clean up validation rows/datastore tables? Should this be in core and the validation extension instead of our theme plugin?

As written this change looks like it will fill our logs with failed deletions when validation rows or datastore tables never existed.

JVickery-TBS:
@wardi This seemed to run asynchronously from the web requests, so as a front-end user I did not see any performance issues.

I did come across a scenario in which it would cause a lot of jobs, and that was without the `Resource` check, so it was queueing jobs for the Packages and then it screamed.

But the jobs will only get queued if the Resource format is not in the supported xloader formats OR if the url has changed for Resources. And it will only happen for changes, not new or deleted Database actions.

I think we could put in more conditions to really limit the job enqueueing here even more. ` remove_outdated_resource_from_datastore` callback checks if the url_type is NOT upload, so we can do that check before even enqueuing the job as well. But of course, this job only gets queued if a Resource's url is changed.

But I did a lot of updates, deletions, and creations and the jobs only got enqueued whenever the change action happened and the url changed or the format was not in the xloader supported ones.

Do you have any tools to test performance of ckan?

ckanext/canada/plugins.py
```diff
@@ -963,3 +983,38 @@ def remove_outdated_resource_from_datastore(resource_id):
...
+    if res.get(u'format', u'').lower() not in SUPPORTED_XLOADER_FORMATS:
+        log.info('Unsupported resource format "{}". Deleting validation reports and datastore tables for resource {}'
+                 .format(res.get(u'format', u'').lower(), res['id']))
+        try:
+            lc.action.resource_validation_delete(resource_id=res['id'])
+            log.info('Validation report deleted for resource %s' % res['id'])
+        except ObjectNotFound:
+            log.error('Validation report for resource %s does not exist' % res['id'])
+
+        if res['url_type'] == 'upload' or res['url_type'] == '':
```
JVickery-TBS, line 1015: @wardi would it be prudent to add `and res['datastore_active']` to this condition?

wardi: yes, but I still think this feature shouldn't be in the canada extension: it's much better implemented in xloader and ckanext-validation so that the community can help maintain it.

JVickery-TBS: @wardi Okay I will go see about putting it in the xloader extension. Do you think it should be in this same hook that fires before the database commit?

wardi: That's your call, if there's a more obvious place to incorporate it into xloader then you could change it.

Oh, but for validation I guess removing a table does only make sense for our site (because we prevent xloader runs until validation passes). That could be a small override of the decision in xloader to schedule a load or table deletion just on our fork.

JVickery-TBS: Okay cool, that makes it a bit easier to implement upstream if we are only dealing with the Datastore tables there. Probably going to put it behind a config option like `ckanext.xloader.limit_datastore_tables`. Using `ckanext.xloader.formats` to check the supported formats for Xloader.

I can add a similar thing to the validation extension which would handle the removal of the validation reports based on the supported validation types (`ckanext.validation.formats`). And have another similar config option to control this like `ckanext.validation.limit_validation_reports`.

PR #1383, opened 2023-06-16
Contracts disclosure to publication

DRH-TBS:


ckanext/canada/tables/contracts.yaml
```diff
@@ -1,31 +1,34 @@
...
+    données qui sont soumis à des règles de validation des données conformément
+    au Guide de la publication proactive des marchés.
+    Les utilisateurs devraient examiner les règles de validation des données dans
+    l’onglet de référence du modèle pour chaque champ de données sous «
+    validation » pour plus d’informations afin de s’assurer que les données peuvent
+    être téléchargées dans le registre.
 
 template_version: 3
-template_updated: 2023-03-14
+template_updated: 2023-08-14
```
wardi, line 24: August?

wardi:
There are a number of other changes to fields coming as well, right?

ckanext/canada/tables/contracts.yaml
DRH-TBS: That is what the client indicated. Copied from the document.

DRH-TBS:
> There are a number of other changes to fields coming as well, right?

Yes, more changes for another ticket due next week.

PR #1384, opened 2023-06-21
Contracts text changes

DRH-TBS:
Implemented text changes for contracts.yaml and contractsa.yaml

There were numerous instances for titles, descriptions, and format types where there was no French (fr:) element and so I added them. I hope that was the correct thing to do!

### DIFF
--- a/ckanext/canada/tables/contracts.yaml
+++ b/ckanext/canada/tables/contracts.yaml
@@ -80,7 +80,7 @@ front_matter:
 
 
 resources:
-- title: Proactive Disclosure - Contracts over 10K
+- title: Proactive Publication – Contracts over $10,000 
   resource_name: contracts
 
   excel_data_num_rows: 5000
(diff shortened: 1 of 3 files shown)


ckanext/canada/tables/contracts.yaml
```diff
@@ -481,8 +459,12 @@ iii. Quand un contrat porte sur plus d’un article économique, il est recomman
...
-      fr: "Il est recommandé d’utiliser les commentaires normalisés (consultez l’annexe C). Évitez d’utiliser des acronymes dans cette zone. Les ministères sont invités à fournir une brève description de chaque marché. Lorsqu’un marché comprend un code d’article économique précisé à l’annexe D, les ministères doivent fournir une brève description du marché pour que le public puisse bénéficier du contexte additionnel (par exemple, le commentaire accompagnant un marché de services de consultation en gestion pourrait se lire ainsi : « Étude de faisabilité portant sur une norme en matière de bâtiments écologiques devant s’appliquer aux projets ministériels »)."
+      en: i. Standardized comments are recommended to be used (refer to Appendix C). Avoid the use of acronyms within this field.
```
wardi, line 462:
```suggestion
      en: |
          i. Standardized comments are recommended to be used (refer to Appendix C). Avoid the use of acronyms within this field.
```

ckanext/canada/tables/contracts.yaml
```diff
@@ -80,7 +80,9 @@ front_matter:
 
 
 resources:
-- title: Proactive Disclosure - Contracts over 10K
+- title: 
+    en: Proactive Publication – Contracts over $10,000 
+    fr: Publication proactive – Contrats attribués de plus de 10 000 $
```
wardi, line 85: for technical reasons these titles need to be English only and are translated in the .po file (same for contractsa below)

PR #1386, opened 2023-06-28
View on Portal States (BS Alerts and Button)

JVickery-TBS:
image (1)
image


wardi (approved):

JVickery-TBS:
Just waiting for translations for the new strings

wardi:
looks good

JVickery-TBS:
Added another new alert for "Record missing approval from the Information Management Senior Official (IMSO)"

And updated the `imso_approval` field label and description to include IMSO. Will most likely be changing the test once it is finalized. And then adding in the translations later.

ckanext/canada/templates/internal/package/read.html
PatLittle, line 1: I think the button&messages are missing the /dataset/ path

JVickery-TBS: fixed

PR #1392, opened 2023-07-06
Prevent Validation Schema URLs

JVickery-TBS:
feat(schema): prevent validation schema URLs;

- Removed Validation Schema URL field from form.
- Extended `resource_schema_validator` validator to prevent URLs.
- Added catch to `resource_schema_validator` for bad Validation Schema JSON.


JVickery-TBS:
@wardi we do not actually need the template for this to work. When uploading a file, the JS from the validation extension will show the URL field but it changes the label to "Uploaded Data Schema".

And with our new validator, we prevent HTTP(s) values. So I think this should be good enough.

wardi (approved):

ckanext/canada/tests/__init__.py
```diff
@@ -0,0 +1,5 @@
+from ckanext.validation.model import create_tables, tables_exist
+
+def canada_tests_init_validation():
+ if not tables_exist():
+  create_tables()
```
wardi, line 5: 1-space indents?

wardi: This would be better in a FunctionalTestBase subclass or a fixture for our tests

ckanext/canada/tests/test_travela.py
```diff
@@ -1,50 +1,52 @@
...
+                resource_id=self.resource_id,
+                records=[record])
+        err = ve.exception.error_dict['records'][0]
+        expected = {
+            'year': [
+                'This must list the year you are reporting on (not the fiscal year).'],
+        }
+        for k in set(err) | set(expected):
+            assert_equal(err.get(k), expected.get(k), (k, err))
+
```
wardi, line 52: did you change all the newlines in this file or something?

ckanext/canada/tests/__init__.py
JVickery-TBS: Huh, my editor config is set to `*.py` but must not work on the `__init__.py` files, interesting.

Will make into a fixture if 2.8 + py2 allows it. Otherwise will make it into a subclass.

ckanext/canada/tests/test_travela.py
JVickery-TBS: Probably just my system saving the file with linux new lines, yeah.

ckanext/canada/tests/__init__.py
JVickery-TBS: @wardi Fixture not supported for this version, so just sub classed it as `CanadaTestBase`

PR #1395, opened 2023-07-24
IDataValidation implementation for Registry 

JVickery-TBS:


ckanext/canada/plugins.py
```diff
@@ -49,6 +49,8 @@ class DataGCCAInternal(p.SingletonPlugin):
     p.implements(p.IPackageController, inherit=True)
     p.implements(p.IActions)
     p.implements(p.IResourceUrlChange)
+    from ckanext.validation.interfaces import IDataValidation
+    p.implements(IDataValidation, inherit=True)
```
wardi, line 53: This would avoid a failure if ckanext-validation isn't present
```suggestion
    try:
        from ckanext.validation.interfaces import IDataValidation
    except ImportError:
        log.warn('failed to import ckanext-validation interface')
    else:
        p.implements(IDataValidation, inherit=True)
```

PR #1387, opened 2023-06-28
Babel Config File & Updated PO's

JVickery-TBS:
Added the `setup.cfg` file used by Babel to do message extraction and cataloging for translation files.

Ran `extract_messages` with the new babel config.

Ran `update_catalog` for `en` and `fr` locales with the new babel config.

PR #1397, opened 2023-08-02
2.9 Messages and Updated Catalogs

JVickery-TBS:
Merged in feature/i18n-babel-config from https://github.com/open-data/ckanext-canada/pull/1387

Solved conflicts.

Fixed fuzzies.

Added missing/new translation strings.


wardi:
We just merged translations into master right? Did those get merged back into 2.9 before this PR?

JVickery-TBS:
@wardi no, did not merge them into master yet. Just was seeing if conflicts would happen if merging into 2.9 branch here.

Should I merge this PR https://github.com/open-data/ckanext-canada/pull/1387 first??

And then just merge master into 2.9 branch and then deal with the conflicts that arose in this branch?

wardi:
If we don't need the translations on 2.8 then we can only merge this one and don't bother with 2.8. Or if we're sure that all 2.8 translations are included in the 2.9 ones we can do a merge where we replace the .po files instead of resolving them after applying translations to both.

This is why we used to add translations the dumb way (add and update the .po file manually) because merging auto-generated changes interspersed with changes we care about is super-painful.

PR #1401, opened 2023-08-31
More Generic Scheming Reapeating Field Removal Text

JVickery-TBS:
fix(lang): scheming repeating field removal text;

- More ambiguous copy.


JVickery-TBS:
Just waiting on French translations.

wardi:
looks good once fr translations are added

PR #1403, opened 2023-09-01
Fix Github Tests

JVickery-TBS:
Fix/GitHub tests system depends


JVickery-TBS:
@wardi I am away on Tuesday, the tests seem to be broken currently.

I was able to fix the one issue with the `Pillow` dependency from ckan dev-requirements.

But the user form tests seem to be returning 403s in the test app instead of the user register pages.

PR #1404, opened 2023-09-06
PD Type Translations

JVickery-TBS:
feat(i18n): translations for PD table yaml files;

- Included strings and translations from the recombinant table yaml files.


wardi:
If you want to get fancy and reduce some duplication you could extract these strings from the yaml files with a custom babel message extractor

JVickery-TBS:
depends on https://github.com/open-data/ckanext-recombinant/pull/116

JVickery-TBS:
@wardi custom message extractor is done now.

ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
```diff
@@ -4882,3 +5161,268 @@ msgstr "Mémoriser mes informations"
...
+#~ "code in A1A format or the value \"NA\"\n"
+#~ msgstr ""
+#~ "Ce champ doit contenir les trois "
+#~ "premiers chiffres d’un code postal dans"
+#~ " le format A1A ou la valeur «"
+#~ " NA. »"
+
+#~ msgid "Please enter a decimal value between zero and one: %"
+#~ msgstr ""
+
```
wardi, line 5428: is this where we accumulate old unmatched translations?

ckanext/canada/extract.py
```diff
@@ -0,0 +1,97 @@
...
+    while line:
+        if phrase in line:
+            return num
+        line = fileobj.readline().decode(encoding)
+        num += 1
+    return None
+
+
+def _extract_string(fileobj, key, value, encoding, comments):
+    fileobj.seek(0)
```
wardi, line 17: instead of repeatedly re-parsing the file ourselves for every string can we use something like the answer to https://stackoverflow.com/questions/13319067/parsing-yaml-return-with-line-number ?

JVickery-TBS: Seems like all of these are based on yaml Nodes that can have keys to them. So far, it seems that the string constructor really is forcing the return of a single value...but will keep trying

ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
JVickery-TBS: Okay this seems to be solved now with the Loader class. albeit we have never had a translation for this one, so waiting on that still.

ckanext/canada/extract.py
JVickery-TBS: @wardi okay well it is a bit strange. Sadly it was really impossible to record the line numbers for str type nodes because of how they are outputted and encoded into the dict later on.

But I was able to override the scalar constructor which the str constructor builtin calls. And then I just recorded line numbers into a class variable. And then return the loader with the `load` method and access the class variable whilst yielding the strings to get the line numbers.

ckanext/canada/extract.py
```diff
@@ -0,0 +1,86 @@
...
+            if isinstance(nil_title, string_types):
+                for line_number in line_numbers.get(nil_title, [0]):
+                    yield (line_number, '', nil_title, ['NIL Resource Title for PD Type: %s' % pd_type])
+
+            nil_triggers_strings = resources[1].get('trigger_strings')  # nil resource sql error messages
+            if nil_triggers_strings:
+                for k, v in nil_triggers_strings.items():
+                    if isinstance(v, string_types):
+                        for line_number in line_numbers.get(v, [0]):
+                            yield (line_number, '', v, ['SQL Trigger String for PD Type: %s' % pd_type])
```
wardi, line 86: I'd rather you loop over the resources instead of coding in the assumption that there will only be two and the second one is always a nil record

ckanext/canada/extract.py
```diff
@@ -0,0 +1,86 @@
...
+    def construct_scalar(self, node):
+        if node.value in self.line_numbers:
+            self.line_numbers[node.value].append(node.start_mark.line + 1)
+        else:
+            self.line_numbers[node.value] = [node.start_mark.line + 1]
+        return super(SafeLineLoader, self).construct_scalar(node)
+
+    def get_single_data(self):
+        data = super(SafeLineLoader, self).get_single_data()
+        return self.line_numbers, data
```
wardi, line 20: Great, this is much nicer than searching across the yaml file to try to locate strings.

ckanext/canada/extract.py
JVickery-TBS: Now just looping through all the resources.

wardi: looks good, thanks

JVickery-TBS:
Trying to get our test moved to circleci. (https://github.com/open-data/ckanext-canada/tree/circleci)

Ran into a roadblock and had to open up a support ticket with circleci. But hopefully that can get solved and we can figure out why these string triggers are failing.

JVickery-TBS:
@wardi okay cool, so with that fix to Recombinant and some minor fixes in the pd tests and yaml files here the tests are all working now, yay!

Are you able to take a final look at this to confirm it is all good to go?

wardi (approved):
are the `  # yaml | not supported` for a linter or something?

JVickery-TBS:
@wardi haha no, they are just for some inline comments for us :D

PR #1405, opened 2023-09-13
Added Missing Confirm Action Views

JVickery-TBS:
fix(views): added missing confirm action pages;

- Added missing confirm member delete.
- Added missing confirm datastore table delete.
- Added missing flash for datastore table delete.


wardi (approved):
Looks good, is the delete datastore table feature not upstream yet?

JVickery-TBS:
@wardi does not look like the delete datastore is anywhere upstream in `datastore`, `datapusher`, or `xloader`.

wardi:
seems like an occasionally useful feature, we could submit it to xloader and datapusher since those create the page it appears on. Something for the backlog

PR #1406, opened 2023-09-13
Datatables Loading Icon

JVickery-TBS:
feat(templates): loading icon for datatables;

- Added html loading icon and behavior for datatables.


wardi:
if you use webassets for these files instead then then they will be given cache-busting urls automatically so we don't have to do the silly hack like we have for the canada1.css file

wardi:
Are you adding this to the ckanext-canada extension because of a BS/WET compatibility issue that would be more complicated to fix in ckan?

JVickery-TBS:
@wardi I did it in the Canada Ext because we had to change the colors, and I personally wanted to reposition the spinner. I also wanted to have the spinner visible on first initialization of the datatable.

I can fix this upstream however if you would want. I just assumed that the behavior was intended, as the `processing` setting of the datatables was turned off and there was no data attribute coded in to control this feature. So thought it was on purpose.

https://github.com/open-data/ckan/commit/b08efcb7508296b69df7936c0791de3c0dc1fbfc <- the note in the commit is `Disable processing spinner - didn't automatically disappear`. Which I have never found to be the case.

We would still probably want to keep the code in this PR for the positioning of it. But in the upstream, I would be adding new datat attributes to the `datatables_view` ckan js module to allow for the spinner to be enabled, and a preInit spinner as well.

wardi:
Sure, this makes sense here if it's for matching our theme.

JVickery-TBS:
@wardi okay it is using webassets now. I think I did it right-ish

wardi:
try moving webassets to an `assets` folder so that they aren't also served as public files

JVickery-TBS:
@wardi okay they are now in an `assets` directory

wardi (approved):

PR #1411, opened 2023-10-18
Add nil report to briefing notes

RabiaSajjad:


wardi:
@JVickery-TBS recently added support for collecting strings for translation from pd files using `python setup.py extract_messages` and `python setup.py update_catalog`, this will add the new strings to the pot and po files so that translation can be added for the resource titles and any other required strings

ckanext/canada/tables/briefingt.yaml
```diff
@@ -303,6 +303,142 @@ resources:
...
 
+- title: Proactive Publication - Briefing Note Titles and Numbers Nothing to Report
+  resource_name: briefingt-nil
+  create_form: true
+  edit_form: false #edit not possible for nil reports since only stored data is primary key
+  fields:
+
+  - datastore_id: year
+    datastore_type: year
+    label: Year
```
wardi, line 314: Should this say "Fiscal Year" and represented ? Does "2022 P10" mean 2023-01 or 2022-01? Seems likely to be confusing

RabiaSajjad: Good point @wardi. For this template @PatLittle has confirmed it is calendar year.

PR #1413, opened 2023-10-19
expand textboxes for fields in presets

RabiaSajjad:


wardi:
Do you know if we're missing some classes on our input boxes to make them look like wet inputs?

RabiaSajjad:
> Do you know if we're missing some classes on our input boxes to make them look like wet inputs?

I have applied the missing `form-control` class

ckanext/canada/internal/static/internal.css
```diff
@@ -660,3 +660,22 @@ div.resource-upload-field .upload-type > label.control-label::before{
   color: #c6898b;
   margin-right: 2.5px;
 }
+.form-group .info-block {
+    padding: 0;
+    font-size: small;
+    color: #2572b4;
+}
+label.control-label:has(span.control-required):after {
+    content: "(required)";
```
wardi, line 669: does this need a translation?

ckanext/canada/templates/internal/base.html
```diff
@@ -8,7 +8,7 @@
 {# Remove when WET v4.0.21 is out #}
 {%- block custom_styles %}
   {{ super() }}
-  <link rel="stylesheet" href="/internal.css?x=10" type="text/css" />
+  <link rel="stylesheet" href="/internal.css?x=11" type="text/css" />
```
wardi, line 11: If we used webassets for our css we wouldn't need this hack anymore

PR #1415, opened 2023-10-25
Webassets, WET/BS Danger Alerts, and Password Reset Button

JVickery-TBS:
fix(templates): webassets, bs alerts, user reset template;

- Fixed alerts to wet/bs version html structure.
- Fixed positioning of core alert close button.
- Moved internal and public css static files to webassets.
- Changed `alert-error` to `alert-danger` for wet/bs.
- Fixed submit button for user password reset template.


wardi (approved):

wardi (approved):
love to see it, no more manual cache-busting and hard-coded urls

PR #1417, opened 2023-10-26
Organization List Search and Org Meta Stats Fix

JVickery-TBS:
fix(templates) organization list and read;

- Added in organization search for org index.
- Did better condition for org member and package count to prevent possible errors.

### DIFF
--- a/ckanext/canada/templates/public/organization/snippets/organization_search.html
+++ b/ckanext/canada/templates/public/organization/snippets/organization_search.html
@@ -1,0 +1,26 @@
+<form class="form-inline mrgn-bttm-sm mrgn-tp-md" role="form" method="get" data-module="select-switch">
+  <div class="input-group col-md-12">
+    <input type="text" class="form-control" id="field-organization-search" autocomplete="off" name="q"
+           value="{{ q }}" placeholder="{{ _('Search organizations...') }}">
+    <span class="input-group-btn">
+      {% block search_input_button %}
+      <button class="btn btn-primary btn-small" type="submit">
+        <i class="fa fa-search" aria-hidden="true"></i>
+        <span class="wb-inv">{{ _('Search') }}</span>
+      </button>
+      {% endblock %}
+    </span>
+  </div>
+</form>
+
+<div class="row mrgn-bttm-lg">
+  {% if count==0 %}
+    <div class="col-md-12 text-center text-warning">
+      <strong>{% snippet 'snippets/search_result_text.html', query=q, count=count, type="organization" %}</strong>
+    </div>
+  {% else %}
+    <div class="col-md-6 mrgn-tp-sm text-left text-success">
+        <strong>{% snippet 'snippets/search_result_text.html', query=q, count=count, type="organization" %}</strong>
+    </div>
+  {% endif %}
+</div>
--- a/ckanext/canada/templates/public/organization/snippets/organization_search.html
+++ b/ckanext/canada/templates/public/organization/snippets/organization_search.html
@@ -14,13 +14,8 @@
 </form>
 
 <div class="row mrgn-bttm-lg">
-  {% if count==0 %}
-    <div class="col-md-12 text-center text-warning">
-      <strong>{% snippet 'snippets/search_result_text.html', query=q, count=count, type="organization" %}</strong>
-    </div>
-  {% else %}
-    <div class="col-md-6 mrgn-tp-sm text-left text-success">
-        <strong>{% snippet 'snippets/search_result_text.html', query=q, count=count, type="organization" %}</strong>
-    </div>
-  {% endif %}
+  <div class="{{ 'col-md-12 text-center text-warning'
+      if not count else 'col-md-6 mrgn-tp-sm text-left text-success' }}">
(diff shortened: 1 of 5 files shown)


ckanext/canada/templates/public/snippets/organization.html
```diff
@@ -25,22 +25,24 @@
...
+            </p>
+          </div>
+          <div class="col-md-6">
+            <h4 class="text-center">{{ _('Datasets') }}</h4>
+            <p class="font-xlarge text-center">
+              {{ h.SI_number_span(organization.package_count) }}
+            </p>
+          </div>
+        {% endif %}
       </div>
```
wardi, line 44: prevent empty row divs?
```suggestion
      {% if member_count >= 0 %}
        <div class="row">
          <div class="col-md-6">
            <h4 class="text-center">{{ _('Members') }}</h4>
            <p class="font-xlarge text-center">
              {{ h.SI_number_span(member_count) }}
            </p>
          </div>
          <div class="col-md-6">
            <h4 class="text-center">{{ _('Datasets') }}</h4>
            <p class="font-xlarge text-center">
              {{ h.SI_number_span(organization.package_count) }}
            </p>
          </div>
        </div>
      {% endif %}
```

ckanext/canada/templates/public/organization/snippets/organization_search.html
```diff
@@ -0,0 +1,26 @@
...
+<div class="row mrgn-bttm-lg">
+  {% if count==0 %}
+    <div class="col-md-12 text-center text-warning">
+      <strong>{% snippet 'snippets/search_result_text.html', query=q, count=count, type="organization" %}</strong>
+    </div>
+  {% else %}
+    <div class="col-md-6 mrgn-tp-sm text-left text-success">
+        <strong>{% snippet 'snippets/search_result_text.html', query=q, count=count, type="organization" %}</strong>
+    </div>
+  {% endif %}
```
wardi, line 25: I think you can
```suggestion
  <div class="{{ 'col-md-12 text-center text-warning'
      if not count else 'col-md-6 mrgn-tp-sm text-left text-success' }}">
    <strong>{% snippet 'snippets/search_result_text.html', query=q, count=count, type="organization" %}</strong>
  </div>
```

ckanext/canada/templates/public/organization/snippets/organization_search.html
```diff
@@ -0,0 +1,26 @@
+<form class="form-inline mrgn-bttm-sm mrgn-tp-md" role="form" method="get" data-module="select-switch">
+  <div class="input-group col-md-12">
+    <input type="text" class="form-control" id="field-organization-search" autocomplete="off" name="q"
+           value="{{ q }}" placeholder="{{ _('Search organizations...') }}">
```
wardi, line 4: is the built-in org search going to do stemming properly to find french org names?

JVickery-TBS: In normal circumstances, no. BUT we are still using our `canada_org_title_translated_save` validator on the org title field.

That is the one which takes the fluent english and french title values and saves it as a piped string in the org's core title field.

The group list action with the `q` arg does `%{q}%` on the org name, title, and description. So yeah it will work with the French title. Shall I add a comment in the validator to never remove it as it is very important?

wardi: Instead of "never remove this" how about a list of the things that depend on it, like "org page search, recombinant csv output, ... (there may be others)"

ckanext/canada/templates/public/snippets/organization.html
```diff
@@ -25,22 +25,24 @@
...
-      <div class="col-md-6">
-        <h4 class="text-center">{{ _('Datasets') }}</h4>
-
-        <p class="font-xlarge text-center">
-          {{ h.SI_number_span(organization.package_count) }}
-        </p>
-      </div>
-    </div>
-  {% endif %}
+      {% set member_count = h.organization_member_count(organization.name) %}
```
wardi, line 28: This feels like a really inefficient way to get this data, and lots of work to remove orgs that fail the test. Can we have the list filtered and the number of members returned by the organization search action instead?

JVickery-TBS: @wardi this snippet is used in `read_base` of the organization templates. So it would only be rendered on the `organization_read` type stuff, not in any of the lists.

We did it this way because getting the users back from group/org show is trapped behind the `ckan.auth.public_user_details` config option

JVickery-TBS: @wardi HOWEVER https://github.com/ckan/ckan/pull/7007 there was this PR I did that was accepted into master which just allowed all of the group and org members to be public.

Backporting this would allow us to easily display the package counts and member counts on orgs

wardi: I'm fine with backporting your upstream change. It simplifies our templates now and will go away when we upgrade in the future

JVickery-TBS:
@wardi that PR for public group/org members is really big and has a lot of typing throughout it. And there is a bunch of it that relies on other upstream code that we do not have yet in out 2.9. So it is probably easier to just do this for now.

And we will just eventually get the public member stuff later on.

PR #1418, opened 2023-10-27
Resource list

RabiaSajjad:
- display resource list when creating new resource
- display language(s) and format within the resource list

### DIFF
--- a/ckanext/canada/templates/internal/package/new_resource_not_draft.html
+++ b/ckanext/canada/templates/internal/package/new_resource_not_draft.html
@@ -1,6 +1,5 @@
 {% ckan_extends %}
 
-{% block subtitle %}
+{% block secondary_content %}
+ {% snippet 'package/snippets/resources.html', pkg=pkg %}
 {% endblock %}
-
-{% block secondary_content %}{% endblock %}
--- a/ckanext/canada/templates/internal/package/snippets/resources.html
+++ b/ckanext/canada/templates/internal/package/snippets/resources.html
@@ -1,0 +1,64 @@
+{% set resources = pkg.resources or [] %}
+{% set can_edit = pkg and h.check_access('package_update', {'id':pkg.id }) %}
+{% if pkg and resources %}
+  {% block resources %}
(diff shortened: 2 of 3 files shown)


ckanext/canada/templates/internal/package/new_resource_not_draft.html
```diff
@@ -1,6 +1,5 @@
 {% ckan_extends %}
 
-{% block subtitle %}
+{% block secondary_content %}
+ {% snippet 'package/snippets/resources.html', pkg=pkg %}
```
wardi, line 4: I think this one-line addition is worth sending upstream too. https://github.com/ckan/ckan/blob/f8865626012770a2e67d4caf93cae7dbd1e4f094/ckan/templates/package/new_resource_not_draft.html#L20

ckanext/canada/templates/internal/package/snippets/resources.html
```diff
@@ -0,0 +1,66 @@
...
+{% set can_edit = pkg and h.check_access('package_update', {'id':pkg.id }) %}
+{% if pkg and resources %}
+  {% block resources %}
+    <section class="panel panel-default resources-side-section">
+      {% block resources_inner %}
+        {% block resources_title %}
+          <div class="panel-heading"><i class="fa fa-bars"></i> {{ _("Resources") }}</div>
+        {% endblock %}
+        <div class="panel-body resources-side-panel">
+          {% block resources_list %}
```
wardi, line 11: Can you `{% ckan_extends %}` and only override this block instead?

RabiaSajjad: the `resource`s block needs to be overwritten as we are displaying it using panel classes

wardi (approved):

ckanext/canada/templates/internal/package/new_resource_not_draft.html
wardi: done here https://github.com/ckan/ckan/pull/7889

PR #1419, opened 2023-11-01
Removal: Undelete Dataset Button & View

JVickery-TBS:
fix(templates): undelete dataset button link;

- Fixed link to undelete a dataset.
- Added CSS overrides for anchor button colors inside of BS alerts.

### DIFF
--- a/ckanext/canada/templates/internal/package/read.html
+++ b/ckanext/canada/templates/internal/package/read.html
@@ -66,7 +66,7 @@
 
     {% if pkg.state == "deleted" %}
       <div class="module-alert alert alert-danger mrgn-tp-sm">
-        <p>{{ _('This dataset has been deleted and is no longer accessible to non-admin users') }}.</p>
+        <p>{{ _('This dataset has been deleted and is no longer accessible') }}.</p>
       </div>
     {% endif %}
   </section>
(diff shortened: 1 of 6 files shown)


ckanext/canada/templates/internal/package/read.html
```diff
@@ -66,11 +66,7 @@
 
     {% if pkg.state == "deleted" %}
       <div class="module-alert alert alert-danger mrgn-tp-sm">
-        <p>{{ _('This dataset has been deleted and is no longer accessible to non-admin users') }}.
-          {% if h.check_access('package_update', {'id': pkg.id}) %}
-            <br><a href="{{ h.url_for('canada.package_undelete', pkg_id=pkg.id) }}" class="btn btn-warning">{{ _('Undelete') }}</a>
-          {% endif %}
-        </p>
+        <p>{{ _('This dataset has been deleted and is no longer accessible to non-admin users') }}.</p>
```
wardi, line 69: I know this was the original text, but since admin users on the portal site aren't a thing I think it's better to drop `to non-admin users` here

JVickery-TBS: @wardi copy has been updated and catalogs updated

wardi (approved):

PR #1420, opened 2023-11-02
remove _site_messaging from source control

RabiaSajjad:


wardi (approved):
looks good assuming all servers have been updated

PR #1423, opened 2023-11-08
add environment variable for temp directory

RabiaSajjad:


bin/pd/Makefile
```diff
@@ -33,6 +33,7 @@ ckanapi := $(REGISTRY_CKANAPI_COMMAND)
 ogc_search := $(OGC_SEARCH_COMMAND)
 oc_search := $(OC_SEARCH_COMMAND)
 backup_dir := $(PD_BACKUP_DIRECTORY)
+tmp_dir := $(PD_TMP_DIRECTORY)
```
wardi, line 36: I think we can default this one to $TMPDIR, the normal system default, so we don't have to update our dev environments with a setting only needed in prod

wardi, line 36:
```suggestion
tmp_dir := $(or $PD_TMP_DIRECTORY,$TMPDIR)
```

wardi: Hmm. this doesn't work exactly because $TMPDIR can be unset then we'll get an error on mktemp

wardi: Updated suggestion, this one should work in all cases and not require changes in our dev environments

bin/pd/Makefile
```diff
@@ -84,6 +85,8 @@ else ifeq ($(oc_search),)
 $(error OC_SEARCH_COMMAND is undefined)
 else ifeq ($(backup_dir),)
 $(error PD_BACKUP_DIRECTORY is undefined)
+else ifeq ($(tmp_dir),)
+$(error PD_TMP_DIRECTORY is undefined)
```
wardi, line 89:
```suggestion
```
Then this error can't happen

PR #1424, opened 2023-11-09
Updated JS and Flash Messages for GA4 Custom Events

JVickery-TBS:
- moved ga scripts to webassets
- updated js scripts to target the flash messages for new CKAN templating
- updated js for GA4 custom events
- added new Flash messages for clearer visuals to user and to be able to do js regex matches better for certain actions (resource vs dataset added and updated)


JVickery-TBS:
Okay the french translations are in now for the new flash messages, and the final js conditions have been added for those strings

wardi:
I thought calling "datasets" "records" was deliberate, since we use the same messages for open information records too, right?

JVickery-TBS:
@wardi yes I think that was the general idea for the Open Information records. Those are the ones added and searchable at `/info/new` and `/info` right?

If so, the page titles and some sub titles, and tab links all say "Dataset" so that might be a bigger issue. I can go through and condition the dataset type on info to show "Record" instead of "Dataset"

That would allow us more GA events for specific dataset types.

JVickery-TBS:
@wardi okay so from the meeting this morning, I guess we are just gunna stick with "Dataset" semantics for all the types. How does the rest of this PR look?

ckanext/canada/assets/internal/canada_internal_analytics.js
```diff
@@ -0,0 +1,319 @@
...
+            'title': title,
+            'referrer': referrer,
+            'messages': messages,
+          });
+
+        }
+
+        // track password reset request
+        if( messages.match(/.*check your e-mail inbox in order to confirm your password reset.*/) ||
+            messages.match(/.*consulter votre boîte à lettres électronique pour confirmer votre demande.*/)
```
wardi, line 62: these messages sound like they're referring to different things

wardi: also this pattern of matching text seems really fragile. Can we add some data attributes or something else that won't change over time?

JVickery-TBS: For the first comment: I just copied from our PO files for the French translations.

Second comment: Yeah, I also did not like how this was done very much. The front-end analytics stuff always gets tricky with back-end frameworks. But as the flash messages are session based, it makes sense to do it on the flash message I guess. As for the data attributes or classes, that would be wildly preferable. It is just that we would probably have to override a bunch of views to get what we want in the flash messages?

Unless if we can think of some super cool fancy way to catch flash messages and see what action called them? and then insert something there. E.g. the `package_create` action did `h.flash_success` and we could see which view method called `package_create` or something. And then just map CKAN actions to GA Custom Events

wardi: we could override the `flash_success` and related helpers to capture information about the caller

JVickery-TBS: Yes, that would be the CORRECT way to do it. But what about throwing pythonian out the window and directly Monkey Patching the `_Flash` class from the core helpers :)

wardi: That's an option if overriding the helpers seems like it would difficult to maintain, but likely better done in our ckan fork instead of monkey patching.

JVickery-TBS:
@wardi okay I have mapped view class and methods to the custom GA4 html attributes which get used in the JS.

This works real nice, but required this PR: https://github.com/open-data/ckan/pull/160 (there is also an upstream PR but this one for us has some specific 2.9 code that we need too).

Let me know if there is a better way to map the callers here.

wardi:
Sounds like a great idea, leave the configuration all on the GA side and automatically handle flash messages from other views without having to update anything in our extension

PR #1425, opened 2023-11-15
ATI Analytics Filter Out Current Month

JVickery-TBS:
feat(dev): ati analytics filtering;

- Filter out the current month's data from the ATI analytics CSV.

### DIFF
--- a/bin/update_ati_analytics.sh
+++ b/bin/update_ati_analytics.sh
@@ -43,6 +43,11 @@ if [[ ! -d "/opt/tbs/tmp" ]]; then
 
 fi
 
+###
+# Get the file path generated from Drupal
+# The SMB directories are different on some servers
+# hence the checks for the different file paths
+###
 generated_file='/opt/tbs/ckan/smb_portal/portal/public/ati-informal-requests-analytics.csv';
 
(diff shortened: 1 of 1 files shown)


bin/update_ati_analytics.sh
```diff
@@ -78,7 +88,12 @@ if [[ ! -d "${tmp_dir}" ]]; then
 
 fi
 
-cp ${generated_file} ${tmp_file};
+###
+# We do not want the current month's counts
+###
+filter_date="$(date +%Y),$(echo $(date +%m) | sed 's/^0*//'),";
+echo "Filtering out rows: ${filter_date}";
+sed "/^${filter_date}*/d" < ${generated_file} > ${tmp_file};
```
wardi, line 96: sed isn't generally safe for use with csv files. Are we sure there aren't any newlines in the data?

wardi: Ok, then maybe a `# should be safe, I think there aren't any newlines` comment would be fine

JVickery-TBS: Comment added

wardi (approved):

PR #1426, opened 2023-11-17
ATI Analytics Shell Script to Python

JVickery-TBS:
feat(bin): moved at analytics from sh to py;

- Remade the ati analytics script from shell to python.


bin/update_ati_analytics.py
```diff
@@ -0,0 +1,125 @@
...
+from datetime import datetime
+import subprocess
+from six import PY2
+from shutil import rmtree
+from codecs import BOM_UTF8
+
+if PY2:
+    from cStringIO import StringIO
+else:
+    from io import StringIO
```
wardi, line 21: this is a standalone script, why bother with python2 support?

bin/update_ati_analytics.py
```diff
@@ -0,0 +1,125 @@
+"""
```
wardi, line 1:
```suggestion
#!/usr/bin/env python3
"""
```
And a `chmod +x` would let us run this script directly

bin/update_ati_analytics.py
```diff
@@ -0,0 +1,125 @@
...
+        temp_dir = tempfile.mkdtemp()
+    else:
+        temp_dir = tempfile.TemporaryDirectory()
+    temp_file = os.path.join(temp_dir, 'ati-informal-requests-analytics.csv')
+
+    today = datetime.today()
+
+    errors = StringIO()
+    skipped_rows = 0
+    command = 'ckanapi action resource_patch -c %s id=%s upload@"%s"' \
```
wardi, line 72: better to use a list so that subprocess.call can use execv and shell=False to skip shell parsing of the string. These parameters may not cause a problem but it's a good habit regardless.

bin/update_ati_analytics.py
JVickery-TBS: Yeah, I guess we would just need to do `pip3 install click` if we do not already have that dependency on the Registry servers for python 3

Or if that GA venv is still on the servers and is using py3, then that could be a fine venv to use

bin/update_ati_analytics.py
JVickery-TBS: Okay using a list now without `shell=True`, it is just failing on the `ckanapi` command usage so will need to look into that

bin/update_ati_analytics.py
JVickery-TBS: executable mode added to the file and py3 bin header

bin/update_ati_analytics.py
JDvickery: @wardi okay that should be fixed now. It was because I had the extra `ckanapi` in the command and I guess that is only needed for the python2 subprocess and not the python3 one.

PR #1429, opened 2023-11-30
Prevent Core Dataset and Resource Views for PD Types

JVickery-TBS:
Oops forgot to put a description here...

So users can actually access the Views to all of our PD types created from Recombinant. Which is fine in a normal usage of Recombinant? But figured we do not want users being able to edit the PD datasets and resources, or even adding new ones vie the UI.

This will just redirect all of the Core views if the package_type is a recombinant type.

Not too sure how we want to handle this limitation in the API? If we can do some simple extension of the API view, and prevent PD types from getting updated like that?

### DIFF
--- a/ckanext/canada/view.py
+++ b/ckanext/canada/view.py
@@ -845,7 +845,7 @@ def action(logic_function, ver=API_DEFAULT_VERSION):
 
 
 def _get_package_from_api_request(logic_function, id, context):
-    # type: (str, str, dict) -> str|None
+    # type: (str, str, dict) -> dict|None
     """
     Tries to return the package for an API request
     """
@@ -877,8 +877,6 @@ def _get_package_from_api_request(logic_function, id, context):
 
 
 def _log_api_access(request_data, pkg_dict):
-    # Core code has stopped passing objects inside of the context.
-    # So we cannot rely on key/values existing inside the context.
     org = model.Group.get(pkg_dict.get('owner_org'))
     g.log_extra = u'org={o} type={t} id={i}'.format(
         o=org.name,
--- a/ckanext/canada/view.py
+++ b/ckanext/canada/view.py
@@ -113,27 +113,31 @@ def logged_in():
         return h.redirect_to('user.login')
 
 
+def _get_package_type_from_dict(package_id, package_type='dataset'):
+    try:
+        context = {
+            u'model': model,
+            u'session': model.Session,
+            u'user': g.user,
+            u'auth_user_obj': g.userobj
+        }
+        pkg_dict = get_action(u'package_show')(
+            dict(context, for_view=True), {
+                u'id': package_id
+            }
+        )
+        return pkg_dict['type']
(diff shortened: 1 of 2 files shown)


JVickery-TBS:
@wardi are we only overriding the Core API view in our extension just to add in logging?

JVickery-TBS:
@wardi I massaged the override for the API view a bunch. It now requires this PR: https://github.com/open-data/ckan/pull/161

So I added the no version route, and the `3` version route here. And just keeping in the `1` and `2` in case any orgs have systems that are using those still.

I had to get the data before the action actually happens to be able to check for Recombinant/PD types to `403` them to prevent any POSTing to those endpoints for pd types. And because of that, we no longer have the `package` in the request context at the time. I could get that context from one of the new methods I think, but figured it would be fine to just access the package dict as I have that now at that point. And as other things can modify the context object, figured it might be better to just pass the package dict in there.

ckanext/canada/view.py
```diff
@@ -813,125 +803,79 @@ def _org_key(org):
...
+    context, request_data, return_dict = api_view_action(logic_function, ver=-1)
+
+    # extra logging here
+    id = request_data.get('id', request_data.get('package_id', request_data.get('resource_id')))
+    pkg_dict = _get_package_from_api_request(logic_function, id, context)
+    if pkg_dict:
+        _log_api_access(request_data, pkg_dict)
+
+    # prevent PD types from being POSTed to via the API
+    if request.method == 'POST':
```
wardi, line 838: We want users to be able to upsert/delete PD records with the API, won't this change prevent that?

JVickery-TBS:
@wardi okay I have added `before_request` methods to the dataset and resource core views. This is because you can still access all of the views for PD types through `/dataset/<pkg id>/resource/<res id>` or `/info/<pkg id>/resource/<res id>`.

So this just does a pre-check on the package type, and redirects if the package type is a Recombinant one. Let me know if you think we should just 404 instead of redirect.

wardi:
redirect sounds good

ckanext/canada/view.py
JVickery-TBS: From testing it, it seems that you are able to still use all of the `DataStore` endpoints. Which is what the users will use to manage the PD data. I am not sure why this is even possible, because yeah, from the look of this new code it SHOULD prevent that

JVickery-TBS: it must not be able to find the package or something. Will be debugging and seeing. I can just make sure that the actions are not a `datastore` one too

JVickery-TBS: @wardi yeah when getting the pkg_dict, I had `if logic_function.startswith('resource')` and then did `resource_show` in that block. But of course the logic actions start with `datastore`. So conditions have been updated.

resource and package action endpoints will `403` for PD types. But datastore action endpoints are allowed.

wardi:
this looks good

PR #1430, opened 2023-12-05
Xloader Status Badge & Conditional DataTable Views

JVickery-TBS:
feat(dev): xloader status badge, conditional datatable views;

- Added status badge for Xloader.
- Added conditions to the datatables views to not display if the Resource is datastore inactive or invalid.

@wardi this is just a proof of concept, it adds a badge for the Xloader status / DataStore status. And it adds some conditions to actually being able to see DataTable views. I made it so that logged in users will still be able to see and edit there tables if possible, but there will be a visual that the View is "disabled".


wardi:
Interesting! Maybe you can demo the feature for us?

JVickery-TBS:
@wardi I think I have everything done in here now.

- Xloader status badge;
- Dynamic display of DataTable views if Xloader failure or Validation failure;
- Job Queue view;
- Redirect Group Bulk Actions blueprint route;
- Disable Group Bulk Actions logic action methods (these core action methods make changes directly to DB so you could actually change datasets to Public without `before_update` happening on them and our plugin forcing Private; I have made an upstream Issue regarding this);
- Hide the Validation Options section in Resource form (we checked the DB and literally nobody uses the options or the schema, but who cares now cuz T A B L E   D E S I G N E R);
- Validation method override for the validation options, for API users mainly just in case (has new pytest for it as well).

Major thing to note is that we will just need to change the loading priority of the `validation` plugin in INI files for the validation method overrides.

So if you have any extra time to review :)

And we can generate the new strings after and get them sent to translation once we are 100% on the code here

ckanext/canada/helpers.py
```diff
@@ -641,3 +646,84 @@ def flash_success(message, allow_html=True):
...
+                break
+            queue_number += 1
+
+    title = t.h.render_datetime(xloader_job.get('last_updated'), with_hours=True) \
+        if xloader_job.get('last_updated') else ''
+    if queue_number:
+        title = _("Number %s in queue") % queue_number
+
+    # TODO: use str instead of unicode for py3
+    return unicode('<a href="{pusher_url}" class="loader-badge"><img src="{badge_url}" alt="{alt}" title="{title}"/></a>').format(
```
wardi, line 725:
```suggestion
    return u'<a href="{pusher_url}" class="loader-badge"><img src="{badge_url}" alt="{alt}" title="{title}"/></a>'.format(
```
should work in both

ckanext/canada/helpers.py
```diff
@@ -641,3 +646,84 @@ def flash_success(message, allow_html=True):
...
+        pusher_url=pusher_url,
+        badge_url=badge_url,
+        alt=messages[status],
+        title=title)
```
wardi, line 729: can alt or title ever include `"`? (if so there's html injection issues here)

ckanext/canada/logic.py
```diff
@@ -185,3 +194,88 @@ def datastore_create_temp_user_table(context):
...
+    We need this method to 404 the resource_view page (e.g. viewing in full screen)
+
+    If a user is logged in, we want them to be able to see and edit the datatables_view
+    views still. We will just add a key to the view dict to be used within templates for visuals.
+    """
+    view_dict = up_func(context, data_dict)
+    if view_dict.get('view_type') == 'datatables_view':
+        # at this point, the core function has been called, calling resource_show
+        # and adding the Resource Object into the context
+        res_dict = context.get('resource').__dict__
```
wardi, line 220: this will not a normal way to pass resource data, either use the resource object or the dictized version from resource_show/package_show not the resource object's `__dict__`

ckanext/canada/logic.py
```diff
@@ -185,3 +194,88 @@ def datastore_create_temp_user_table(context):
...
+        res_extras = res_dict.get('extras', {})
```
wardi, line 221:
```suggestion
        res_extras = getattr(context.get('resource'), 'extras', {})
```

ckanext/canada/logic.py
```diff
@@ -185,3 +194,88 @@ def datastore_create_temp_user_table(context):
...
+    This will delete any datatables_view from the list of views.
+
+    If a user is logged in, we want them to be able to see and edit the datatables_view
+    views still. We will just add a key to the view dict to be used within templates for visuals.
+    """
+    view_list = up_func(context, data_dict)
+    # at this point, the core function has been called, calling resource_show
+    # and adding the Resource Object into the context
+    res_dict = context.get('resource').__dict__
+    res_extras = res_dict.get('extras', {})
```
wardi, line 262: same here, there's rarely a good reason to access `__dict__`

ckanext/canada/logic.py
```diff
@@ -185,3 +194,88 @@ def datastore_create_temp_user_table(context):
...
+
+        if view_dict.get('view_type') == 'datatables_view' and \
+        ( not res_extras.get('datastore_active', False) or
+          res_extras.get('validation_status') != 'success'):
+            # if the Resource is not active in the DataStore or has not
+            # passed validation, we do not want to show its datatables_view views
+
+            if not g.user:
+                # only remove from the list if a user is not logged in
+                del view_list[i]
```
wardi, line 274: it's unsafe to modify a list while iterating over it

ckanext/canada/plugins.py
```diff
@@ -88,6 +88,7 @@ class CanadaDatasetsPlugin(SchemingDatasetsPlugin):
     p.implements(p.IDatasetForm, inherit=True)
     p.implements(p.IPackageController, inherit=True)
     p.implements(p.IBlueprint)
+    p.implements(p.IActions)
```
wardi, line 91: all the other plugins in this file implement IActions, maybe add it to one of those instead?

ckanext/canada/templates/internal/scheming/package/snippets/resource_form.html
```diff
@@ -57,7 +57,7 @@
     {% endif %}
   </div>
 
-  <div class="mrgn-bttm-lg">
+  <div class="mrgn-bttm-lg hidden">
```
wardi, line 60: instead of hiding the form do we prefer to have the same set of rules for all resources? We could do that by removing this from the form and having an `output_validator` in our schema that always returns something like `{}`

Or we could remove the field completely and make ckanext-validation not require a value from the user

ckanext/canada/templates/public/admin/jobs.html
```diff
@@ -0,0 +1,60 @@
...
+        <h3 class="panel-title">
+            <i class="fa fa-info-circle"></i>
+            {{ _('Job Queue') }}
+        </h3>
+    </header>
+    <div class="panel-body">
+      <p>{{ _("View the current job queue and running jobs.") }}</p>
+    </div>
+  </section>
+{% endblock %}
```
wardi, line 60: an admin page for listing jobs would be a great upstream feature

ckanext/canada/templates/public/datatables/datatables_view.html
```diff
@@ -1,6 +1,20 @@
 {% ckan_extends %}
 
+{% block bodytag %}
+  {{- super() -}}
+  {% if resource_view.get('canada_disabled_view') %}
+    id="canada-disabled-view-body"
```
wardi, line 6: should this be a class? seems odd to use an id here

ckanext/canada/view.py
```diff
@@ -579,6 +596,79 @@ def ckanadmin_publish_datasets():
...
+    except NotAuthorized:
+        abort(403, _('Not authorized to see this page'))
+
+    warning = False
+    if jobs and datetime.strptime(jobs[0]['created'], '%Y-%m-%dT%H:%M:%S') < (datetime.now() - timedelta(minutes=18)):
+        warning = True
+
+    job_list = []
+    mystery_job_count = 0
+    for job in jobs[:25]:
```
wardi, line 615: give this constant al ALLCAPS_NAME and put it at the module level, or better yet make it a config option

ckanext/canada/view.py
```diff
@@ -579,6 +596,79 @@ def ckanadmin_publish_datasets():
...
+        try:
+            job_obj = Job.fetch(job.get('id'))
+            job_args = job_obj.args[0]
+        except (NoSuchJobError, KeyError):
+            mystery_job_count += 1
+            continue
+
+        if '.run_validation_job' in job_obj.func_name:
```
wardi, line 623: this is kind of unpleasant. maybe if it's a core feature we could add labels to the jobs and use a helper to provide human-readable descriptions so plugins can add their own human-readable versions

ckanext/canada/view.py
JVickery-TBS: which const? `mystery_job_count`? I decided not to use that in the output, so we can probably just remove it from the code here

ckanext/canada/view.py
JVickery-TBS: Redis job queue does support metadata titles, either CKAN does not use them, or Xloader and Validation is not using them. So we can just get the job object from `rq` and print out the title.

And then I guess we can just add those translations into our `strings.py`

ckanext/canada/templates/public/datatables/datatables_view.html
JVickery-TBS: I believe I had to do this because of some plugin priority thing in which the class attribute gets overridden. I can just change this to a `data` attribute though, like `data-view-disabled`

ckanext/canada/templates/internal/scheming/package/snippets/resource_form.html
JVickery-TBS: Yeah, this was the easiest/quickest way to do it. As our schema uses the preset from Validation. We could copy the schema from validation and just change the form output value?

Or are we able to specific preset AND a form output snippet as well?

ckanext/canada/logic.py
JVickery-TBS: save index numbers to a new list, and then bisect after the loop is done??

ckanext/canada/logic.py
JVickery-TBS: yup, not sure why I did that. I can just access `context.get('resource').extras` right??

JVickery-TBS: oops sorry did not see ur above suggestion. sweeeeet

ckanext/canada/helpers.py
JVickery-TBS: No, that would surely break the html structure. Hmmm for the `alt`, it should be fine because those are static UNLESS gettext returns something else...so yeah we will need to html encode those for sure.

Is there a CKAN way to do that? or just using python libs:
```
try:
    from html import escape  # python 3.x
except ImportError:
    from cgi import escape  # python 2.x
```

ckanext/canada/view.py
wardi: I mean the `25` cut-off

ckanext/canada/templates/public/admin/jobs.html
JVickery-TBS: will make a ticket in the backlog for it

ckanext/canada/helpers.py
JVickery-TBS: sweet thanks!

ckanext/canada/logic.py
JVickery-TBS: done

ckanext/canada/logic.py
JVickery-TBS: okay, just saved index numbers and then removed after loop

ckanext/canada/helpers.py
JVickery-TBS: escaping html with webhelpers that is already in use in this script. And escaping double quotes as well.

ckanext/canada/plugins.py
JVickery-TBS: moved to the Public plugin class

ckanext/canada/templates/internal/scheming/package/snippets/resource_form.html
JVickery-TBS: Okay, so we need to have the `validation_options` in the form for sure, as there is the whole thing with the default validation options to set the max row limit.

So I have now just made some presets in our `presets.yaml` and used the hidden snippet. I have also changed it so we are not overriding the `validation_options_validator` and `resource_schema_validator` methods. So the validation plugin loading priority will not matter now.

@wardi  lemme know if you think it would be better to just remove the fields fully, and we can just hard code the row limit into our validation fork??

ckanext/canada/templates/public/datatables/datatables_view.html
JVickery-TBS: data attribute works

ckanext/canada/view.py
JVickery-TBS: In a constant now, will make an option in the upstream contrib. But for now, constant is fine

ckanext/canada/templates/internal/scheming/package/snippets/resource_form.html
JVickery-TBS: @wardi nevermind, these fields are annoying me greatly right now, so I am going to remove them and just force the row limit json into the `validation_options` field value.

JVickery-TBS: @wardi okay...not sure if that is the best way to "remove" a field or not. But as the validation extension is still expecting these fields, I have just made harsh validator methods which always forcefully set the field values to what we want.

I am very bad with these schema yaml files, so lemme know. Thanks!

ckanext/canada/view.py
JVickery-TBS: So yeah, you can just pass a `title` attribute to the enqueue toolkit method and it will add the meta title.

Hmmm right. Well we will still need to do some nasty conditions because the job data is not the same for everything, and we want the Resource information.

@wardi let me know if you have any ideas on how to make this not ugly and a bit better

But the new code here would required these two PRs:
- https://github.com/open-data/ckanext-xloader/pull/17
- https://github.com/open-data/ckanext-validation/pull/10

JVickery-TBS: @wardi semantics are in, but as we still need to have the if/else statements to be able to get the Resource info, it might just be better to condition on the callback method?

ckanext/canada/logic.py
```diff
@@ -185,3 +194,89 @@ def datastore_create_temp_user_table(context):
...
+                disabled_views_indexes.append(i)
+                continue
+
+            if g.user != site_user:
+                # only add key/value if not system process
+                view_list[i]['canada_disabled_view'] = True
+    for i in disabled_views_indexes:
+        # actually delete the views from the list
+        del view_list[i]
+    return view_list
```
wardi, line 282: There's a bug here when multiple views are disabled (later indexes are wrong, could cause an IndexError or removing the wrong views). Here's a better way to do this:
```suggestion
    return [v for i, v in enumerate(view_list) if i not in disabled_views_indexes]
```
Or you could build up a list in the loop above instead of accumulating indexes to remove

ckanext/canada/schemas/presets.yaml
```diff
@@ -4075,3 +4075,25 @@ presets:
...
+
+# Field = Schema.
+- preset_name: canada_resource_validation_schema
+  values:
+    field_name: schema
+    form_snippet: null
+    form_panel: _exclude_from_form
+    required: false
+    display_snippet: null
+    validators: canada_resource_schema_validator
```
wardi, line 4087: we could use `validators: ignore` and a custom `output_validators` that returns the value we want instead. That way we don't need to update every resource in the db, all we need to do is reindex the datasets (the fixed value we choose will always be output by the output_validators)

Or we could hack the validation extension to only use our options and not require any special resource fields..

ckanext/canada/templates/internal/scheming/package/snippets/resource_form.html
wardi: please do the __extras thing as a separate PR just in case it causes problems with something

ckanext/canada/schemas/presets.yaml
JVickery-TBS: @wardi it seems that using `validators: ignore` and an output validator does not work? Does there need to be a value for the output validator to fire off??

JVickery-TBS: @wardi I have added in a `canada_blank_value_creator` which just sets the validation options and schema fields to an empty value so that the output validators work.

I just do not really wish to modify the validation fork/extension more than we have to.

ckanext/canada/tests/test_validators.py
```diff
@@ -401,17 +401,70 @@ def test_validation_schema(self):
...
+                'sheet': 2,
+                'fill_merged_cells': True,
+                'workbook_cache': 'lambda',
+                'row_limit': 'STR to 1 MIL'
+            }
+        }
+
+        resource = self.sysadmin_action.resource_create(**resource_data)
+        assert resource['validation_options'] == {'row_limit': 1000000,
+                                                  'dialect': {
```
wardi, line 451: I don't think it makes sense to export this into the metadata of every resource for every dataset on our site considering it's a fixed value. Use a configuration option instead?

I don't see the utf-8 encoding here, did we force that in another part of the code?

JVickery-TBS: @wardi Okay sounds good, I suppose we can remove the stuff from here really. I will just always return blank values for these fields.

And then in `ckanext-validation` we have our our fork/branch so will just hardcode 1M rows in there with a config option?

And then as Validation and Xloader both use the dialects, I will make a config option like `ckanext.canada.ds_resource_dialects` and have that be used in Xloader and Validation for static dialects.

As for the encoding, we do not have static encoding, goodtables/xloader/validation will try to guess encoding and decode it. So I suppose I will do another config option like `ckanext.canada.ds_resource_encoding` and use that value in Xloader and Validation.

This way we will get our goodtables/frictionless to use the `utf-8` encoding and static, basic dialects for TSV and CSV.

wardi: That sounds good, and instead of `canada` this and that it could be a proper ckanext-validation feature that lets you set static dialects and encoding instead of detecting them. I imagine there are others that would be interested in standardizing their csv files.

JVickery-TBS: @wardi okay so we now have these PRs:
- https://github.com/open-data/ckanext-validation/pull/11
- https://github.com/open-data/ckanext-xloader/pull/20

Which will use the new `ckanext.validation.static_validation_options` config. That will handle the static encoding and static dialects that goodtables and tabulator expects.

And then we just parse the same stuff in the Xloader fork to use the same static encoding and dialects.

ckanext/canada/helpers.py
```diff
@@ -719,6 +728,86 @@ def flash_success(message, allow_html=True):
...
+            if job.get('id') == xloader_job.get('job_id'):
+                break
+            queue_number += 1
+
+    title = t.h.render_datetime(xloader_job.get('last_updated'), with_hours=True) \
+        if xloader_job.get('last_updated') else ''
+    if queue_number:
+        title = _("Number %s in queue") % queue_number
+
+    return u'<a href="{pusher_url}" class="loader-badge"><img src="{badge_url}" alt="{alt}" title="{title}"/></a>'.format(
```
wardi, line 804: this should be returned as Markdown, not `safe`d on the template side

ckanext/canada/logic.py
```diff
@@ -185,3 +195,105 @@ def datastore_create_temp_user_table(context):
...
+    We need this method to 404 the resource_view page (e.g. viewing in full screen)
+
+    If a user is logged in, we want them to be able to see and edit the datatables_view
+    views still. We will just add a key to the view dict to be used within templates for visuals.
+    """
+    view_dict = up_func(context, data_dict)
+    if view_dict.get('view_type') == 'datatables_view':
+        # at this point, the core function has been called, calling resource_show
+        # and adding the Resource Object into the context
+        res_extras = getattr(context.get('resource'), 'extras', {})
```
wardi, line 222: I believe this changes upstream, we've stopped caching resource etc. in the context dict because it's unsafe when contexts are passed around and just as fast to rely on the sqlalchemy cache

ckanext/canada/schemas/presets.yaml
```diff
@@ -4075,3 +4075,25 @@ presets:
...
+
+# Field = Validation Options.
+- preset_name: canada_resource_validation_options
+  values:
+    field_name: validation_options
+    form_snippet: null
+    form_panel: _exclude_from_form
+    required: false
+    display_snippet: null
+    validators: canada_blank_value_creator
```
wardi, line 4098: We need a blank value stored and can't use `ignore` because of ckanext-validation?

JVickery-TBS: Okay I got `ignore` working as the validator. I am just going to keep the `None` value output validator for existing Resources that may have values in these fields.

ckanext/canada/helpers.py
```diff
@@ -719,6 +728,86 @@ def flash_success(message, allow_html=True):
...
+    # FIXME: is this worth it if we cannot display in the Xloader page?
+    # is it worth it if we cannot display for the Validation extension?
+    queue_number = None
+    if status == 'pending':
+        queue_number = 1
+        jobs_in_queue = t.get_action('job_list')({"ignore_auth": True}, {})
+        for job in jobs_in_queue:
+            if job.get('id') == xloader_job.get('job_id'):
+                break
+            queue_number += 1
```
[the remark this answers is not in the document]
JVickery-TBS: Ah yes...well funny thing about that. I came across that issue when our xloader queue got backed up to 4k jobs. The `job_list` action method returns a dictization of every single job in the queue. Which I suppose might be an issue/improvement in Core to make.

So I tested it by loading up 800 jobs in the queue locally and it really broke the Job Queue admin page I was working on. So that is why in the Job Queue view, I do `MAX_JOB_QUEUE_LIST` of 25 to output on the page.

In the RQ class from redis lib, there is a `job_ids` class property, CKAN Core just uses the `jobs` prop and dictizes each job object. But another action method in ckan core to just return the job_ids list could be useful too.

But for now, and for this, I did put in the `FIXME` to discuss if getting the number in queue was even worth it. Especially as we cannot know a Validation job's id without being inside of the loop of job objects. So it does not really seem worth it, especially in the badges which show up on critical UX pages.

So removing the queue count I think for now.

ckanext/canada/view.py
```diff
@@ -579,6 +625,64 @@ def ckanadmin_publish_datasets():
...
+        jobs = get_action('job_list')({}, {})
+    except NotAuthorized:
+        abort(403, _('Not authorized to see this page'))
+
+    warning = False
+    if jobs and datetime.strptime(jobs[0]['created'], '%Y-%m-%dT%H:%M:%S') < (datetime.now() - timedelta(minutes=18)):
+        warning = True
+
+    job_list = []
+    for job in jobs[:MAX_JOB_QUEUE_LIST_SIZE]:
```
wardi, line 643: MAX_JOB_QUEUE_LIST_SIZE should be a job_list `limit` parameter instead to save wasted work getting jobs that we discard.

ckanext/canada/view.py
```diff
@@ -579,6 +625,64 @@ def ckanadmin_publish_datasets():
...
+        try:
+            job_obj = Job.fetch(job.get('id'))
```
wardi, line 645: Can `job_list` be extended to return the information you need instead of querying the back end from a view function?

ckanext/canada/logic.py
JVickery-TBS: Yeahhhh all these object in the context passed between methods have always confused me anyways. Okay well, same thing, I guess at that point we can assume the Resource exists so I will just use `model.Resource.get`

ckanext/canada/view.py
JVickery-TBS: @wardi here is a PR to our fork/branch to add support for offset & limit to the `job_list` action method. https://github.com/open-data/ckan/pull/165

I noted in the PR about an issue if we want to submit that upstream.

ckanext/canada/view.py
JVickery-TBS: @wardi all of this logic from the view method has now been moved into a chained logic action for `job_list`.

ckanext/canada/helpers.py
```diff
@@ -719,6 +728,70 @@ def flash_success(message, allow_html=True):
...
+    badge_url = t.h.url_for_static('/static/img/badges/{lang}/datastore-{status}.svg'.format(lang=t.h.lang(), status=status))
+
+    title = t.h.render_datetime(xloader_job.get('last_updated'), with_hours=True) \
+        if xloader_job.get('last_updated') else ''
+
+    return Markup(u'<a href="{pusher_url}" class="loader-badge"><img src="{badge_url}" alt="{alt}" title="{title}"/></a>'.format(
+        pusher_url=pusher_url,
+        badge_url=badge_url,
+        alt=html.escape(messages[status].replace('"', '\"')),
+        title=html.escape(title.replace('"', '\"'))))
```
wardi, line 792: doesn't https://docs.python.org/3/library/html.html#html.escape already handle quotes? Maybe only needs `quote=True` for python 2 compatibility

ckanext/canada/logic.py
```diff
@@ -185,3 +226,176 @@ def datastore_create_temp_user_table(context):
...
+                continue
+            job_info = {'name_translated': resource.get('name_translated'),
+                        'resource_id': rid,
+                        'url': h.url_for('dataset_resource.read',
+                                         id=resource.get('package_id'),
+                                         resource_id=rid)}
+
+        job['created_human_readable'] = canada_date_str_to_datetime(job.get('created')) \
+                .replace(tzinfo=utc).astimezone(ottawa_tz) \
+                .strftime('%Y-%m-%d %H:%M:%S %Z')
```
wardi, line 396: All the other APIs return UTC dates so this is unusual. Also why is it called human_readable when it looks almost like ISO8601?

wardi (commented):
just a couple minor things otherwise looks good

ckanext/canada/logic.py
JVickery-TBS: Okay, I have left `created` as the normal UTC in the logic method, and just changed the output in the view/template.

ckanext/canada/templates/public/admin/jobs.html
```diff
@@ -0,0 +1,56 @@
...
+    </div>
+    <ul class="canada-job-list activity">
+      {% for job in job_list %}
+        <li class="item">
+          <i class="fa icon {{ job.get('icon') }}"></i>
+          <p>
+            <span class="canada-job-item-title">{{ job.get('type') }}</span>
+            <br/>
+            <span class="date">
+              {{ job.get('since_time') }} ({{ job.get('created_human_readable') }}) |
```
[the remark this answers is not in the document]
JVickery-TBS: Yeah, I don't think we care that much for the EST time haha. Especially if we are giving a time ago as well. Okay I am only doing this in the template now then.

JVickery-TBS:
@wardi okay fixed those things up now. And I also fixed up the User Gravatar stuff as well, and removed some old, unused code/scripts.

wardi (approved):

PR #1432, opened 2023-12-18
Qp nil clean

PatLittle:
QPnotes nil report

### DIFF
--- a/bin/pd/Makefile
+++ b/bin/pd/Makefile
@@ -6,7 +6,7 @@ targets := ati briefingt qpnotes consultations contracts contractsa dac experime
 csv_files := \
 	ati.csv ati-nil.csv \
 	briefingt.csv \
-	qpnotes.csv \
+	qpnotes.csv qpnotes-nil.csv \
         contractsa.csv \
         contracts.csv contracts-nil.csv \
 	consultations.csv \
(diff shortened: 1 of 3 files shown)


bin/pd/Makefile
```diff
@@ -206,23 +206,34 @@ $(workdir)/filtered/briefingt.csv: $(workdir)/briefingt.csv
...
 .PHONY: upload-qpnotes
-upload-qpnotes: $(workdir)/filtered/qpnotes.csv
+upload-qpnotes: $(workdir)/filtered/qpnotes.csv $(workdir)/filtered/qpnotes-nil.csv
 	@$(echo_date) Uploading Question Period Notes...
 	@$(ckanapi) action resource_patch -c $(registry_ini) \
 	id=c55a2862-7ec4-462c-a844-22acab664812 upload@"$(workdir)/filtered/qpnotes.csv" \
 	> /dev/null
+	# @$(ckanapi) action resource_patch -c $(registry_ini) \
+	# id=TBD upload@"$(workdir)/filtered/qpnotes-nil.csv" \
+	> /dev/null
```
wardi, line 216:
```suggestion
	# > /dev/null
```

ckanext/canada/tables/qpnotes.yaml
```diff
@@ -357,3 +357,123 @@ excel_column_heading_style:
   PatternFill:
     patternType: solid
     fgColor: FFEFEFEF
+
+- title: Proactive Publication - Question Period Notes Nothing to Report
```
wardi, line 361: We need a corresponding entry in the .po for this

wardi:
otherwise looks good

ckanext/canada/tables/qpnotes.yaml
PatLittle: I think I got that figured out now

wardi (approved):

PR #1433, opened 2023-12-20
Gettext for Quicklinks Template

JVickery-TBS:
fix(i18n): gettext for quicklinks;

- Added getext in quicklinks template links and modals.

Looking back at the history, it looks like we were using gettext at one point, but then dropped it when it switched to using the `quick_link` macro. Technically, it still used gettext inside of the macro, but as we use the message extractors, they got dropped


wardi:
Might as well remove the gettext from the macro too in case double-translating causes a problem somewhere

wardi (approved):

PR #1435, opened 2023-12-29
More Flash Messages (Translations)

JVickery-TBS:
For: https://github.com/open-data/ckan/pull/163

translations so we can get all these flash messages into our forks, and then contribute to upstream translations

### DIFF
--- a/ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
+++ b/ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2023-12-29 14:13+0000\n"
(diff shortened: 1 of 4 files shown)


ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
```diff
@@ -248,31 +248,86 @@ msgstr "Supprimer"
...
 
-#: ckanext/canada/strings.py:43
+#: ckanext/canada/strings.py:44
 msgid "Beta"
 msgstr "Bêta"
 
+#: ckanext/canada/strings.py:45
+#, TODO: add translation, python-format
+msgid "Deleted API token %s"
+msgstr ""
```
wardi, line 258: when we have empty values here does it remove translations that were defined in ckan?

JVickery-TBS: Not sure, but this string is not in CKAN (yet, there is an upstream PR for that https://github.com/ckan/ckan/pull/7996)

wardi: ok, if the strings we've added aren't upstream then this is fine once we have translations here.

PR #1438, opened 2024-01-19
Repurpose Click Method for Resubmit All DataStore Resources

JVickery-TBS:
feat(cli): repurposed command method;

- Modified click method to add `empty-only` argument.

@wardi already had this click method to fix all of those empty DataStore resources from the misuse of the Resource Validation Options. So just changed it up a bit to add an `empty-only` argument. Otherwise, it will resubmit all DataStore resources of `upload` type.

### DIFF
--- a/ckanext/canada/scripts.py
+++ b/ckanext/canada/scripts.py
@@ -227,15 +227,16 @@ def set_datastore_false_for_invalid_resources(resource_id=None, delete_table_vie
         _success_message('There are no Invalid Resources that have the datastore_active flag at this time.')
 
 
-@canada_scripts.command(short_help="Re-submits valid, empty DataStore Resources to Validation OR Xloader.")
+@canada_scripts.command(short_help="Re-submits valid DataStore Resources to Validation OR Xloader.")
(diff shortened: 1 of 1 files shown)


ckanext/canada/scripts.py
```diff
@@ -266,13 +267,18 @@ def resubmit_empty_datastore_resources(resource_id=None, verbose=False, quiet=Fa
                 continue
             if resource_id in datastore_tables:
                 try:
-                    count = _get_datastore_count(context, resource_id, verbose=verbose, status=status, max=max)
-                    if int(count) == 0:
-                        if verbose:
-                            click.echo("%s/%s -- Resource %s has %s rows in DataStore. Let's fix this one..." % (status, max, resource_id, count))
+                    if empty_only:
+                        count = _get_datastore_count(context, resource_id, verbose=verbose, status=status, max=max)
+                        if int(count) == 0:
```
wardi, line 272: This pattern suggests that _get_datastore_count should already be returning an int.

wardi (approved):

PR #1439, opened 2024-01-25
Resource View & Data Dictionary Activities

JVickery-TBS:
https://github.com/open-data/ckan/pull/164 <- moved resource view activities from core into this extension.

Added some styling and templating for the DS Audit activities.

Added `created datastore` to the changed packages since so that Data Dictionary changes will create publish to portal pools


ckanext/canada/plugins.py
```diff
@@ -668,11 +669,14 @@ def get_helpers(self):
...
             'flash_error',
             'flash_success',
+            'get_resource_view',
+            'resource_view_type',
         ])
 
     # IActions
     def get_actions(self):
         return {
+                #'datastore_create': logic.canada_datastore_create,
```
wardi, line 679: please remove instead of leaving commented-out

wardi (approved):

PR #1441, opened 2024-01-29
Resource Format Guessing

JVickery-TBS:


ckanext/canada/plugins.py
```diff
@@ -303,6 +304,15 @@ def before_index(self, data_dict):
         return data_dict
 
 
+    # IResourceController
+    def before_update(self, context, old_dict, new_dict):
+        """
+        Set the old resource url in the context for use in the validator: canada_guess_resource_format
+        """
+        if old_dict.get('url', None) and old_dict.get('package_id', None):  # ensure that it is a resource
+            context['old_resource_url'] = old_dict.get('url', None)
```
wardi, line 313: this doesn't work because the whole package metadata is validated together, not one resource at a time. You can grab the resource from the db to get the current value instead (we already have some validators that do this)

ckanext/canada/assets/internal/canada_guess_mimetype.js
```diff
@@ -0,0 +1,140 @@
+window.addEventListener('load', function(){
+  $(document).ready(function() {
+
+    const lang = $('html').attr('lang').length > 0 ? $('html').attr('lang') : 'en';
+
+    // slightly different error messages for the front-end users.
+    // as the UX is different than the API users.
+    let failedGuessErrorMessage = 'Could not determine a resource format. Please supply a format below.';
+    let failedChoicesErrorMessage = 'Could not determine a valid resource format. Please supply a different format below.';
```
wardi, line 9: I don't think we need error messages, we could leave the format drop-down blank and let the form validation tell them they need to choose something

wardi:
WDYT about adding the "select CSV for table designer resources" feature to your JS here? Then you can steal that ticket from my backlog and close it at the same time.

ckanext/canada/assets/internal/canada_guess_mimetype.js
JVickery-TBS: Yeah that makes sense. The logic method will just raise the error and then it will be added to the errors in the validator method.

ckanext/canada/plugins.py
JVickery-TBS: Okay sounds good, just used the model orm to get the resource by id and check if it exists and has url now. All done inside of the validator method.

JVickery-TBS:
@wardi okay I did above comments. And added in the JS event binding for the Table Designer just to notify users that the fields have been set, but I also needed to add the validators in to force CSV format and UTF-8 charset for tabledesigner url types.

wardi (approved):

PR #1443, opened 2024-02-13
No Org Alert for Members

JVickery-TBS:
fix(views): no org alert;

- Fixed no org alert for org members.

### DIFF
--- a/ckanext/canada/view.py
+++ b/ckanext/canada/view.py
@@ -10,6 +10,7 @@ import unicodecsv
 from codecs import BOM_UTF8
 from six import string_types, PY2
 from datetime import datetime, timedelta
+from sqlalchemy import or_, and_
 
 from ckan.plugins.toolkit import (
     abort,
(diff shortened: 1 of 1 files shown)


ckanext/canada/view.py
```diff
@@ -948,8 +949,15 @@ def _log_api_access(request_data, pkg_dict):
...
-    '''flash_notice if logged-in user can't actually do anything yet'''
-    if h.check_access('package_create'):
+    '''flash_notice if logged-in user can't actually do anything yet.'''
+    has_org_membership = model.Session.query(model.Member.id).filter(
+                            and_(model.Member.state == 'active',
+                                 model.Member.table_name == 'user',
+                                 model.Member.table_id == g.userobj.id,
+                                 or_(model.Member.capacity == 'admin',
+                                     model.Member.capacity == 'editor',
+                                     model.Member.capacity == 'member'))).first()
```
wardi, line 959: I don't think there *are* any other capacities so this part doesn't do anything.

ckanext/canada/view.py
```diff
@@ -948,8 +949,15 @@ def _log_api_access(request_data, pkg_dict):
 
 
 def notice_no_access():
-    '''flash_notice if logged-in user can't actually do anything yet'''
-    if h.check_access('package_create'):
+    '''flash_notice if logged-in user can't actually do anything yet.'''
+    has_org_membership = model.Session.query(model.Member.id).filter(
```
wardi, line 953: it's a bit of a layering violation to go straight to the db from a view. We could call the `organization_list_for_user` action instead so that this code doesn't break with future db changes

wardi: Yes this action returns the list for the current user, not one passed in as a parameter. The performance shouldn't be an issue, it's only on login and register IIUC. If we wanted to add a parameter to return only the org ids from this action we could do that as a separate thing but it's not critical

wardi:
Need to update a test?

JVickery-TBS:
Just gunna run the tests again...I have had this issue before in which CircleCI SMTP fails or something

PR #1444, opened 2024-02-20
add registry_access field to organization schema

RabiaSajjad:


ckanext/canada/validators.py
```diff
@@ -491,3 +491,28 @@ def canada_guess_resource_format(key, data, errors, context):
...
+    if is_sysadmin(context['user']):
+        return
+
+    original = ''
+    org = context.get('group')
+    if org:
+        original = org.extras.get('registry_access', [])
+    value = data.get(key, [])
+
+    if not value:
```
wardi, line 509: Need to check for `missing` here too

ckanext/canada/schemas/organization.yaml
```diff
@@ -227,3 +227,26 @@ fields:
...
+      en: "Public network"
+      fr: "TBD"
+    value: "public"
+  - label:
+      en: "Government of Canada network"
+      fr: "TBD"
+    value: "goc"
+  form_snippet: radio.html
+  display_snippet: select.html
+  validators: protect_registry_access scheming_choices
```
wardi, line 252: Something like this might work to make it appear that all current orgs are set to `goc`

```suggestion
  validators: protect_registry_access scheming_choices
  output_validators: default(goc)
  required: true
```

ckanext/canada/validators.py
wardi: if you mark it required and add a `scheming_required` validator to the front of the list we won't get missing at this point

bin/migrate-org-registry-access.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+# coding=utf-8
+
+"""
+Migrate existing organizations to add default value for
+registry_access field in schema defined in organization.yaml
+
+Usage: ckanapi dump organizations --all -c $CONFIG_INI | python
+migrate-org-registry-access.py --output outfile.jsonl """
```
wardi, line 9: `jq` is great for this kind of thing. We could replace this whole script with:
```suggestion
Usage: ckanapi dump organizations --all -c $CONFIG_INI | jq -c '.+{"registry_access":"goc"}' > outfile.jsonl """
```

RabiaSajjad: in that case, we don't need a script. I can remove it and add this command in the ticket as an important step when deploying in production?

ckanext/canada/validators.py
```diff
@@ -491,3 +491,28 @@ def canada_guess_resource_format(key, data, errors, context):
...
+
+def protect_registry_access(key, data, errors, context):
+    """
+    Ensure the registry_access field is not changed by an unauthorized user.
+    """
+    if is_sysadmin(context['user']):
+        return
+
+    original = ''
+    org = context.get('group')
```
JVickery-TBS, line 504: @RabiaSajjad best not to use objects passed around in the context variable.
```
org_id = data.get(key[:-1] + ('id',))
if org_id:
    org = model.Group.get(org_id)
```

ckanext/canada/validators.py
```diff
@@ -491,3 +491,28 @@ def canada_guess_resource_format(key, data, errors, context):
...
+        original = org.extras.get('registry_access', [])
+    value = data.get(key, [])
+
+    if not value:
+        data[key] = original
+        return
+    elif value == original:
+        return
+    else:
+        errors[key].append("Cannot change value of registry_access field"
```
JVickery-TBS, line 515: Did we need this error message to be translated? If so, wrap in gettext (`_`)?

ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
```diff
@@ -450,13 +450,21 @@ msgstr "L’objet JSON doit contenir la clé \"en\""
 msgid "JSON object must contain \"fr\" key"
 msgstr "L’objet JSON doit contenir la clé \"fr\""
 
+#: ckanext/canada/validators.py:527
+msgid "JSON object must contain \"fr\" key"
+msgstr "L’objet JSON doit contenir la clé \"fr\""
+
```
wardi, line 456: is this comment for the new string below? The msgid/msgstr looks like a duplication from above?

JVickery-TBS (approved):

PR #1446, opened 2024-02-27
add ux changes to 'View on Map' button

RabiaSajjad:


ckanext/canada/templates/public/package/snippets/resources_list.html
```diff
@@ -0,0 +1,15 @@
...
+
+{% block resource_list_title %}
+  <h2>{{ _('Data and Resources') }}
+  <div class="pull-right mrgn-bttm-md">
+    {% if 'fgp_viewer' in pkg.get('display_flags', []) %}
+      {% set openmap_uri = 'openmap' %}
+      {% if h.lang() == 'fr' %}
+        {% set openmap_uri = 'carteouverte' %}
+      {% endif %}
+      <a href="{{ h.adv_search_url() }}/{{ openmap_uri }}/{{ pkg.name }}" class="btn btn-success" target="_blank">
```
wardi, line 11: you could save some repetition and business-logic-in-templates here by creating a helper for the openmap_uri instead

RabiaSajjad: good idea. I have added a helper `fgp_viewer_url` and remove old helper `fgp_url` which was not used by any templates

ckanext/canada/templates/public/package/snippets/resource_item.html
```diff
@@ -88,20 +88,14 @@
         </a>
       </li>
       {% endif %}
-      {% if 'fgp_viewer' in pkg.get('display_flags', []) %}
-        {% if res.url and h.is_url(res.url) and res.format not in ['PDF','HTML', 'PNG', 'CSV', 'TXT',] %}
+      {% set fgp_url = h.fgp_viewer_url(pkg) %}
+      {% if fgp_url and res.url and h.is_url(res.url) and res.format not in ['PDF','HTML', 'PNG', 'CSV', 'TXT',] %}
```
wardi, line 92: do you want to pull the format checking logic into the helper too, or does this need to be different in resource_item.html and resources_listh.html?

RabiaSajjad: I am displaying the link at two places on the dataset page. One is just relocating the existing "View on Map" button closer to the resource list so that it is more prominent. The other is within the dropdown list for certain resources. I am only doing a format check with the second option.

wardi (approved):

PR #1448, opened 2024-03-01
30 Day Activity Text

JVickery-TBS:
There have been some questions about the Activity Streams. So just adding in an info message at the top of the activity lists, saying that it is gunna be showing only 30 days of activity.


wardi (approved):
30 days seems really short, can't remember how we came to that decision

JVickery-TBS:
@wardi Ross had to look at your CKAN databases slowly go over 100GB over the years and never wants to see it ever again hahaha.

I mean, we did JUST have the issue with those PD records and they skidded by the 30 days by a week.

If we used an environment variable for this, we could use the envvar in the CRON script, and use it in `update_config` in the Public plugin class to set a config value to then show in this message? Would allow us to easily change this value whenever we wanted to.

We have until the 26th of this month to decide if we want to allow for a longer duration of activity

PR #1449, opened 2024-03-01
Username Recovery

JVickery-TBS:
feat(views): username recovery;

- Recover username view and templates.

Though 2.10 will have username/email login, biz still sees a use to have this functionality because they get a fair amount of intake emails regarding this.

There are conditions to GC Notify being loaded, because of this: https://github.com/open-data/ckanext-gcnotify/commit/9423b8f79e50a0dbcd437bda957478bdab1860e8

I have also made it to always include the username inside of the password reset emails: https://github.com/open-data/ckanext-gcnotify/commit/7bbbeb6ebe1475373ef0a99344a89b6af1c113ee , hopefully that will also help the biz team on these intakes.

### DIFF
--- a/ckanext/canada/view.py
+++ b/ckanext/canada/view.py
@@ -23,6 +23,7 @@ from ckan.plugins.toolkit import (
     request,
     render
 )
+import ckan.lib.mailer as mailer
 from ckan.lib.base import model
 import ckan.lib.jsonp as jsonp
 from ckan.lib.helpers import (
@@ -75,6 +76,10 @@ if PY2:
     from cStringIO import StringIO
 else:
     from io import StringIO
+
+from logging import getLogger
+
+log = getLogger(__name__)
 
 MAX_JOB_QUEUE_LIST_SIZE = 25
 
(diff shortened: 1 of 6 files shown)


ckanext/canada/view.py
```diff
@@ -228,6 +233,67 @@ def post(self):
...
+        if user_list:
+            for user_dict in user_list:
+                username_list.append(user_dict['name'])
+
+        if not username_list:
+            log.info('User requested username recovery for unknown email: {}'
+                     .format(email))
+        else:
+            log.info('Emailing username recovery to email: {}'
+                     .format(email))
```
wardi, line 275: usually these things don't give away whether there are any user accounts with the email for privacy reasons. Worth thinking about with the registry accessible on the public internet at some point.

JVickery-TBS: @wardi I copied the core code from Password Reset for this, and modified a bit to just send the one email with a list of usernames.

This is just the system logging that core also had.

It will always do the flash message:
```
An email has been sent to you containing your username(s). (unless the account specified does not exist)
```
and redirect user back to home page, just like the password reset form

wardi (approved):

PR #1456, opened 2024-03-08
Reclassification Sub Group and Code Prefices

JVickery-TBS:
Feature/reclass sub groups

@wardi I just did the prefixing in our custom templates in this plugin. But let me know if the `form_choices_prefix_code` stuff should be done in Recombinant here instead: https://github.com/open-data/ckanext-recombinant/blob/master/ckanext/recombinant/helpers.py#L157

### DIFF
--- a/ckanext/canada/tables/choices/classification_group.yaml
+++ b/ckanext/canada/tables/choices/classification_group.yaml
@@ -1,3 +1,7 @@
+#
+# This file is managed manually from the datasource:
+# https://www.canada.ca/en/treasury-board-secretariat/services/collective-agreements/occupational-groups.html
+#
 AC:
   en: Actuarial Science
   fr: Actuariat
@@ -440,10 +444,13 @@ SO-MAO:
 SO-RAD:
   en: Ships' Officers - Radio
   fr: Officiers de navire - Radio
-SR-APA:  # 4th year
-  en: Ship Repair - Apprentice
-  fr: Réparation des navires - Apprentissage
-SR-APB:  # 5th year
+SR-APA:
+  en: Ship Repair - Apprentices 4 years
+  fr: Réparation des navires - Apprentissage, 4 ans
+SR-APB:
+  en: Ship Repair - Apprentices 5 years
+  fr: Réparation des navires - Apprentissage, 5 ans
+SR-APP:
   en: Ship Repair - Apprentice
   fr: Réparation des navires - Apprentissage
 SR-BOB:
--- a/ckanext/canada/tables/choices/classification_group.yaml
+++ b/ckanext/canada/tables/choices/classification_group.yaml
@@ -152,7 +152,7 @@ GL-AMW:
   en: General Labour and Trades - Ammunition Working
   fr: Manoeuvres et hommes de métier - Fabrication de munitions
 GL-BOB:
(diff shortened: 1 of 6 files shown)


ckanext/canada/tables/choices/classification_group.yaml
```diff
@@ -29,181 +67,482 @@ CT-FIN:
...
+NU-CHN:
+  en: Nursing - Community Health
+  fr: Sciences infirmières - Infirmières en santé communautaire
+NU-EMA:
+  en: Nursing - Medical Adjudicator Nursing
+  fr: Sciences infirmières - Évaluateurs médicaux
+NU-HOS:
+  en: Nursing - Hospital Nursing
+  fr: Sciences infirmières - Infirmières d’hôpital
+#TODO: find OC group/classification/sub-group
```
wardi, line 307: These changes are all manual? Would be great to add a comment at the top with the data sources used

ckanext/canada/tables/choices/classification_group.yaml
```diff
@@ -29,181 +67,482 @@ CT-FIN:
...
+SO-INS:
+  en: Ships' Officers - Instructor
+  fr: Officiers de navire - Instructeur
+SO-MAO:
+  en: Ships' Officers - Marine Operations
+  fr: Officiers de navire - Navigation maritime
+SO-RAD:
+  en: Ships' Officers - Radio
+  fr: Officiers de navire - Radio
+SR-APA:  # 4th year
```
wardi, line 443: should the year be in the label?

wardi:
It's fine to do this in our add/edit forms because they're not part of ckanext-recombinant. You could roll the new code into a form-choices-specific helper if you want to reduce some duplicated logic in the templates but it's not strictly required.

ckanext/canada/tables/choices/classification_group.yaml
JVickery-TBS: Ah yeah these were from an archived classification. Got the labels and translations in now.

ckanext/canada/tables/choices/classification_group.yaml
JVickery-TBS: Comment on top of file there with datasource and statement that it is managed manually

JVickery-TBS:
@wardi surprise! Apparently we have this functionality already. `form_full_text_choices` is used in NAP5, but it works here. Just made the `choice_fields` code a bit more reusable in the views script now.

ckanext/canada/tables/reclassification.yaml
```diff
@@ -110,7 +110,7 @@ resources:
     format_type: Controlled List
     validation: This field must not be empty
     excel_full_text_choices: True
-    form_choices_prefix_code: True
+    form_full_text_choices: True
```
wardi, line 113: prefix_code is a better name because the form always shows the full text. For excel the default is to only show the code

wardi: :shrug: but if that's how we've already implemented it I guess it's fine

JVickery-TBS: @wardi okay I changed it: `form_full_text_choices` -> `form_choices_prefix_code`. Updated usages in NAP5 schema as well.

wardi (approved):

PR #1457, opened 2024-03-15
Promote/Demote Sysadmins

JVickery-TBS:
feat(views): promote sysadmin;

- Promote/demote sysadmin view and form button.
- Minore layout and style fixes.


ckanext/canada/view.py
```diff
@@ -1200,3 +1201,61 @@ def members(id):
...
+        user.sysadmin = sysadmin
+        model.Session.add(user)
+        model.repo.commit_and_remove()
+        if sysadmin:
+            log.info('%s promoted %s to a sysadmin', g.user, user.name)
+            h.flash_success(_('Promoted %s to a sysadmin') % user.name)
+        else:
+            log.info('%s demoted %s from a sysadmin', g.user, user.name)
+            h.flash_success(_('Demoted %s from a sysadmin') % user.name)
+    except Exception:
```
wardi, line 1245: Can you catch something specific here? Otherwise might be better to raise the error so it doesn't get hidden from the logs

wardi:
Everything but the specific tbs logic would be good for an upstream contribution too.

ckanext/canada/view.py
JVickery-TBS: @wardi I went into Core code and SQLalchemy to see what the commit/remove would raise. Does not look like anything. And Core CKAN also does not try/catch any of these.

So should be fine without the try/catch

JVickery-TBS:
@wardi because our user list is so long, and Core CKAN (at least in 2.9) does not have a way to order user list action bt desc or asc, I have just added an admin view in our plugin to show the sysadmins.

See view, logic, and auth methods `list_sysadmins`

JVickery-TBS:
Ah...well... @wardi this is embarrassing...this uhh is already a feature in CKAN Core and it is so much better in Core! hahaha damn. Okay I am just going to go backport the Core one into this PR I guess

JVickery-TBS:
Oh okay, it is embarrassing for both of us. https://github.com/ckan/ckan/pull/5522 <- you pulled it. Should have remembered that ONE pull request from FOUR YEARS AGO

JVickery-TBS:
@wardi okay we have something going now.

https://github.com/open-data/ckan/pull/169 <- here is the PR for the cherry pick. Plus a small change as we need to output all of our different Authorization error messages from this plugin now.

I have updated the code here.

Major things are:
* the `ckan-admin` view override to exclude the site user from the list.
* the chained `user_update` action method to harden the `sysadmin` value if it is being changed.

JVickery-TBS:
Gotta fix a couple things, and writing automated tests now.

JVickery-TBS:
New test cases work.

Just have to:
- do an auth call to get auth user obj in context;
- make a condition so if auth user obj is the site user, we pass over (e.g. `ckanapi` usage).

JVickery-TBS:
@wardi okay this one should be good for review now

ckanext/canada/logic.py
```diff
@@ -460,3 +464,67 @@ def canada_job_list(up_func, context, data_dict):
...
+
+    We have to do it in this chained action, and not auth method
+    because sysadmins skip auth checks.
+    """
+    user = model.User.get(get_or_bust(data_dict, 'id'))
+
+    if not user:
+        raise ObjectNotFound(_('User not found'))
+
+    if data_dict.get('sysadmin', user.sysadmin) != user.sysadmin:
```
wardi, line 484: validation of the data_dict hasn't happened here yet, so if someone passes a string version of this parameter it will always show as not equal.

ckanext/canada/logic.py
```diff
@@ -460,3 +464,67 @@ def canada_job_list(up_func, context, data_dict):
...
+        if user.name == site_id:
+            raise NotAuthorized(_('Cannot modify sysadmin privileges for system user'))
+
+        # extra logging
+        if data_dict.get('sysadmin'):
+            log.info('%s promoted %s to a sysadmin', contextual_user.name, user.name)
+        else:
+            log.info('%s demoted %s from a sysadmin', contextual_user.name, user.name)
+
+    return up_func(context, data_dict)
```
wardi, line 530: If some staging sysadmin is really determined they can still assign themselves to tbs then have full permission to do anything. This is a lot of code to prevent something unlikely and only possible on staging. If it's a concern why don't we remove the sysadmin permissions from non-tbs accounts on staging instead?

wardi (commented):
Other than my comment above this looks good.

ckanext/canada/logic.py
JVickery-TBS: @wardi I just copied whatever was in core: https://github.com/ckan/ckan/blob/master/ckan/logic/action/update.py#L797

ckanext/canada/logic.py
JVickery-TBS: Okay that makes sense. Did we still want to leave some of this code in? Such as preventing the modification of the system user? or do we not care about that too much?

JVickery-TBS: How about I just keep in the extra logging, and we can just remove all the other logic and just go with what Core does.

ckanext/canada/logic.py
wardi: Not sure what you mean, I don't see where core is doing `data_dict.get('sysadmin', user.sysadmin) != user.sysadmin`. In core there isn't a place to hook in to the action after validation but before model_save. This could be a new validator in the update user schema instead.

JVickery-TBS: @wardi oops my bad, read that comment wrong. Okay I am now calling the validate and then checking `sysadmin` in the validated data

ckanext/canada/logic.py
JVickery-TBS: @wardi okay, so I removed all of the TBS membership checks. Kept the no modify system user, and no modify your own privs. And kept in the extra logging

Will make a ticket to do an upstream contribution to the User Update schema later.

ckanext/canada/logic.py
```diff
@@ -476,12 +478,20 @@ def canada_user_update(up_func, context, data_dict):
...
     if not user:
         raise ObjectNotFound(_('User not found'))
 
-    if data_dict.get('sysadmin', user.sysadmin) != user.sysadmin:
+    data, errors = _validate(data_dict, schema, context)
+    if errors:
+        session.rollback()
+        raise ValidationError(errors)
+
+    if data.get('sysadmin', user.sysadmin) != user.sysadmin:
```
wardi, line 494: If this was done as a validator in the user update schema we wouldn't have to validate all fields passed twice. Not a big deal for our site but important for an upstream contribution.

JVickery-TBS: Yeaaah, I am having troubles with the tests when calling validate myself in this chained logic method...so I am just going to make a validator and update the schema in the chained logic method and see if that works.

wardi: or you could update the schema in ckan because that's what the upstream PR would do

PR #1458, opened 2024-03-19
Empty Resource is Safe

JVickery-TBS:
feat(dev): allow empty resources;

- Allow empty resources if no url and no upload.
- Resource view tab layout fixes.
- Add PBIX resource format.

### DIFF
--- a/ckanext/canada/assets/public/canada_public.css
+++ b/ckanext/canada/assets/public/canada_public.css
@@ -586,25 +586,25 @@ ul.resource-list > li.resource-item span.format-label[data-format="pbix"]{
 }
 /* END - Resource Action Buttons - END */
 
-/* Resource Views List Bar */
-body ul.nav-tabs a[href*="?view_id="]{
+/* Resource Views Lists */
+body ul.nav li.view-item a{
   display: flex;
   flex-direction: row;
   flex-wrap: nowrap;
   width: 100%;
   align-items: center;
 }
-body ul.nav-tabs a[href*="?view_id="] i.icon{
+body ul.nav li.view-item a i.icon{
   position: relative !important;
 }
-body ul.nav-tabs a[href*="?view_id="] i.icon::before{
+body ul.nav li.view-item a i.icon::before{
   position: absolute !important;
   top: 50%;
   left: 50%;
   transform: translate(-50%, -50%);
   font-size: 24px !important;
 }
-/* END - Resource Views List Bar - END */
+/* END - Resource Views Lists - END */
 
 /*#################
 ##      END      ##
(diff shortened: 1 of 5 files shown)


ckanext/canada/assets/public/canada_public.css
```diff
@@ -439,6 +439,16 @@ ul.resource-list > li.resource-item > div.canada-resource-list-badges a.validati
 ul.resource-list > li.resource-item > div.canada-resource-list-badges{
   align-self: last baseline;
 }
+ul.resource-list > li.resource-item span.format-label[data-format="pbix"]{
+  background-image: url("/static/img/power_bi.svg");
```
wardi, line 443: this won't work on the portal

ckanext/canada/schemas/presets.yaml
```diff
@@ -3457,6 +3457,8 @@ presets:
     - value: ODT
       mimetype: application/vnd.oasis.opendocument.text
       replaces: [odt]
+    - value: PBIX
+      replaces: [pbix]
```
wardi, line 3461: these `replaces` lists were only used for importing data from an earlier system

ckanext/canada/assets/public/canada_public.css
```diff
@@ -576,6 +586,26 @@ ul.resource-list > li.resource-item > div.canada-resource-list-badges{
 }
 /* END - Resource Action Buttons - END */
 
+/* Resource Views List Bar */
+body ul.nav-tabs a[href*="?view_id="]{
```
wardi, line 590: this quite a hack. Is it really that hard to add another icon to the ones in the ckan formats?

ckanext/canada/schemas/presets.yaml
JVickery-TBS: @wardi we need the replaces for the Format guessing as the Core CKAN guess format does not return them in uppercase. And there are cases in which the guessing format could return jpeg instead of jpg, in which case we want the replacement.

ckanext/canada/assets/public/canada_public.css
JVickery-TBS: @wardi yeah, sadly there was no block to easily extend in that template. And there were no specific classes used in any of the html structure, so had to do the dirty CSS selection.

wardi: A small change in ckan to add a class based on the format would make this a lot cleaner right? Might be worth it because this kind of solution doesn't scale well.

JVickery-TBS: Okay I will make an upstream PR with some block or html class

JVickery-TBS: @wardi we are actually fully overriding the template in which I would need to add these classes to.

So I have done it in this PR, and will make a ticket to do upstream contrib later.

PR #1459, opened 2024-03-22
Static Dialect Validation Checks & Tabulator Subclasses

JVickery-TBS:
Requires and Required-by:

https://github.com/open-data/ckanext-xloader/pull/25
https://github.com/open-data/ckanext-validation/pull/13
https://github.com/open-data/goodtables/pull/3


JVickery-TBS:
Okay this is good to go now and has gone through so so so much testing locally. But will test on test ad staging vigorously as well.

Did a whole different take on it, as it was almost impossible to really bake this into the CKAN framework (Validation and Xloader hooks). This was mainly because of how the two plugins use Tabulator, as well as all of the private/protected methods and properties the Tabulator library has.

So it is mainly all done in the Canada plugin now, monkey-patching the Xloader and Goodtables (Validation) usages of the Tabulator Stream class. Our fork of goodtables has some minor additions to parse the custom errors.

The premise of this is that the subclasses won't actually do anything unless the `ckanext.canada.tabulator_dialects` or `ckanext.canada.tabulator_encoding` config options are set.

ckanext/canada/tabulate.py
```diff
@@ -0,0 +1,311 @@
...
+#       CSV dialect value is low, but never zero...
+ERROR_SPLITTER = '||||'
+
+
+class LimitStreamLogging(Filter):
+    logged_messages = []
+    def filter(self, record):
+        current_log = (record.module, record.levelno, record.msg)
+        if current_log not in self.logged_messages:
+            self.logged_messages.append(current_log)
```
wardi, line 35: this will grow without bounds and searching a list is O(n). You could use a set to make the search O(1) but the growth growth will still be a problem.

Can we store just the last message or last n messages to solve the problem?

ckanext/canada/tabulate.py
```diff
@@ -0,0 +1,311 @@
...
+
+from logging import getLogger, Filter
+
+#FIXME: the chances of a user using quad pipes as a
+#       CSV dialect value is low, but never zero...
+ERROR_SPLITTER = '||||'
+
+
+class LimitStreamLogging(Filter):
+    logged_messages = []
```
wardi, line 31: this is a (global) class variable, better to use an instance variable.

ckanext/canada/tabulate.py
```diff
@@ -0,0 +1,311 @@
...
+        # a bit nasty, but need to compare from __mangle__prepare_dialect
+        guessed_dialect = guessed_dialect.__dict__
+        static_dialect = static_dialect.__dict__
+        guessed_delimiter_var = 'delimiter'
+        guessed_quoteChar_var = 'quotechar'
+        guessed_doubleQuote_var = 'doublequote'
+
+    if guessed_dialect and static_dialect:
+        if guessed_delimiter_var in guessed_dialect and guessed_dialect[guessed_delimiter_var] != static_dialect['delimiter']:
+            raise TabulatorException('{0}{1}{2}'.format(
```
wardi, line 284: This is a little odd, you could do something like
```suggestion
            raise TabulatorException(''.join(
```
instead, or even better: `ERROR_SPLITTER.join(` with only two parameters.

PR #1462, opened 2024-04-04
translate site_title in meta tags

RabiaSajjad:


ckanext/canada/templates/public/base.html
```diff
@@ -5,12 +5,12 @@
...
 
   {% block og_meta_title %}
-    <meta property="og:title" content="{{ self.title() }}">
+    <meta property="og:title" content="{{ _(self.title()) }}">
   {% endblock %}
   {% block adobe_analytics_meta %}
     {% if not h.is_registry() %}
       {% block adobe_analytics_title %}
-        <meta property="dcterms:title" content="{{ self.title() }}" />
+        <meta property="dcterms:title" content="{{ _(self.title()) }}" />
```
JVickery-TBS, line 13: @RabiaSajjad this will just fully wrap the `title` block in the template in gettext which might cause issues with doubling into gettext.

We will need to update it in the `title` block instead. Looking at it, the `g.site_title` is already in a gettext call. So we may just need to add whatever our `ckan.site_title` is from our INIs into the `strings.py` and do `python setup.py extract_messages` and `python setup.py update_catalog`

JVickery-TBS (changes requested):
Just the above comments. So we only need to add our site titles to the `strings.py` so they get picked up by Babel in the message extractions. Won't need any template changes

PR #1463, opened 2024-04-08
add trigger to trim contract pd fields

RabiaSajjad:


JVickery-TBS (approved):
Looks goooood. Our deployment SHOULD run create/update triggers for our plugin and the Recombinant plugin, so hopefully that works and these get created fine during deploy

PR #1466, opened 2024-04-11
Nightly Make Test Coverage

JVickery-TBS:


ckanext/canada/tests/test_make.py
```diff
@@ -0,0 +1,482 @@
...
+        assert self.ckan_ini
+        self._setup_ini(self.ckan_ini)
+
+        self._setup_pd(type='ati', nil_type='ati-nil')
+
+        make_process = subprocess.Popen(["make upload-ati"], shell=True, cwd=MAKE_PATH, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        stdout, stderr = make_process.communicate()
+
+        assert "Usage:" not in stdout
+        assert "upload-ati] Error" not in stdout
```
wardi, line 141: instead of looking for a specific string, check the error code on the subprocess

PR #1468, opened 2024-04-15
Context Manager for DS Temp User Table

JVickery-TBS:
fix(logic): use context manager;

- Transformed `datastore_create_temp_user_table` into a context manager.
- Chained `datastore_run_triggers` action.

We need to be able to create the temporary table when updating triggers. However, the `ON COMMIT DROP` in the creation of the table seemed to not work with the execution statement in the `datastore_run_triggers` action.

Requires: https://github.com/open-data/ckan/pull/170

Want to confirm that it is fine to do a python context manager with `DROP TABLE` in the exit, instead of `ON COMMIT DROP` in the pSQL table creation.

### DIFF
--- a/ckanext/canada/logic.py
+++ b/ckanext/canada/logic.py
@@ -2,6 +2,7 @@ from ckan.logic.action import get as core_get
 from ckan.logic.validators import isodate, Invalid
 from ckan.lib.dictization import model_dictize
 from ckan import model
+from contextlib import contextmanager
 
 from ckan.plugins.toolkit import (
     get_or_bust,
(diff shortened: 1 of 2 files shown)


JVickery-TBS:
Reworked the context manager class with the `contextmanager` decorator. And as the none `ON COMMIT DROP` did not work with the upsert data, I have now made the context manager method a bit more dynamic, allowing for either `ON COMMIT DROP` or not.

wardi:
`ON COMMIT DROP` is preferred when it works because postgres will guarantee it gets run. But, it is a temporary table so it should get removed regardless, kind of a belt-and-suspenders thing.

wardi (approved):

PR #1472, opened 2024-05-09
Tighten Viewable Routes

JVickery-TBS:
feat(logic,views): tighten view auths;

- Added aborts and redirects to PD searches.


thriuin:
Seems good to me, but because of the plug-in changes might be nice to check with @wardi too if it can wait to Friday.

ckanext/canada/auth.py
```diff
@@ -35,3 +35,11 @@ def view_org_members(context, data_dict):
 
 def registry_jobs_running(context, data_dict):
     return {'success': _is_reporting_user(context)}
+
+
+def group_list(context, data_dict):
+    return {'success': context.get('user')}
```
wardi, line 41: Do you mean maybe
```suggestion
    return {'success': bool(context.get('user'))}
```

wardi (commented):
Just those two auth functions returning unusual types, otherwise LGTM

PR #1473, opened 2024-05-10
Search Facet Fixes

JVickery-TBS:
Fixed up a bunch of issues with the search facets.

Removed a bunch of old, unused code.


ckanext/canada/helpers.py
```diff
@@ -758,3 +766,98 @@ def fgp_viewer_url(package):
...
+                preset_choices.append({'value': pd_type, 'label': _(h.recombinant_get_chromo(pd_type).get('title'))})
+
+    def remove_filter_button_html(_field, _value, s_extras, _choices):
+
+        extrs = s_extras.copy() if s_extras else {}
+        extrs.update(request.view_args)
+        link = h.remove_url_param(_field, value=_value, extras=extrs)
+
+        html = '<a href="{link_href}" class="btn btn-info btn-xs mrgn-lft-sm mrgn-bttm-sm facet-pill" title="{link_title}">'.format(
+                    link_href=link,
```
wardi, line 829: in general I avoid building HTML in helpers because it's easy to miss escaping a value and create opportunities for XSS or broken HTML. Consider returning the data you need from the helper and using a template to render the HTML elements instead.

Also repeatedly appending to is an O(n²) operation (although I think newer pythons have a hack to speed them up a bit)

ckanext/canada/plugins.py
```diff
@@ -245,6 +247,12 @@ def before_search(self, search_params):
             search_params['extras']['imso_approval'] = u'true'
             search_params['fq'] += '+ready_to_publish:"true", +imso_approval:"true", -portal_release_date:*'
 
+        # CKAN Core search view wraps all fq values with double quotes.
+        # We need to remove double quotes from the portal_release_date queries.
```
wardi, line 251: do they also escape values inside the quotes that we might need to worry about?

JVickery-TBS: Does not look like anything is escaped: https://github.com/ckan/ckan/blob/master/ckan/views/dataset.py#L193

ckanext/canada/helpers.py
JVickery-TBS: @wardi doing a list of tuples now in the helper for the return. And outputting the html in jinja. Are you able to review again? Thanks!

ckanext/canada/plugins.py
wardi: that's concerning

JVickery-TBS: Haha upstream issue??

ckanext/canada/templates/internal/admin/publish_search.html
```diff
@@ -55,7 +55,11 @@ <h2 class="mrgn-tp-0 mrgn-bttm-0">
           {% block filter_list %}
             {% for field in g.fields_grouped %}
               {% if field != 'ready_to_publish' and field != 'imso_approval' %}
-                {{ h.search_filter_pill(field, extras) }}
+                {% for link_label in h.search_filter_pill_link_label(field, extras) %}
+                  <a href="{{ link_label[0] }}" class="btn btn-info btn-xs mrgn-lft-sm mrgn-bttm-sm facet-pill" title="{{ _('Remove') }}">
+                    {{ link_label[1] }}&nbsp;<i class="fa fa-times" aria-hidden="true"></i></a>
```
wardi, line 60: this is prettier and loudly fails if you get anything but two items back
```suggestion
                {% for link, label in h.search_filter_pill_link_label(field, extras) %}
                  <a href="{{ link }}" class="btn btn-info btn-xs mrgn-lft-sm mrgn-bttm-sm facet-pill" title="{{ _('Remove') }}">
                    {{ label }}&nbsp;<i class="fa fa-times" aria-hidden="true"></i></a>
```

ckanext/canada/plugins.py
wardi: yes to me it says "I'm just praying there aren't any any quotes or other characters that need special handling here"

JVickery-TBS:
@wardi okay changes applied. (forgot that you can go right into tuple values from list loop in python)

ckanext/canada/templates/public/snippets/search_form.html
```diff
@@ -86,15 +82,15 @@ <h2 class="mrgn-tp-0 mrgn-bttm-0">
...
-  {% for field in c.fields_grouped %}
-    {{ cr.filter_buttons(h, c, field, extras) }}
-  {% endfor %}
-  </ul>
-</div>
+  <div class="filter-list">
+    {% for field in g.fields_grouped %}
+      {% for link_label in h.search_filter_pill_link_label(field, extras) %}
+        <a href="{{ link_label[0] }}" class="btn btn-info btn-xs mrgn-lft-sm mrgn-bttm-sm facet-pill" title="{{ _('Remove') }}">
+          {{ link_label[1] }}&nbsp;<i class="fa fa-times" aria-hidden="true"></i></a>
```
wardi, line 89:
```suggestion
      {% for link, label in h.search_filter_pill_link_label(field, extras) %}
        <a href="{{ link }}" class="btn btn-info btn-xs mrgn-lft-sm mrgn-bttm-sm facet-pill" title="{{ _('Remove') }}">
          {{ label }}&nbsp;<i class="fa fa-times" aria-hidden="true"></i></a>
```

JVickery-TBS: Ah right, in the other place. Thanks!

wardi:
> @wardi okay changes applied. (forgot that you can go right into tuple values from list loop in python)

yeah, and unpacking with nested lists/tuples and * is a thing too. The new `case`, `match` matching rules are even wilder.

PR #1476, opened 2024-05-17
check network access for Request an Account page

RabiaSajjad:


ckanext/canada/view.py
```diff
@@ -202,6 +202,15 @@ def post(self, package_type, id):
...
 class CanadaUserRegisterView(UserRegisterView):
+    def get(self, data=None, errors=None, error_summary=None):
+        # additional check to ensure user can access the Request an Account page
+        # only possible if accessing from GOC network
+        remote_addr = request.headers.get('X-Forwarded-For') or \
+                      request.environ.get('REMOTE_ADDR')
+        if not h.registry_network_access(remote_addr):
+            abort(403, _('Not authorized to see this page'))
+        return super(CanadaUserRegisterView, self).get()
+
```
wardi, line 213: This check should be done in an overridden user_create auth function. That way it applies to the views, links and APIs equally

wardi (approved):

ckanext/canada/auth.py
```diff
@@ -27,6 +28,17 @@ def datastore_upsert(up_func, context, data_dict):
     return up_func(context, data_dict)
 
 
+@chained_auth_function
+def user_create(up_func, context, data_dict=None):
```
JVickery-TBS, line 32: @RabiaSajjad you will have to register this auth method in the `get_auth_functions` method of the `DataGCCAInternal` plugin class in the `plugins.py` script.

And since it will only be registered in the internal plugin class, you would not need the check for:
```
if 'canada_internal' in config.get('ckan.plugins')
```

JVickery-TBS (changes requested):

ckanext/canada/auth.py
JVickery-TBS: Ahhh, another issue here actually...with upstream code. The User view is not using the toolkit (framework) `check_access` method. So our `user_create` auth method will never get called.

Will make a PR against our ckan 2.9 fork/branch to fix this.

JVickery-TBS: @RabiaSajjad okay, figure out the issue. Your `user_create` auth method will also need the decorator:
```
auth_allow_anonymous_access
```

Which can be imported from `ckan.plugins.toolkit`

ckanext/canada/helpers.py
```diff
@@ -773,6 +775,40 @@ def date_field(field, pkg):
         return pkg.get(field).split(' ')[0]
     return pkg.get(field, None)
 
+def registry_network_access(remote_addr):
+    """
+    Only allow requests from GOC network to access
+    user account registration view
+   """
+    try:
```
JVickery-TBS, line 783: @RabiaSajjad just above the try, you may need to add:
```
remote_addr = text_type(remote_addr)
```
to convert bytes and strings to correct types that the ipaddress module expects

JVickery-TBS (approved):

ckanext/canada/auth.py
```diff
@@ -27,6 +31,14 @@ def datastore_upsert(up_func, context, data_dict):
     return up_func(context, data_dict)
 
 
+@chained_auth_function
+@auth_allow_anonymous_access
+def user_create(up_func, context, data_dict=None):
+    # additional check to ensure user can access the Request an Account page
+    # only possible if accessing from GOC network
+    goc_auth()
```
JVickery-TBS, line 39: @RabiaSajjad the auth methods cannot contain the Flask request and aborts as they are used outside of the Request context as well.

So this will just need to do something like:
```
if not registry_network_access():
        return {'success': False}
return up_func(context, data_dict)
```

ckanext/canada/helpers.py
```diff
@@ -774,6 +776,53 @@ def date_field(field, pkg):
     return pkg.get(field, None)
 
 
+def goc_auth():
```
JVickery-TBS, line 779: @RabiaSajjad You can get rid of `goc_auth`, and just put the:
```
remote_addr = request.headers.get('X-Forwarded-For') or \
                           request.environ.get('REMOTE_ADDR')
```
into the `__registry_network_access` method.

and rename that method to `registry_network_access` (no double underscore). and then it does not need to accept the `remote_addr` param.

You can then register the `registry_network_access` as a template helper in the `DataGCCAInternal` plugin class as we will need to use it in these template conditions:
- https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/templates/internal/error_document_template.html#L14
- https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/templates/internal/header.html#L29

ckanext/canada/plugins.py
```diff
@@ -68,6 +68,7 @@ class CanadaSecurityPlugin(CkanSecurityPlugin):
     p.implements(p.IResourceController, inherit=True)
     p.implements(p.IValidators, inherit=True)
     p.implements(p.IConfigurer)
+    p.implements(p.IAuthenticator)
```
JVickery-TBS, line 71: @RabiaSajjad if you add `, inherit=True` to the `IAuthenticator` plugin you will not need to define the `logout` and `abort` methods here.

ckanext/canada/plugins.py
```diff
@@ -102,6 +103,17 @@ def get_validators(self):
                 'canada_security_upload_presence':
                     validators.canada_security_upload_presence}
 
+    # IAuthenticator
+    def identify(self):
+        if p.toolkit.request.path == p.toolkit.url_for(u'user.login'):
```
JVickery-TBS, line 108: @RabiaSajjad you will have to move your `abort` with the message from the helpers into this method. and just use the `if not helpers.registry_network_access()` condition.

`request` should also already be imported here, so you can change `p.toolkit.request` to just `request`.

And the `path == p.toolkit.url_for(u'user.login')` will not work in all cases as it will not pass for the `/en` and `/fr` paths.

The best thing would be to do it by the Flask Blueprints so that if any routes change, this will pick it up. Something like this:
```
        controller, action = p.toolkit.get_endpoint()
        blueprint = '.'.join((controller, action))
        restricted_blueprints = [
            'canada.login',
            'user.login',
        ]
        if blueprint in restricted_blueprints....
```

This would also allow us to restrict the API blueprints like `canada.action` / `api.action`

ckanext/canada/tests/test_webforms.py
```diff
@@ -224,6 +250,15 @@ def test_new_user_missing_fields(self, app):
         assert 'Password: Please enter both passwords' in response.body
 
 
+    @mock.patch('os.path.isfile', _mock_isfile)
+    @mock.patch('__builtin__.open', _mock_open_ip_list)
+    def test_new_user_bad_ip_address(self, app):
+        offset = h.url_for('user.register')
+        response = app.get(offset, extra_environ=self.extra_environ_tester_bad_ip)
+
+        assert response.status_code == 423
```
JVickery-TBS, line 259: @RabiaSajjad as discussed, the 423 is not really Nginx: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/423

So we can just use 403 and condition the templates better.

ckanext/canada/helpers.py
```diff
@@ -774,6 +776,53 @@ def date_field(field, pkg):
...
+                  request.environ.get('REMOTE_ADDR')
+    if not __registry_network_access(remote_addr):
+        # can't return 403 because it shows the Login button
+        return t.abort(423, _(u'This application is only available to authorized '
+                            u'Government of Canada departments and agencies. '
+                            u'Please contact the support team at '
+                            u'open-ouvert@tbs-sct.gc.ca to request access.'))
+
+
+def __registry_network_access(remote_addr):
```
JVickery-TBS, line 790: @RabiaSajjad as the auth function uses this, it would be good to check `flask.has_request_context()` here. As if we are outside the Request context, it is generally tests or system stuff. E.g.
```
from flask import has_request_context
.....
if not has_request_context():
    return True
```

JVickery-TBS (changes requested):

ckanext/canada/helpers.py
```diff
@@ -774,6 +776,52 @@ def date_field(field, pkg):
...
+
+def registry_network_access():
+    """
+    Only allow requests from GOC network to access
+    user account registration view
+   """
+    remote_addr = request.headers.get('X-Forwarded-For') or \
+                  request.environ.get('REMOTE_ADDR')
+    try:
+        client_ip = ipaddress.ip_address(unicode(remote_addr))
```
JVickery-TBS, line 795: @RabiaSajjad Can you also change the usage of `unicode` to `text_type` so this is Python3 ready. I think the `text_type` from `six` should already be imported in the helpers here.

ckanext/canada/helpers.py
```diff
@@ -774,6 +777,48 @@ def date_field(field, pkg):
     return pkg.get(field, None)
 
 
+def registry_network_access():
+    """
+    Only allow requests from GOC network to access
+    user account registration view
+    """
+    if not has_request_context():
```
JVickery-TBS, line 785: @RabiaSajjad Yeah, so looks like the `identify` only happens in `identify_user` in the Flask app class. So there is always be a context. So you can get rid of this `if not has_request_context` and the `has_request_context` import.

JVickery-TBS (changes requested):

ckanext/canada/helpers.py
```diff
@@ -774,6 +776,44 @@ def date_field(field, pkg):
...
+                  request.environ.get('REMOTE_ADDR')
+    try:
+        client_ip = ipaddress.ip_address(text_type(remote_addr))
+    except ValueError:
+        return False
+
+    netlist_path = config.get('ckanext.canada.registry_network_list', '')
+    if not netlist_path:
+        return False
+    if not os.path.isfile(netlist_path):
```
JVickery-TBS, line 794:
```suggestion
    if not os.path.isfile(netlist_path) or not os.path.isfile(netlist_path):
        return False
```

@RabiaSajjad these two conditions can just go into an OR

JVickery-TBS (changes requested):

JVickery-TBS (approved):

JVickery-TBS:
@RabiaSajjad I rebased this now as python3 code is in Master branch. I updated the pytests for python3 too. So we will see if those are successful.

ckanext/canada/plugins.py
```diff
@@ -102,6 +104,22 @@ def get_validators(self):
...
+            'canada.login',
+            'user.login',
+            'user.request_reset',
+            'canada.recover_username',
+            'canada.register',
+            'canada.action',
+            'api.action', # change if need to narrow down the scope
+        ]
+        if blueprint in restricted_blueprints and not helpers.registry_network_access():
+            return abort(403)
```
wardi, line 121: might want to log failed access attempts so we can report on them

JVickery-TBS: Good idea. Do you think we need a separate log handler for "Registry Access"? As we will need to also log failed login attempts and all login sessions.

We could have a new log handler for `ckanext.canada.user_access` to go to its own log file `ckan_registry_access.log`.

Or should we just keep it in the normal logs and have Log Analytics handle all this stuff @wardi ?

wardi: @JVickery-TBS I don't have a strong opinion either way. As long as the data can be pulled out of Log Analytics it shouldn't matter.

RabiaSajjad: @wardi and @JVickery-TBS let's implement logging as a separate feature for both registry network access and login access. It is one of the requirements given to us by imtd/security. I would prefer to implement it as a separate log handler so that it is easier for us to investigate any issues reported by helpdesk.

PR #1477, opened 2024-05-24
Canada Theme Plugin

JVickery-TBS:
Contains the merge of the internal and public templates.

Has a new plugin which helps us with load order of templates.

@wardi major things to look at would be the new `CanadaThemePlugin` class. And the registration of the app global for the `is_registry` value.

This is against the Python 3 branch as the plan is to deploy a week or 2 after the PY3 upgrade. Our plan is to also put this onto staging first (as we have a separate staging branch up to date with py3), so we can further test all of the template changes.

I have tested a lot this week locally, but will be nice to test on staging site.


ckanext/canada/plugins.py
```diff
@@ -61,6 +64,82 @@
...
+
+    # IConfigurer
+    def update_config(self, config):
+        p.toolkit.add_template_directory(config, 'templates')
+        p.toolkit.add_public_directory(config, 'public')
+        p.toolkit.add_resource('public/static/js', 'js')
+        p.toolkit.add_resource('assets/internal', 'canada_internal')
+        p.toolkit.add_resource('assets/datatables', 'canada_datatables')
+        p.toolkit.add_resource('assets/public', 'canada_public')
+        set_app_global('is_registry', plugin_loaded('canada_internal'))
```
wardi, line 80: maybe wouldn't need a separate plugin if you don't call plugin_loaded from inside update_config? You can always check the plugin config instead of using this function.

JVickery-TBS: @wardi the the plugin is more for the loading of templates over other plugins.

The `plugin_loaded` should be fine in the `update_config` hook as all the plugins should already be in the `_PLUGINS` at that point in `ckan.plugins.core`. So even if `canada_theme` is loaded first, it should still be able to see the loaded plugins during `update_config`

ckanext/canada/templates/base.html
```diff
@@ -1,5 +1,15 @@
 {% ckan_extends %}
 
+{% block site_title %}
+  {% if g.is_registry %}
+    {# Registry #}
```
wardi, line 5: I think the {% if %} speaks for itself :-)

ckanext/canada/templates/error_document_template.html
```diff
@@ -38,6 +43,15 @@ <h1>{{ code }} {{ name }}</h1>
           <p class="mrgn-tp-lg">{{ _('We encountered an error') }}</p>
         {% endif %}
       {% endif %}
+      {% if g.debug %}
```
wardi, line 46: in debug mode we get the werkzeug debugger not this page right?

ckanext/canada/templates/help.html
```diff
@@ -0,0 +1,13 @@
...
+  {% if g.is_registry %}
+    {# Registry #}
+    <h1>{{ _('Frequently Asked Questions (FAQ)') }}</h1>
+    {% block primary_content %}{% endblock primary_content %}
+    {{ faq_html|safe }}
+  {% else %}
+    {# Portal #}
+    {{ super() }}
+  {% endif %}
+{% endblock %}
```
wardi, line 13: Just curious, does this kind of construct work? If so it's a little less complicated.
```suggestion
{% if g.is_registry %}
  {% block primary %}
    <h1>{{ _('Frequently Asked Questions (FAQ)') }}</h1>
    {% block primary_content %}{% endblock primary_content %}
    {{ faq_html|safe }}
  {% endblock %}
{% endif %}
```

ckanext/canada/templates/home/index.html
```diff
@@ -0,0 +1,7 @@
+{% ckan_extends %}
+
+{% if not g.is_registry %}
+  {# Portal #}
+  {# hide CKAN front page on public site #}
+  {{ h.redirect_to('dataset.search') }}
```
wardi, line 6: I don't think this works with flask does it?

ckanext/canada/templates/snippets/user.html
```diff
@@ -0,0 +1,69 @@
...
+                      </a>
+                    </li>
+                  {% endif %}
+                {% endblock %}
+                {% block header_account_log_out_link %}{% endblock %}
+              {% endblock %}
+            {% else %}
+              {% block header_account_notlogged %}
+                {% if g.is_registry %}
+                  {# Registry #}
```
wardi, line 60: {# I WAS HERE #}

ckanext/canada/templates/user/perform_reset.html
```diff
@@ -0,0 +1,44 @@
...
+        <section class="alert alert-info">
+          <p>
+            {{ h.render_markdown(_(
+                '1. Passwords must contain characters from at least 3 of the following 4 classes:\n'
+                '    * Upper Case Letters: A, B, C, … Z\n'
+                '    * Lower Case Letters: a, b, c, … z\n'
+                '    * Westernised Arabic Numerals: 0, 1, 2, … 9\n'
+                '    * Special characters: ({}[],.<>;:\'"?/|\\`~!@#$%%^&*()_-+=).\n'
+                '2. No accented characters are permitted.\n'
+                '3. All passwords must be at least 8 characters long.\n')) }}
```
wardi, line 17: One day we implement the modern password standards I hope...

ckanext/canada/templates/header.html
```diff
@@ -0,0 +1,188 @@
...
+                    <i class="fa fa-search" aria-hidden="true"></i><span class="wb-inv">{{ _('Search') }}</span></button>
+              </div>
+            </form>
+          </section>
+        {% endif %}
+      {% endblock %}
+      {% block user_logged %}
+        {% if g.is_registry %}
+          {# Registry #}
+          {% if c.userobj %}
```
wardi, line 44: while making changes you could update the `c.`'s to `g.`'s, though I assume the `c.` compatibility code will be with ckan for a while yet

ckanext/canada/templates/error_document_template.html
JVickery-TBS: I think that is only if you have the dev-requirements installed. Otherwise it outputs like a small accordion section. I think we can just safely remove this from our templates, we should not even output it in prod servers. I think we can all read server logs.

ckanext/canada/templates/home/index.html
JVickery-TBS: No it does not, we have this logic in the views now I think for:
```
@canada_views.route('/', methods=['GET'])
def home():
```
and
```
@canada_views.route('/links', methods=['GET'])
def links():
```
so we can remove that for sure.

ckanext/canada/templates/help.html
JVickery-TBS: sadly this does not work :(

https://stackoverflow.com/questions/22655031/jinja2-how-to-put-a-block-in-an-if-statement#answer-22667404 Confirmed the answer here.

ckanext/canada/templates/user/perform_reset.html
JVickery-TBS: I think there was some requirements stated for having the Registry go public. But I think the requirements that were given are just these ones. :/

ckanext/canada/templates/admin/index.html
```diff
@@ -1,21 +1,17 @@
 {% ckan_extends %}
 
 {% block primary_content_inner %}
-  {% if h.is_registry() %}
-    {{ super() }}
-  {% endif %}
+  {{ super() }}
 {% endblock %}
```
wardi, line 5: unconditionally calling `{{ super() }}` means this whole block can be removed instead.

ckanext/canada/templates/admin/snippets/publish_package_list.html
```diff
@@ -0,0 +1,34 @@
...
+    </ul>
+
+    <fieldset class="admin-publish-date-fieldset mrgn-tp-md">
+      <div>
+        <label for="publish_date" class="admin-publish-date-label">
+          {% if h.lang() == 'en' %}
+            Publish Date (<abbr title="Four digits year">YYYY</abbr>-<abbr title="Two digits month">MM</abbr>-<abbr title="Two digits day">DD</abbr>):
+          {% else %}
+            Date de publication <abbr title="Quatre chiffres de l'année">(AAAA</abbr>-<abbr title="Deux chiffres du mois">MM</abbr>-<abbr title="Deux chiffres du jour">JJ</abbr>) :
+          {% endif %}
```
wardi, line 25: `{% trans %}` is good for marking translated blocks that have tags in them

ckanext/canada/templates/error_document_template.html
```diff
@@ -2,33 +2,31 @@
...
-      {{ gettext('Error %(error_code)s', error_code=code) }}
-    {% else %}
+    {% if g.is_registry %}
       {{ _("Open Government Registry") }}
+    {% else %}
+      {% if code == 404 or code == 403 %}
+        {{ _('We couldn\'t find that Web page') }}
+      {% else %}
+        {{ _('We encountered an error') }}
+      {% endif %}
```
wardi, line 12: we don't show error text on the registry?

JVickery-TBS: @wardi yeah our error page is bad. It is better now in the CDTS stuff. Idk why we output a custom misc message for 404/403. Was probably a decision long ago.

ckanext/canada/templates/package/snippets/resource_form.html
```diff
@@ -1,12 +1,13 @@
 {% ckan_extends %}
 
 {% block basic_fields_url %}
-      {{ form.input('url', id='field-url', label=_('URL'), placeholder=_('http://whatever'), value=data.url, error=errors.name) }}
+  {{ form.input('url', id='field-url', label=_('URL'), placeholder=_('http://whatever'), value=data.url, error=errors.name) }}
```
wardi, line 4: IIRC WET makes placeholders invisible? maybe better to not have any placeholder value used?

ckanext/canada/templates/user/perform_reset.html
wardi: Check out the new password complexity standards from the link in the ticket @RabiaSajjad showed this morning. We don't need all this with the new standards.

ckanext/canada/templates/package/snippets/resource_form.html
JVickery-TBS: Yeah, depending on the form class (or field class?) the placeholder will be hidden. I think inside of these form fields, they are. So will remove this.

ckanext/canada/templates/user/perform_reset.html
JVickery-TBS: Yeah, the new standards will probably be done separately from the template merge. We will update the text here then.

ckanext/canada/templates/package/snippets/resource_form.html
JVickery-TBS: @wardi turns out the placeholders do show. But this particular template is never used for us because we scheming. So I think this is just an outdated template (which just copies stuff from core anyways). So I have deleted this template now.

ckanext/canada/templates/admin/snippets/publish_package_list.html
JVickery-TBS: This is now in PO file. and I also looked for any other places with these silly IF statements. Didn't find anymore, but was able to clean up some of the `if lang == 'en'` statements a bit.

ckanext/canada/templates/snippets/cdts/header.html
```diff
@@ -0,0 +1,259 @@
...
+          },
+        {% endif %}
+        {% if h.check_access('package_create') %}
+          {
+            "text": "{{ _('Open Data') }}",
+            "id": "open-data",
+            "subLinks": [
+              {
+                "subhref": "{{ h.url_for('dataset_new') }}",
+                "subtext": "{{ _('Create an Open Data Record') }}",
```
wardi, line 36: all of these variables that are inserting values into json will be escaped by jinja2 like they are HTML. Instead of `"{{...}}"` we should use `{{ ...|tojson }}` for correct escaping.

```suggestion
                "subhref": {{ h.url_for('dataset_new')|tojson }},
                "subtext": {{ _('Create an Open Data Record')|tojson }},
```

wardi (approved):
otherwise LGTM

ckanext/canada/templates/snippets/cdts/header.html
JVickery-TBS: Cool. @wardi this is done now for all the calls to gettext and helper methods etc in the CDTS headers.html here.

PR #1480, opened 2024-06-03
Use CDTS Templates and CDN Scripts

JVickery-TBS:
- Load the GCWeb an WET theme styles and scripts from the CDTS CDN.
- Use the wet JS builder for the CDTS templates.


ckanext/canada/templates/error_document_template.html
```diff
@@ -1,24 +1,30 @@
...
+    </div>
+    <div class="col-xs-9 col-sm-10 col-md-11">
+        <h1 class="mrgn-tp-md">
+          {% if code == 404 or code == 403 %}
+            {{ _('We couldn\'t find that Web page') }}
+          {% else %}
+            {{ _('We encountered an error') }}
+          {% endif %}
+        </h1>
+        <p class="pagetag"><b>Error {{ code }}</b></p>
```
wardi, line 18: should `Error` be translated here?

wardi (approved):

ckanext/canada/templates/error_document_template.html
JVickery-TBS: It absolutely should be haha. Thanks!

PR #1483, opened 2024-06-04
Space Before Colons for French

JVickery-TBS:
- Space before colons for inline colons in templates.


wardi (approved):
This works but I'm sure this isn't the recommended way to localize colons :thinking:

JVickery-TBS:
Mmmm good point, not sure why I did not just move the colons into the `gettext`s. Will do that and update catalogues.

wardi:
I'm sure there are cases where we want the same string translated and used as a label in some cases but not in others. Just wondering if there's something like the date/number formatting stuff built in to babel that covers colon spacing or something.

Having two versions of all the strings that might have a following colon doesn't seem like a great solution either

JVickery-TBS:
Yeah after looking online for quite some time, there does not really seem to be a way to localize punctuation like this in a perfect way. Either have two strings in PO file or do the dynamic thing I am doing here. Or just choose to use no colon/punctuation.

So I think we can just put the colons in the PO files and be "normal" in that way.

PR #1484, opened 2024-06-05
DOI Validation

JVickery-TBS:
Adds validation to the DOI field. Also, the `validators` in the schema yaml was missing a tab in, so this field has never had any validators for it (very not good).


ckanext/canada/tests/test_validators.py
```diff
@@ -514,6 +514,81 @@ def test_validation_options(self):
         resource = self.sysadmin_action.resource_create(**resource_data)
         assert 'validation_options' not in resource or resource['validation_options'] == None
 
+    def test_digital_object_identifier(self):
+        "DOI values should conform to the standar DOI practices."
```
wardi, line 518:
```suggestion
        "DOI values should conform to the standard DOI practices."
```

ckanext/canada/validators.py
```diff
@@ -31,6 +31,39 @@
...
+#                    &
+#                    ?
+#
+# Examples: 10.1000/xyz-123
+#           10.1109/5.771073
+#           10.231/JIM.0b013e31820bab4c
+#           10.1016.12.31/nature.S0735-1097(98)2000/12/31/34:7-7
+#           10.1002/(SICI)1522-2594(199911)42:5<952::AID-MRM16>3.0.CO;2-S
+#
+doi_match = re.compile(r'\b(10[.][0-9]{3,}(?:[.][0-9]+)*/(?:(?!["&\'\?])\S)+)\b')
```
wardi, line 65: This will allow text to appear after the DOI, is that intended?

JVickery-TBS: @wardi the rest of the text is caught in the validator method. So would not allow text after this regex match

wardi: Oh, I didn't see `or match.group(0) != value`. I would have spelled that `$` in the regex instead.

JVickery-TBS: Used regex `$` instead now. Works as expected.

PR #1485, opened 2024-06-06
FAQ Markdown to Jinja Template & PO

JVickery-TBS:
Moving crazy Markdown thing for the static FAQ page into a Jinja template and moving translated strings into the PO file.

The FAQ page needs a lot of content updates as it is 8 years old and a lot of the info is outdated or incorrect now.

For now, let's get this into master for a11y fixes.


ckanext/canada/templates/internal/help.html
```diff
@@ -1,7 +1,181 @@
...
+    <h3>{{ _('General Registry Information') }}</h3>
+    <details>
+      <summary>{{ _("Why can't I create an account?") }}</summary>
+      <p>{{ _('The most common reason is duplicate accounts. If you tried to create an account and you receive this error message "That login name is not available", it means you already have an account with this username, but you either forgot or you are not certain it was already created. If you remember the password, try to log into the Registry using the same username and password. If you forgot the password, refer to the Reset your Password function, and a new one will be emailed to you. For any other reasons, ask for assistance via e-mail to open-ouvert@tbs-sct.gc.ca.') }}</p>
+    </details>
+    <details>
+      <summary>{{ _('Have you forgotten your password?') }}</summary>
+      <p>{{ _('If you failed to login to the Registry and received this error massage "Login failed. Bad username or password", start by requesting a password reset following these steps:') }}</p>
+      <ol>
+        <li>{{ _('Click the "<strong>Forgotten your password?</strong>" button. The "Reset Your Password" page is displayed.') }}</li>
```
wardi, line 19: IIRC you need to use `{% trans %}` blocks instead of `_()` if you want to have HTML tags mixed with the strings

wardi: (also they let you wrap much larger blocks for translation)

PR #1486, opened 2024-06-18
Security Plugin New Code

JVickery-TBS:
Plugin changes for PR: https://github.com/open-data/ckanext-security/pull/4

Also adds in some string we need to translate in this PR (translations to come).

### DIFF
--- a/ckanext/canada/plugins.py
+++ b/ckanext/canada/plugins.py
@@ -70,6 +70,7 @@ class CanadaSecurityPlugin(CkanSecurityPlugin):
     p.implements(p.IConfigurer)
 
     def update_config(self, config):
+        super(CanadaSecurityPlugin, self).update_config(config)
         # Disable auth settings
         config['ckan.auth.anon_create_dataset'] = False
         config['ckan.auth.create_unowned_dataset'] = False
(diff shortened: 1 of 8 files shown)


JVickery-TBS:
requires: https://github.com/open-data/ckanext-security/pull/4

ckanext/canada/plugins.py
```diff
@@ -97,10 +98,12 @@ def before_update(self, context, current, resource):
...
-        return {'canada_security_upload_type':
-                    validators.canada_security_upload_type,
-                'canada_security_upload_presence':
-                    validators.canada_security_upload_presence}
+        validators_dict = super(CanadaSecurityPlugin, self).get_validators() or {}
+        validators_dict.update({'canada_security_upload_type':
+                                    validators.canada_security_upload_type,
+                               'canada_security_upload_presence':
+                                    validators.canada_security_upload_presence})
+        return validators_dict
```
wardi, line 106: doesn't matter in this case, but for when it does I like to avoid mutating dicts as a best-practice.
```suggestion
        return dict(
            validators_dict,
            canada_security_upload_type=validators.canada_security_upload_type,
            canada_security_upload_presence=validators.canada_security_upload_presence,
        )
```

wardi (approved):

JVickery-TBS:
Waiting on translations for this, will get them on Monday.

PR #1487, opened 2024-06-25
minor updates to faq

RabiaSajjad:


ckanext/canada/templates/internal/help.html
```diff
@@ -71,7 +70,6 @@ <h3>{{ _('Proactive Disclosure Information') }}</h3>
...
         <li>{{ _('Prepare Data in the Proper Template: Organize your data in the appropriate template for the specific organization and Proactive publication type. Each template corresponds to particular context, so choose the relevant one.') }}</li>
-        <li>{{ _('Maintain the Abbreviated Organization Name: In the template, the abbreviated name of the organization (e.g., "aandc-aadnc") must remain in cell C1 exactly as downloaded.') }}</li>
         <li>{{ _("Include All Required Fields: Ensure that the template you're uploading contains all the required fields for the specific Proactive publication type. Missing fields can lead to issues during the upload process.") }}</li>
         <li>{{ _('Enter Valid Data: Populate the template with valid and accurate data. Incorrect or incomplete information may cause errors.') }}</li>
         <li>{{ _('File Format and Naming: Your completed data template should be an Excel spreadsheet in either XLS or XLSX format. Do not rename the file, as the system relies on a unique ID associated with each template. Renaming it could lead to complications.') }}</li>
```
JVickery-TBS, line 75: File naming does not matter anymore (which I noted in my review of this content), but if you did not get updated text for this, then we can just ignore and go ahead I guess...

JVickery-TBS (commented):
Also noted `or use the facets on the right-hand side` -> `left-hand side` for the "How Do I Search the Registry?" question. But again, if you did not get update copy for that let's just continue at this point?

@RabiaSajjad Looks good. Are you able to commit the `pot` file as well from the `extract_message` as described here: https://github.com/open-data/ckanext-canada?tab=readme-ov-file#localization-strings

JVickery-TBS:
@RabiaSajjad also don't worry about the 4 tests failing here, that is solved by the PR for the password stuff.

PR #1489, opened 2024-07-02
Trim Text Fields for PD Record Solr Indices

JVickery-TBS:
fix(cli): solr index max length;

- Max length for PD records into solr for text fields.


thriuin (approved):
Perfect

PR #1492, opened 2024-07-11
fix bug for validation report not loading

RabiaSajjad:


JVickery-TBS (approved):
@RabiaSajjad looks good, I imagine this should solve the issue in prod.

I think there is also a JS error that also happens with the `report.js` from `ckanext-validation`. https://github.com/open-data/ckanext-validation/blob/canada-py3/ckanext/validation/webassets/webassets.yml So we might need to do this there as well?

RabiaSajjad:
> @RabiaSajjad looks good, I imagine this should solve the issue in prod.
> 
> I think there is also a JS error that also happens with the `report.js` from `ckanext-validation`. https://github.com/open-data/ckanext-validation/blob/canada-py3/ckanext/validation/webassets/webassets.yml So we might need to do this there as well?

Committed the same change to ckanext-validation https://github.com/open-data/ckanext-validation/commit/e94671fdd0c69f21a9e07ad8c476b78144b254be

PR #1495, opened 2024-07-18
integrate survey pop-up using GC invitation-manager

RabiaSajjad:
Source: https://github.com/ServiceCanada/invitation-manager


JVickery-TBS (commented):
@RabiaSajjad just wanted to confirm if we needed the IM on JUST "dataset pages" or also the resource pages?

Also wanted to confirm if we need IM for the Registry as well? If not, then in the templates you can use `g.is_registry`

ckanext/canada/assets/invitation-manager/InvitationManager.js
JVickery-TBS, line 1: @RabiaSajjad could you actually put these scripts into `invitation-manager/vendor/` as these are not our JS files. And then just update the `webassets.yml` outputs to include `vendor/`

JVickery-TBS (changes requested):

RabiaSajjad:
> @RabiaSajjad just wanted to confirm if we needed the IM on JUST "dataset pages" or also the resource pages?
> 
> Also wanted to confirm if we need IM for the Registry as well? If not, then in the templates you can use `g.is_registry`

We only need the survey pop-up on Portal dataset pages.

PR #1496, opened 2024-07-23
Round Money DS Types in Triggers

JVickery-TBS:
The money psql type is not rounding to 2 decimal places. Use DB triggers to round before inserting into tables.

### DIFF
--- a/changes/1496.changes
+++ b/changes/1496.changes
@@ -1,0 +1,1 @@
+Money fields for PD types now get rounded to 2 decimals when inserting into the database.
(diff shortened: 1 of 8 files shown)


JVickery-TBS:
Got some basic rounding tests, and some tests for nullish values. E.g. `None` or empty string on non required fields should stay null.

wardi:
note that while we mark these fields as "money" recombinant is converting that to "numeric" for postgres because the postgres "money" type inserts `$` in the output making the values strings instead of numbers

wardi (approved):

PR #1498, opened 2024-07-29
add feature to delete selected pd records

RabiaSajjad:


ckanext/canada/view.py
```diff
@@ -579,6 +580,72 @@ def type_redirect(resource_name, dataset_id=None):
                                    owner_org=owner_org_name))
 
 
+@canada_views.route('/recombinant/delete-selected-records/<resource_id>', methods=['GET', 'POST'])
+def delete_selected_records(resource_id):
+    lc = LocalCKAN(username=g.user)
+
+    if request.method != 'POST' or not h.check_access('datastore_records_delete',
```
JVickery-TBS, line 587: You will want to move the `request.method` condition to below. Condition the `request.method == 'POST'` along with the `'cancel' in request.form` stuff. E.g. `if request.method == 'POST' and 'confirm' in request.form ...then do the actual delete`

ckanext/canada/view.py
```diff
@@ -579,6 +580,72 @@ def type_redirect(resource_name, dataset_id=None):
...
+        abort(403, _('User {0} not authorized to update resource {1}'
+                     .format(str(g.user), resource_id)))
+
+    try:
+        res = lc.action.resource_show(id=resource_id)
+        pkg = lc.action.package_show(id=res['package_id'])
+        org = lc.action.organization_show(id=pkg['owner_org'])
+        dataset = lc.action.recombinant_show(
+            dataset_type=pkg['type'], owner_org=org['name'])
+    except NotFound:
```
JVickery-TBS, line 598: We should also catch `NotAuthorized` here as well just in case so we can avoid any 500 errors. CKAN and Ian tend to like to return 404's for NotAuthorized errors instead of 403s. So you can just do `except (NotFound, NotAuthorizaed)`

ckanext/canada/templates/package/wet_datatable.html
```diff
@@ -99,6 +99,40 @@ <h2 id="prtable">{{_("Preview")}}</h2>
...
+              name: 'resource_name'
+            }));
+            form.append($('<input>').attr({
+              type: 'hidden',
+              value: "{{request.cookies.token}}",
+              name: 'token'
+            }));
+            form.append($('<input>').attr({
+              type: 'hidden',
+              value: {{primary_keys}},
```
JVickery-TBS, line 121: Add `| tojson` filter for the `primary_keys`, which solves any encoding issues with Jinja2

JVickery-TBS (changes requested):
@RabiaSajjad sweet looks pretty good, basically just a modified version of Recombinant's `records_delete`. So if Ian has any feedback we can just blame him for his `records_delete` from Recombinant.

Just the minor above comments.

Did you end up doing it by the `_id` field? Or were you able to just do it by the primary keys? and we are going to fix that validation on the commas in datastore primary keys in that other PR?

RabiaSajjad:
> @RabiaSajjad sweet looks pretty good, basically just a modified version of Recombinant's `records_delete`. So if Ian has any feedback we can just blame him for his `records_delete` from Recombinant.
> 
> Just the minor above comments.
> 
> Did you end up doing it by the `_id` field? Or were you able to just do it by the primary keys? and we are going to fix that validation on the commas in datastore primary keys in that other PR?

@JVickery-TBS it is done using primary keys. I wasn't able to access the `_id` field since `get_chromo` returns fields from the yaml file.

ckanext/canada/view.py
```diff
@@ -579,6 +580,72 @@ def type_redirect(resource_name, dataset_id=None):
...
+        pkg = lc.action.package_show(id=res['package_id'])
+        org = lc.action.organization_show(id=pkg['owner_org'])
+        dataset = lc.action.recombinant_show(
+            dataset_type=pkg['type'], owner_org=org['name'])
+    except (NotFound, NotAuthorized):
+        abort(404, _('Not found'))
+
+    records = request.form.getlist('select-delete')
+
+    if request.method != 'POST' or 'cancel' in request.form:
```
JVickery-TBS, line 603: @RabiaSajjad sorry, the condition should go in the below if, something like `if request.method == 'GET` or 'confirm' not in request.form:`

That way in CKAN 2.10, the GET will be handled correctly for rendering the confirm delete template.

JVickery-TBS (approved):
Looks good

JVickery-TBS:
@RabiaSajjad I threw in a quick example of showing the primary keys in a table like DS Audit shows. Of course it will only show the primary keys as we do not have any of the other data in the request, so up to you if you want to show more with some datastore search thing

JVickery-TBS:
@RabiaSajjad This one is hot to go

PR #1501, opened 2024-08-09
Translate errors for triggers + fix incorrect translation for resource type

RabiaSajjad:


ckanext/canada/triggers.py
```diff
@@ -18,7 +18,7 @@ def update_triggers():
         definition=u'''
             BEGIN
                 IF (value = '') IS NOT FALSE THEN
-                    RETURN ARRAY[[field_name, 'This field must not be empty']];
+                    RETURN ARRAY[[field_name, _('This field must not be empty')]];
```
JVickery-TBS, line 21: @RabiaSajjad we cannot use the gettext in here as these are just literals that go the DB functions. So this would create the DB trigger with `_()` which does not exist in PSQL.

Not too sure how to accomplish getting these through gettext. Will take a gander after lunch

JVickery-TBS (approved):
@RabiaSajjad this one is good to go too

PR #1503, opened 2024-08-16
2024 Service Inventory

JVickery-TBS:
Need to figure out a couple of things for this still. But almost at a point to test out things one field at a time to make sure we got everything.


ckanext/canada/triggers.py
```diff
@@ -110,12 +125,39 @@ def update_triggers():
...
+            {u'argname': u'error', u'argtype': u'_text', u'argmode': u'out'}],
+        rettype=u'record',
+        definition=u'''
+            DECLARE
+                value_trimmed text := trim(both E'\t\n\x0b\x0c\r ' from value);
+            BEGIN
+                IF value IS NOT NULL AND value <> '' AND LENGTH(value_trimmed) > max_chars THEN
+                    error := ARRAY[[field_name, 'This field has a maximum length of {} characters.\uF8FF' || max_chars]];
+                END IF;
+                IF value IS NULL OR value = '' THEN
```
wardi, line 154: empty strings get converted to NULL but all-whitespace strings get trimmed to an empty string? that's not idempotent

ckanext/canada/triggers.py
```diff
@@ -110,12 +125,39 @@ def update_triggers():
...
+                    clean := NULL;
+                ELSE
+                    clean := ARRAY(SELECT value_trimmed);
```
wardi, line 157: I don't understand why this is needed

ckanext/canada/triggers.py
JVickery-TBS: what should happen with empty strings? just allow them?

ckanext/canada/triggers.py
JVickery-TBS: the null part? or the trigger in general haha?

Will `value_trimmed` be null/empty string and we can just do the `clean := ARRAY(SELECT value_trimmed)`

ckanext/canada/triggers.py
wardi: I think it's better to always store strings in string values (i.e. empty string instead of a mix of empty strings and NULL values) because there's no useful distinction.

ckanext/canada/triggers.py
wardi: I mean the `ARRAY(SELECT value_trimmed)`, why can't we use `value_trimmed` directly?

wardi: Also it might be simpler to make the trimming and error generation separate functions that are called from the trigger one after the next.

ckanext/canada/tables/service.yaml
```diff
@@ -74,866 +98,1156 @@ resources:
...
+      en: |
+        This field must not be empty.
+        This field has a maximum length of 1800 characters.
+      fr: |
+        Cette zone ne doit pas être vide.
+        Ce champ ne peut excéder une longueur maximale de 1800 caractères.
+    occurrence: Single
+    datastore_type: text
+    max_chars: 1800
+    excel_error_formula: 'OR({default_formula},LEN(TRIM({cell}))>1800)'
```
wardi, line 170: this part is repeated enough that it might be better as a new `max_chars` feature in recombinant's excel error formula generation

ckanext/canada/triggers.py
JVickery-TBS: move trimmed to its own function. update yaml now.

ckanext/canada/tables/service.yaml
JVickery-TBS: got this in recombinant, just need to update all places in YAML. And then test

ckanext/canada/cli.py
```diff
@@ -1873,3 +1873,30 @@ def iter_records():
...
+@click.option('-v', '--verbose', is_flag=True, type=click.BOOL, help='Increase verbosity.')
+def delete_old_triggers(verbose=False):
+    """
+    Delete old, unused database triggers.
+    """
+    _drop_function('no_surrounding_whitespace_error', verbose)
+    _drop_function('year_optional_month_day_error', verbose)
+    _drop_function('choices_from', verbose)
+    _drop_function('integer_or_na_nd_error', verbose)
+    _drop_function('inventory_trigger', verbose)
```
wardi, line 1888: https://docs.ckan.org/en/2.10/maintaining/datastore.html#ckanext.datastore.logic.action.datastore_function_delete

ckanext/canada/tables/choices/boolean.yaml
```diff
@@ -0,0 +1,6 @@
+'N':
```
wardi, line 1: I might name this `yes_no.yaml` because to me (and to postgres) boolean means true/false

ckanext/canada/cli.py
JVickery-TBS: @wardi because of our addition to the argmode and extra args stuff, the `datastore_delete` CKAN action did not work for me. So just did it in this crappy way for now

ckanext/canada/tables/choices/boolean.yaml
JVickery-TBS: Yeah makes sense, will update the "online_service_boolean.yaml" to "online_service_yes_no.yaml" as well.

PR #1505, opened 2024-08-20
add title to truncated links

RabiaSajjad:


ckanext/canada/templates/package/base.html
```diff
@@ -7,7 +7,6 @@
   {% set pkg_url = h.url_for(pkg.type ~ '.read', id=pkg.id if is_activity_archive else pkg.name, **({'activity_id': request.args['activity_id']} if 'activity_id' in request.args else {})) %}
   {% if g.action != 'new' and pkg %}
     {% set dataset = h.get_translated(pkg, 'title') %}
-    <li><a href="{{ pkg_url }}">{{ dataset|truncate(30) }}</a></li>
+    <li><a href="{{ pkg_url }}" title="{{ dataset }}">{{ dataset|truncate(30) }}</a></li>
```
JVickery-TBS, line 10: @RabiaSajjad increase the `truncate` part here to more than `30`? or nah?

ckanext/canada/helpers.py
```diff
@@ -861,6 +861,7 @@ def ckan_to_cdts_breadcrumbs(breadcrumb_content):
         anchor = breadcrumb.find('a')
         cdts_breadcrumbs.append({
             'title': breadcrumb.text if not anchor else anchor.text,
+            'acronym': '' if not anchor else anchor.get('title'),
```
JVickery-TBS, line 864: @RabiaSajjad cannot remember, I think I had problems with CDTS and the Jinja JSON filter with None/null values. Add empty string as default value for title?

```
anchor.get('title', '')
```

ckanext/canada/templates/admin/trash.html
```diff
@@ -8,7 +8,7 @@
       <ul class="user-list list-group">
         {% for pkg in g.deleted_packages %}
           {% set title = pkg.title or pkg.name %}
-          <li class="list-group-item">{{ h.link_to(h.truncate(title, truncate_title), h.url_for(pkg.type + '.read', id=pkg.name)) }}</li>
+          <li class="list-group-item">{{ h.link_to(h.truncate(title, truncate_title), h.url_for(pkg.type + '.read', id=pkg.name), title=title) }}</li>
```
JVickery-TBS, line 11: What is the value of `truncate_title` in this template, and does it need to be increased?

ckanext/canada/templates/package/resource_read.html
```diff
@@ -31,6 +31,11 @@
   {% endif %}
 {% endblock %}
 
+{% block breadcrumb_content %}
+  {{ super.super() }}
+  <li class="active"><a href="" title="{{ h.resource_display_name(res) }}">{{ h.resource_display_name(res)|truncate(30) }}</a></li>
```
JVickery-TBS, line 36: I think the Resource names get pretty long, but we might still be able to support more than `30` chars here?

JVickery-TBS (changes requested):

ckanext/canada/templates/package/base.html
RabiaSajjad: I can change it 80 so that it's the same everywhere.

ckanext/canada/helpers.py
RabiaSajjad: I had no issues in testing, but I'll default to empty string in case I have missed anything.

ckanext/canada/templates/package/resource_read.html
RabiaSajjad: I'll do 80 here as well.

ckanext/canada/templates/package/resource_read.html
```diff
@@ -31,6 +31,11 @@
   {% endif %}
 {% endblock %}
 
+{% block breadcrumb_content %}
+  {{ super.super() }}
+  <li class="active"><a href="" title="{{ h.resource_display_name(res) }}">{{ h.resource_display_name(res)|truncate(80) }}</a></li>
```
wardi, line 36: AFAIK `<a>` tags without an href are bad for a11y. Are we allowed to add a title to another type of tag?

ckanext/canada/templates/snippets/cdts/header.html
```diff
@@ -109,10 +109,6 @@
                 "subhref": {{ recombinant_menu_link('travela') | tojson }},
                 "subtext": {{ recombinant_menu_text('travela') | tojson }},
               },
-              {
-                "subhref": {{ recombinant_menu_link('qpnotes') | tojson }},
-                "subtext": {{ recombinant_menu_text('qpnotes') | tojson }},
-              },
```
wardi, line 115: unrelated change?

RabiaSajjad: yes, this is just removing a double entry of qpnotes in the menu

ckanext/canada/templates/package/resource_read.html
JVickery-TBS: I think you can do javascript void into the href to fix the a11y complaints.

JVickery-TBS: Also put `tabindex="-1"` to prevent tabbing

JVickery-TBS (approved):

wardi (approved):

PR #1508, opened 2024-08-22
remove the URL from resource page

RabiaSajjad:


thriuin:
Pretty straightforward. Go for it!

PR #1510, opened 2024-08-23
add filters and update makefile for SI

RabiaSajjad:


bin/filter/filter_service.py
```diff
@@ -21,15 +27,48 @@ def main():
...
-        except ValueError:
-            pass
+        for rem in REMOVE_COLUMNS:
+            del row[rem]
+
+        # calculate total number of applications
+        row['num_applications_total'] = 0
+        num_fields = [ 'num_phone_enquiries',
+                       'num_applications_by_phone',
+                       'num_website_visits',
```
JVickery-TBS, line 42: Same as `num_phone_enquiries`, I don't think we need `num_website_visits` as we are just calculating the `num_applications_total`

bin/filter/filter_service_std.py
```diff
@@ -21,35 +22,44 @@ def main():
...
-            num, den = num + int(row['q3_performance_result']), den + int(row['q3_business_volume'])
-        except ValueError:
-            pass
+
+        num = int(row['volume_meeting_target']) if row['volume_meeting_target'] else 0
+        den = int(row['total_volume']) if row['total_volume'] else 0
+
+        # performance = volume_meeting_target / total_volume
+        if num <= 0 or den <=0:
+            row['performance'] = 0
```
JVickery-TBS, line 38: `Performance` in their DEP is listed as "Optional", as are the `volume_meeting_target` and `total_volume` with the descriptions that no value inputted specifies that no data is available.

So I think `performance` should be the same, if `volume_meeting_target` or `total_volume`  <=0 then `row['performance'] = None` so that the value is null in the DS.

bin/filter/filter_service_std.py
```diff
@@ -21,35 +22,44 @@ def main():
...
+        else:
+            row['performance'] = num / den
+
+        # negative performance is not possible
+        if row['performance'] < 0:
+            row['performance'] = 0
+
         try:
-            num, den = num + int(row['q4_performance_result']), den + int(row['q4_business_volume'])
+            target = float(row['target'])
```
JVickery-TBS, line 47: Similar to `Performance`, the `target_met` is Optional in their DEP. So basically if the target is None, then the `target_met` also is None.

Otherwise, we can do the calculation you have.

bin/filter/filter_service_std.py
```diff
@@ -21,35 +22,44 @@ def main():
...
         except ValueError:
-            pass
-        if not den:
-            row['performance'] = 'ND'
+            target = 0
+
+        # if no total_volume then target_met is not possible
+        if den <= 0:
+            row['target_met'] = -1
```
JVickery-TBS, line 53: Sorry, we changed the control list to `NA`, `N`, `Y`,

so update `-1` to `"NA"`

bin/filter/filter_service_std.py
```diff
@@ -21,35 +22,44 @@ def main():
...
+
+        # if performance >= target then target is met
+        elif row['performance'] >= target:
+            row['target_met'] = 1
```
JVickery-TBS, line 57: update `1` to `"Y"`

bin/filter/filter_service_std.py
```diff
@@ -21,35 +22,44 @@ def main():
...
+
+        # otherwise target_met is not met
         else:
-            row['performance'] = '%0.5f' % (float(num) / den)
+            row['target_met'] = 0
```
JVickery-TBS, line 61: update `0` to `"N"`

JVickery-TBS (approved):

JVickery-TBS:
@RabiaSajjad as discussed, we can see if we can get the order of fields.

For service:
 - `program_name_en` and `program_name_fr` can go right after `program_id` column
 - `num_applications_total` can go right after the last type of those fields: `num_applications_by_other`

For service standard:
 - `performance` can go after `total_volume` (its like that in their DEP)
 - `target_met` can go after `comments_fr` (its like that in their DEP)

JVickery-TBS:
@RabiaSajjad amazing! thanks so much. I think this is good? unless if we want @wardi  to review it if he has time?

bin/filter/filter_service_std.py
```diff
@@ -20,36 +32,44 @@ def main():
...
+        # performance = volume_meeting_target / total_volume
+        if num <= 0 or den <=0:
+            row['performance'] = None
         else:
-            row['performance'] = '%0.5f' % (float(num) / den)
+            row['performance'] = num / den
+
+        # negative performance is not possible
+        if row['performance'] and row['performance'] < 0:
+            row['performance'] = 0
```
wardi, line 53:
```suggestion
            row['performance'] = max(num / den, 0)
```

wardi:
Don't know how large the data is, but these filter scripts would be good candidates for using pandas or similar if there's interest in speeding them up

PR #1511, opened 2024-08-30
nap5: enforce validation on status, progress fields

wardi:
OPEN-3581 points out missing progress_fr checks, but they're also missing on status and progress_en


thriuin:
@RabiaSajjad  I recommend merging this PR. Can you please handle during my absence? Thanks

PR #1513, opened 2024-09-10
add alert for users to change password on login page

RabiaSajjad:


ckanext/canada/templates/user/edit_user_form.html
```diff
@@ -16,7 +16,12 @@
     {% endif %}
   </fieldset>
   <fieldset>
-    <legend>{{ _('Change password') }}</legend>
+      <legend class="h2">{{ _('Change password') }}</legend>
+      <legend>
+        <section class="alert alert-info">
+          <p>{% trans %}password reset requirements{% endtrans %}</p>
```
wardi, line 22: render these as markdown?

PR #1515, opened 2024-09-27
leave spaces in datatables column labels

wardi:
missing_spaces


JVickery-TBS (approved):
@wardi looks like the Recombinant template is fine (no &nbsp stuff for the labels).

There is `assets/datatables/canada_datatables.css` if you need to put the `text-wrap: nowrap;` for the headers. Unless if it is just working fine without it?

Also add changelog file for our patch notes?

wardi:
added changes, yes it seems to work without the css change (long titles are hidden past the right edge of the cell)

PR #1516, opened 2024-09-27
Do not exit in Post to Portal

JVickery-TBS:
fix(cli): post to portal no flop;

- Do not exit on ValidationErrors.

### DIFF
--- a/ckanext/canada/cli.py
+++ b/ckanext/canada/cli.py
@@ -355,17 +355,43 @@ def _copy_datasets(source, user, mirror=False, verbose=False):
             elif target_deleted:
                 action = 'updated'
                 reason = 'undeleting on target'
-                portal.action.package_update(**source_pkg)
-                for r in source_pkg['resources']:
-                    target_hash[r['id']] = r.get('hash')
-                action += _add_datastore_and_views(source_pkg, portal, target_hash, source, verbose=verbose)
+                try:
+                    portal.action.package_update(**source_pkg)
+                except ValidationError as e:
+                    #TODO: flag source package with failure
+                    #TODO: report errors to sysadmins??
+                    action += '\n package-update-failed for ' + package_id + ' ' + str(e)
+                    if verbose:
+                        action += '\n    Failed with ValidationError: %s' % traceback.format_exc()
+                    pass
+                else:
+                    for r in source_pkg['resources']:
+                        target_hash[r['id']] = r.get('hash')
+                    action += _add_datastore_and_views(source_pkg, portal, target_hash, source, verbose=verbose)
             elif target_pkg is None:
                 action = 'created'
-                portal.action.package_create(**source_pkg)
-                action += _add_datastore_and_views(source_pkg, portal, target_hash, source, verbose=verbose)
+                try:
+                    portal.action.package_create(**source_pkg)
+                except ValidationError as e:
+                    #TODO: flag source package with failure
+                    #TODO: report errors to sysadmins??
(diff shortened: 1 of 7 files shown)


wardi (approved):

ckanext/canada/model.py
```diff
@@ -0,0 +1,83 @@
...
+        if not package_id:
+            return None
+
+        q = meta.Session.query(self).autoflush(True).filter_by(package_id=package_id)
+        if for_update:
+            q = q.with_for_update()
+        return q.first()
+
+    @classmethod
+    def upsert(self, package_id, reason_code=None, error=None):
```
wardi, line 38: by convention
```suggestion
    def upsert(cls, package_id, reason_code=None, error=None):
```

ckanext/canada/model.py
```diff
@@ -0,0 +1,83 @@
...
+    __tablename__ = 'package_sync'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    package_id = Column(Unicode)
+    created = Column(DateTime, default=datetime.datetime.now())
+    reason_code = Column(Unicode)
+    error = Column(Unicode)
+
+    @classmethod
+    def get(self, package_id, for_update=False):
```
wardi, line 27: `cls` here too

ckanext/canada/model.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @classmethod
+    def upsert(self, package_id, reason_code=None, error=None):
+        '''Sets and returns a package_sync object referenced by its package_id.'''
+        package_sync = self.get(package_id, for_update=True)
+
+        if package_sync:
+            package_sync.reason_code = reason_code
+            package_sync.error = error
+            package_sync.created = datetime.datetime.now()
```
wardi, line 45: datetime.now() is a naive local datetime, better to use utc or a timezone-aware datetime

ckanext/canada/schemas/dataset.yaml
```diff
@@ -56,6 +56,25 @@ sidebar_show_fields:
...
+
+- field_name: out_of_sync
+  label:
+    en: Out of Sync with Portal
+    fr: FR Out of Sync with Portal FR
+  form_snippet: null
+  display_snippet: null
+  validators: ignore_missing ignore_not_sysadmin boolean_validator
+  #FIXME: output_validators not working...
+  output_validators: output_boolean_validator
```
wardi, line 76: I wouldn't duplicate these fields in the dataset schema because updating dataset metadata is really slow. Can't you query them from your new table when you need them instead?

ckanext/canada/cli.py
```diff
@@ -1947,3 +2036,76 @@ def _drop_function(name, verbose=False):
...
+                    if verbose:
+                        click.echo('Package %s does not exist on Registry. Skipping...' % _package['id'])
+                    pass
+        else:
+            results = False
+    click.echo('Done!')
+
+
+@canada.command(short_help="Reinitalize database tables.")
+def init_db():
```
wardi, line 2100: https://docs.ckan.org/en/2.10/extensions/best-practices.html?highlight=migration#use-migrations-when-introducing-new-models

JVickery-TBS:
Because the flagging in this cannot really be done as it is inside of the Portal global app context at that point, going to close this one in favour of: https://github.com/open-data/ckanext-canada/pull/1519

The Harvest extension will give us more useful features like UIs and emailing failures etc. Post to Portal is a main feature of our workflow and we have nothing that alerts us of failures, and no great way to debug. We just have one log file which gets truncated every 10-15 minutes.

PR #1519, opened 2024-10-08
PortalSync Harvester (ckanext-harvest)

JVickery-TBS:
@wardi this implements ckanext-harvest as a subclassed plugin (to tighten up some security things with views and for some translation stuff to come).

This is mainly for the really good reports and database stuff from ckanext-harvest. The idea was to just rework the old CLI code for post-to-portal and copy-datasets into a Harvester class for the harvest plugin.

This way we get to see what failed, what didn't need updating, what got deleted, and what is getting created/updated. All of this available in sysadmin views. There is a command to clear out these reports, which I imagine we will do a 30 day type thing and set that to a cron execution. So the DB won't get bogged with these "post to portal" reports.

Let me know what your first thoughts are for it.


JVickery-TBS:
I think my big issue here is the switching between the Registry and Portal application stacks, because even switching to the Portal one in the Fetch and Import stages, we need to save object errors in there, and as it is currently, those object errors would be saved into the Portal database, which is absolutely not want I want it to do.

So I am beginning to understand why the original CLI command did the subprocess with the data piping into the Portal ckan commands. So I think that might be the only way to actually accomplish this still. I will try to pipe the data into a `ckanapi action <action_name> @<pipe_json_data> -c <portal.ini>`.

Because the harvest plugin does one source object at a time in the Fetch and Import stages, I don't think there is a good way to be able to pipe all of the gathered datasets into one ckanapi command.  The only way would be to have one source object to store ALL the gathered packages, but that would prevent any non-broken packages from getting synced, which is really what we are trying to get past right now.

So this needs some work still for sure, just to figure out how to get the stuff into the Portal in the Fetch and Import stages whilst still being in the Registry app context.

ckanext/canada/harvesters.py
```diff
@@ -0,0 +1,616 @@
...
+                           and_(HarvestObject.harvest_job_id == HarvestJob.id,
+                                HarvestObject.current == False,  # noqa: E712
+                                HarvestObject.report_status != 'not modified'))
+                .options(contains_eager(HarvestJob.objects))
+                .order_by(HarvestJob.gather_started.desc()))
+        # now check them until we find one with no fetch/import errors
+        # if objects count is 0, job was error free
+        for job in jobs:
+            if len(job.objects) == 0:
+                return job
```
wardi, line 126: Do this part in the SQL so that you don't have postgres sending you all the records to iterate over in python, which could be really slow.

ckanext/canada/harvesters.py
```diff
@@ -0,0 +1,616 @@
...
+        log.info('Searching for activity since: %s UTC', get_changes_since)
+
+        data = self.registry.action.changed_packages_activity_timestamp_since(since_time=last_time.isoformat())
+        if not data:
+            # There is no activity since last successful run
+            return []
+
+        registry_packages = []
+
+        for result in data:
```
wardi, line 194: the same package id will often appear many times in a row when updates are done, better to get a set of package ids to eliminate duplicate work like the old script does

ckanext/canada/harvesters.py
```diff
@@ -0,0 +1,616 @@
...
+        log.info('Last error-free job: %r', last_error_free_job)
+        last_time = last_error_free_job.gather_started
+
+        # Note: SOLR works in UTC, and gather_started is also UTC, so
+        # this should work as long as local and remote clocks are
+        # relatively accurate. Going back a little earlier, just in case.
+        get_changes_since = (last_time - datetime.timedelta(hours=1)).isoformat()
+        log.info('Searching for activity since: %s UTC', get_changes_since)
+
+        data = self.registry.action.changed_packages_activity_timestamp_since(since_time=last_time.isoformat())
```
wardi, line 187: this call needs to be paginated if you want to get all records

ckanext/canada/harvesters.py
```diff
@@ -0,0 +1,616 @@
...
+                    # Portal packages are public, do this to match
+                    source_package['private'] = False
+
+            log.info('Creating HarvestObject for package %s', registry_package['id'])
+            obj = HarvestObject(guid=registry_package['id'],
+                                harvest_source_id=harvest_job.source.id,
+                                harvest_job_id=harvest_job.id,
+                                job=harvest_job,
+                                package_id=registry_package['id'],
+                                content=json.dumps(source_package) if source_package else None)
```
wardi, line 238: It seems weird to be getting storing the data in the gather stage and not the fetch stage. Also, copying this data into a new table just because the harvester uses separate stages seems like extra work for no benefit.

ckanext/canada/harvesters.py
```diff
@@ -0,0 +1,616 @@
...
+        portal_config.read(portal_ini)
+        try:
+            portal_datastore_url = portal_config.get('app:main', 'ckan.datastore.write_url')
+        except NoOptionError:
+            raise Exception('Cannot find ckan.datastore.write_url in app:main section of Portal INI')
+        self.portal_datastore_url = str(_get_engine_from_url(portal_datastore_url).url)  # ensure that this is actually a DS database
+
+        #FIXME: figure out how to get the data from and to the Portal in fetch_stage and import_stage
+        self.registry = LocalCKAN()
+        self.portal = LocalCKAN()
```
wardi, line 102: I don't think this is possible. LocalCKAN is only for the local instance of CKAN that is running, you can't have two of them with different configurations because the configuration is global.

ckanext/canada/harvesters.py
JVickery-TBS: Class method was copied from the base class of the harvest core code. Not sure how/where to add in the `.filter(HarvestJob.objects == 0)`

ckanext/canada/harvesters.py
JVickery-TBS:
```
# hard limit this api to reduce opportunity for abuse
limit = int(config.get('ckan.activity_timestamp_since_limit', 10000))
```

I would hope that we do not have more than `10000` activities in one run. But I see that the old code does paginate in an interesting way:
```
if data:
        since_time = isodate(data[-1]['timestamp'], None)
```

Will just do this here.

ckanext/canada/harvesters.py
JVickery-TBS: Yeah from the docs and code of the ckanext-harvest plugin I did not really like how it stores so much in the DB tables. We can probably do everything in the gather stage in terms of getting the Registry and Portal packages and comparing them.

I imagine that the gather_errors will still show us something at least in the UI and database.

The framework of the harvest plugin does kind of require us to actually use the import_stage though. So we would need to either store the data in the DB tables to use in the import_stage, or just go get everything again. Which would be annoying that it does everything again, but I don't think it would hurt that much. I would rather have it go get the Registry package, views, and data dictionary again than store giant objects in the database.

I think another issue is how the fetch_stage works. From what I can tell, the harvest plugin will do stuff and show us things for the return of the fetch. I guess we could save a small object in the DB from gather_stage. Not the whole objects, but some small thing like `{"state": "unchanged"} | {"state": "success"} | {"state": "failed"}`. Will check out what the column is for the object content, might not need to JSON it.

ckanext/canada/harvesters.py
JVickery-TBS: Yeah I am going to look into the threading library and see if that works for us at all.

ckanext/canada/harvesters.py
JVickery-TBS: Chaos, I have moved the HarvestObjectExtras to just the content here now. So we have the status of the fetch stage, the import action, and the resource hashes.

ckanext/canada/harvesters.py
JVickery-TBS: Even more chaos, I made a context manager in the class which initializes the Portal, effectively switching the entire app context and everything to the Portal. And then on exit, it will switch back to the Registry app context.

ckanext/canada/logic.py
```diff
@@ -529,3 +532,64 @@ def canada_datastore_run_triggers(up_func, context, data_dict):
...
+                'target': package.get('target', _extras.get('target', None)),
+                'metadata_created': package.get('metadata_created'),
+                'metadata_modified': package.get('metadata_modified'),
+                'private': package.get('private'),
+                'state': package.get('state'),
+                'status': package.get('status'),
+                'resources': [],  # required for core
+            }
+
+    return package
```
wardi, line 571: what's going on here? This won't prevent creation of harvest packages or prevent them showing up in search results.

ckanext/canada/schemas/harvest.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+#
+#   DATASET FIELDS
+#
+dataset_fields:
+
+
+- field_name: id
+  create_validators: canada_validate_generate_uuid
+      unicode name_validator package_id_does_not_exist portal_sync_id portal_sync_limit
```
wardi, line 30: If you want to prevent more than one from being created you could use a validator that always returns the same id

ckanext/canada/templates/base.html
```diff
@@ -76,4 +76,19 @@
...
+  {% if dataset_type %}
+    data-package-type="{{ dataset_type }}"
+  {% elif harvest_source %}
+    data-package-type="{{ harvest_source.get('type') }}"
+  {% elif pkg %}
+    data-package-type="{{ pkg.get('type') }}"
+  {% elif pkg_dict %}
+    data-package-type="{{ pkg_dict.get('type') }}"
+  {% elif package %}
+    data-package-type="{{ package.get('type') }}"
```
wardi, line 92: interesting, what's this used for?

ckanext/canada/cli.py
```diff
@@ -1900,3 +1163,64 @@ def _drop_function(name, verbose=False):
...
+        pass
+
+
+@canada.command(short_help="Creates harvest database tables.")
+@click.option('-f', '--refresh', is_flag=True, type=click.BOOL, help='Forces the refresh of all the source objects in the database.')
+@click.option('-q', '--quiet', is_flag=True, type=click.BOOL, help='Suppresses human inspection.')
+def init_harvest_plugin(refresh=False, quiet=False):
+    """Creates harvest database tables."""
+
+    _maybe_init_harvest_db()
```
wardi, line 1202: can we not https://docs.ckan.org/en/2.10/extensions/best-practices.html#use-migrations-when-introducing-new-models or is that just not how the harvest extension works?

ckanext/canada/harvesters.py
wardi: Yikes, and that has to be really slow if it works at all.

ckanext/canada/logic.py
JVickery-TBS: My bad, I think that was some crazy thing I was initially trying to do. But that chained action is not used so will remove.

The harvest sources NOT showing up in search results is handled by the ckanext-harvest plugin.

ckanext/canada/logic.py
```diff
@@ -529,3 +532,64 @@ def canada_datastore_run_triggers(up_func, context, data_dict):
...
+    if not source:
+        log.error('Harvest source %s does not exist', harvest_source_id)
+        raise ObjectNotFound('Harvest source %s does not exist' % harvest_source_id)
+
+    harvest_source_id = source.id
+
+    if source.type == HARVESTER_ID:
+        return get_action('harvest_source_job_history_clear')({'user': g.user}, {'id': harvest_source_id, 'keep_current': False})
+
+    return super__harvest_source_clear(context, data_dict)
```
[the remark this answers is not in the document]
JVickery-TBS: I don't think so? as the action is defined in the parent Harvest plugin, so if we do a chained action, there is not actually a parent action registered right?

ckanext/canada/schemas/harvest.yaml
JVickery-TBS: I initially tried doing that, but then got some SQL errors because the Harvest plugin does some crazy table reference stuff during package creation

ckanext/canada/templates/base.html
JVickery-TBS: Used for some CSS to color coordinate package types. I have made the Harvest Source type the dark red color from Canada wet theme

ckanext/canada/cli.py
JVickery-TBS: The parent Harvest plugin does use that. The issue might be a CKAN Core thing, or at least a dependency thing (whichever dependency is used for the migration stuff).

So if you sub-class a plugin, and try to run the migrations, e.g. in this case `ckan db upgrade -p canada_harvest` it is not going to work. It raises a bunch of errors having to do with the name of the plugin and the directory of the migration scripts. But doing `ckan db upgrade -p harvest` does work. Which is what this crazy CLI command does programatically, it unloads `canada_harvest`, loads the parent `harvest` plugin and then runs the migrations for `harvest`.

ckanext/canada/harvesters.py
JVickery-TBS: It does in fact work interestingly enough haha.

In the gather_stage I made sure to initiate the context manager before the looping of packages so it doesn't do it for each package. But my concern is in the import_stage as the import_stage is only passed one package at a time.

It isn't as slow as you would think, but still unsure of it. Do you have any ideas to solve this? I tried using the Thread lib but that did not really work to change to the Portal app context while still maintaining the Registry app context afterwards :/

JVickery-TBS:
trashing this officially

PR #1522, opened 2024-10-18
PortalUpdater Improvements and PackageSync Model

JVickery-TBS:
feat(dev): heavily improve post to portal;

- Added DB model for portal sync tracking.
- Heaving improved copy-datasets and PortalUpdater code to catch all exceptions and log them.
- Added logic actions and views to show all the new stuff.

### DIFF
--- a/ckanext/canada/cli.py
+++ b/ckanext/canada/cli.py
@@ -12,10 +12,12 @@ import gzip
 import requests
 from collections import defaultdict
 
+from typing import Optional, Union, Tuple
+
 from contextlib import contextmanager
 from urllib.request import URLError
 from urllib.parse import urlparse
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 from ckan.logic import get_action
 from ckan import model
@@ -38,6 +40,7 @@ from ckanapi.cli.utils import completion_stats
 import ckanext.datastore.backend.postgres as datastore
 
 from ckanext.canada import triggers
+from ckanext.canada import model as canada_model
 
 BOM = "\N{bom}"
 
@@ -143,9 +146,9 @@ class PortalUpdater(object):
             log = open(self.log, 'a')
 
         registry = LocalCKAN()
-        source_ds = str(datastore.get_write_engine().url)
+        source_datastore_uri = str(datastore.get_write_engine().url)
 
-        def changed_package_id_runs(start_date, verbose=False):
+        def changed_package_id_runs(start_date, verbose:Optional[bool]=False):
             # retrieve a list of changed packages from the registry
             while True:
                 packages, next_date = _changed_packages_since(
@@ -163,7 +166,7 @@ class PortalUpdater(object):
             'canada',
             'copy-datasets',
             '-o',
-            source_ds,
+            source_datastore_uri,
             '-u',
             self.ckan_user
         ]
@@ -183,7 +186,7 @@ class PortalUpdater(object):
         # Advance generator so we may call send() below
         next(pool)
 
-        def append_log(finished, package_id, action, reason):
+        def append_log(finished, package_id, action, reason, error: Optional[Union[str, None]]=None):
             if not log:
                 return
             log.write(json.dumps([
@@ -193,6 +196,8 @@ class PortalUpdater(object):
                 action,
                 reason,
                 ]) + '\n')
+            if error:
+                log.write(error + '\n')
             log.flush()
 
         with _quiet_int_pipe():
@@ -209,16 +214,33 @@ class PortalUpdater(object):
                 stats = completion_stats(self.processes)
                 while result is not None:
                     try:
-                        package_id, action, reason = json.loads(result)
+                        package_id, action, reason, error, failure_reason, failure_trace, do_update_sync_success_time = json.loads(result)
                     except Exception as e:
                         if self.verbose:
                             print("Worker proccess failed on:")
                             print(result)
                         raise Exception(e)
-                    print(job_ids, next(stats), finished, package_id, \
-                        action, reason)
-                    append_log(finished, package_id, action, reason)
+                    _stats = next(stats)
+                    print(job_ids, _stats, finished, package_id, action, reason)
+                    if error:
+                        print(job_ids, _stats, finished, package_id, error)
+                    append_log(finished, package_id, action, reason, error)
                     job_ids, finished, result = next(pool)
+
+                    #TODO: need to actually output error to stderr!!!
+
+                    # save the sync state in the database
+                    last_successful_sync = None
+                    if do_update_sync_success_time:
+                        last_successful_sync = datetime.now(timezone.utc)
+                    else:
+                        sync_obj = canada_model.PackageSync.get(package_id=package_id)
+                        if sync_obj:
+                            last_successful_sync = sync_obj.last_successful_sync
+                    canada_model.PackageSync.upsert(package_id=package_id,
+                                                    last_successful_sync=last_successful_sync,
+                                                    error_on=failure_reason or None,
+                                                    error=failure_trace or None)
 
                 print(" --- next batch starting at: " + next_date.isoformat())
                 append_log(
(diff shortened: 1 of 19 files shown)


ckanext/canada/cli.py
```diff
@@ -209,17 +214,34 @@ def append_log(finished, package_id, action, reason):
...
-                        action, reason)
-                    append_log(finished, package_id, action, reason)
+                    _stats = next(stats)
+                    print(job_ids, _stats, finished, package_id, action, reason)
+                    if error:
+                        print(job_ids, _stats, finished, package_id, error)
+                    append_log(finished, package_id, action, reason, error)
                     job_ids, finished, result = next(pool)
 
+                    #TODO: need to actually output error to stderr!!!
```
wardi, line 230: it's ok for the parent to print to stdout for normal status messages, for errors you can `print(..., file=sys.stderr)`

ckanext/canada/cli.py
```diff
@@ -699,81 +800,140 @@ def _trim_package(pkg):
...
+        if failure_reason:
+            failure_reason += ',%s' % _failure_reason  # comma separate multiple failure reasons
+            failure_trace += '\n\n%s' % _failure_trace  # separate multiple failure traces with newlines
+        else:
+            failure_reason = _failure_reason
+            failure_trace = _failure_trace
+        error += '\n  datastore-create-failed for %s %s' % (resource['id'], str(e))
+        if verbose:
+            error += '\n    Failed with Error: %s' % failure_trace
+        pass
```
wardi, line 896: These try: except: blocks are repeated a number of times. You could create e.g. a `capture_exception_details(trying_to)` context manager to reduce the duplication and these blocks could look like:
```python
with capture_exception_details('datastore-create'):
    portal.call_action('datastore_create',
                   {"resource_id": resource['id'],
                    "fields": resource_details['data_dict'],
                    "force": True})
    action += '\n  datastore-created for ' + resource['id']
    datastore_created = True
```

ckanext/canada/model.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+Base = declarative_base(metadata=meta.metadata)
+
+
+class PackageSync(Base):
+    __tablename__ = 'package_sync'
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    package_id = Column(Unicode)
```
wardi, line 21: The package_id can be the PK, right?

ckanext/canada/model.py
```diff
@@ -0,0 +1,87 @@
...
+            package_sync.error_on = error_on
+            package_sync.error = error
+            package_sync.created = datetime.datetime.now(datetime.timezone.utc)
+            package_sync.last_successful_sync = last_successful_sync
+        else:
+            package_sync = cls(package_id=package_id, last_successful_sync=last_successful_sync,
+                               error_on=error_on, error=error)
+
+        meta.Session.add(package_sync)
+        meta.Session.commit()
```
wardi, line 54: I don't love having commit inside model methods because there's a side effect of committing all other changes that might be in the session, and looking at code that calls the method it's not obvious that that's happening.

wardi: but, we're the only user of this so I suppose it's fine

JVickery-TBS: The CKAN Core DomainObject does this with the `save` method:

```
def save(self):
        self.add()
        self.commit()
```

kinda just based it off of that. But it does use `self.Session` / `cls.Session` which just saves a copy of `meta.Session` at class instantiation ?? I think?? Should I do it that way and have a `Session = meta.Session` and use `cls.Session` in the `PackageSync` model class?

wardi: yeah it's just a pattern that I don't think is great because it doesn't mix well with the "dictization" code in ckan that adds things to the session without committing.

In this case it really doesn't matter, though might be slightly less efficient to commit after every row being created/updated.

ckanext/canada/migration/canada_public/versions/0ef791477ff0_.py
```diff
@@ -0,0 +1,36 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '0ef791477ff0'
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    from ckanext.canada.model import PackageSync
```
wardi, line 20: we can't reference the model from a migration because it will change over time. Instead either have alembic generate the migration for you ( https://github.com/ckan/ckan/pull/8239 ) , or copy and modify an example from another extension.

ckanext/canada/model.py
JVickery-TBS: from testing, there are scenarios because we are in the CLI context in which we could exit the CLI and the commits would not take affect. @wardi I could handle the session commit in the `_quiet_int_pipe` context manager? that way if it does exit, then we commit to the DB?

Otherwise we can just keep with the commit per package inside the loop

ckanext/canada/model.py
JVickery-TBS: Yeee, just gunna keep it called `package_id` so we know what the reference is. Not gunna do any crazy backrefs for this table it that is fine?

ckanext/canada/cli.py
JVickery-TBS: @wardi I added a context manager class which handles the exceptions on exit and formats them into class variables. They are then used across the code to do the remainder of the formatting/logic for the `copy-dataset` exceptions. Because we do not want to blanket all of the action calls and allow for as much of the package to get synced as possible, there is still some code reuse in the methods as we need to keep the loops to allow for this.

At least it is a bit better now, unless if there is an even better way to store the formatted exceptions, reasons, and errors in an object? I will keep trying to work on that so I can minimize the code even further

JVickery-TBS: yeee I cannot really think of a way to do this. I think it is fine though? this way it would allow us to do other things in conditional statements in the future

ckanext/canada/cli.py
```diff
@@ -77,6 +80,61 @@
...
+class CaptureExceptionDetails(object):
+    failure_reason = ''
+    failure_trace = ''
+    error = ''
+    has_exceptions = False
+    __reason = None
+    __verbose = False
+    __package_id = None
+    __resource_id = None
+    __view_id = None
```
wardi, line 92: these are class variable which are later shadowed by the instance variables created below, so they're not doing anything here.

Also I strongly recommend against using `__` in front of variables because python's "private" munging behaviour is really weird and tends to just cause problems

But even better would be replacing this whole class with a function using the @contextmanager decorator. That lets you write less code and catch exceptions in the normal python way.

JVickery-TBS: @wardi as discussed in our call, I have applied all of the new enclosure context managers with those `nonlocal` variables.

It saved especially a lot of code inside of the add_views and add_datastore methods.

ckanext/canada/tests/__init__.py
```diff
@@ -5,15 +5,18 @@
...
 from ckanext.security.model import db_setup as security_db_setup
+from ckan.cli.db import _run_migrations
 
 class CanadaTestBase(object):
     @classmethod
     def setup_method(self, method):
         """Method is called at class level before EACH test methods of the class are called.
         Setup any state specific to the execution of the given class methods.
         """
+        _run_migrations('canada_public')
```
wardi, line 16: It should be possible to use the with_plugins and clean_db fixtures (like the activity plugin tests do) instead of running migrations directly

ckanext/canada/migration/canada_public/versions/0ef791477ff0_.py
```diff
@@ -0,0 +1,51 @@
...
+    pass
+
+
+def downgrade():
+    tables = _get_tables()
+    if "package_sync" not in tables:
+        print('package_sync database table not present')
+        return
+    op.drop_table("package_sync")
+    print('package_sync database table deleted')
```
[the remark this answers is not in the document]
JVickery-TBS: some of the CKAN core ones do. Figured it would be good for some verbosity output so we know what is happening

ckanext/canada/tests/__init__.py
JVickery-TBS: @wardi yeah, so after a lot of debugging, the database in the Docker Image for CKAN is always going to be ahead of the migrations we have. That is also why I have to call `validation_create_tables` and `security_db_setup` in the test base class too.

PR #1523, opened 2024-10-22
show available resource metadata under Additional Information

RabiaSajjad:


JVickery-TBS (approved):
@RabiaSajjad as long as the Resource fields all have a good display snippets defined in our Schema so no JSON or ugly stuff displays, this looks all good.

PR #1530, opened 2024-11-08
Resource Limit per Dataset

JVickery-TBS:
feat(dev): max resources;

- Max resources per dataset.

### DIFF
--- a/ckanext/canada/helpers.py
+++ b/ckanext/canada/helpers.py
@@ -901,3 +901,9 @@ def is_user_locked(user_name):
         return True
 
     return False
+
+
+def max_resources_per_dataset():
+    max_resource_count = config.get('ckanext.canada.max_resources_per_dataset', None)
+    if max_resource_count:
+        return int(max_resource_count)
--- a/ckanext/canada/templates/package/new_resource.html
+++ b/ckanext/canada/templates/package/new_resource.html
@@ -8,7 +8,7 @@
 
 {% block form %}
   {% set max_resource_count = h.max_resources_per_dataset() %}
-  {% if g.is_registry and max_resource_count and (pkg_dict.resources|length + 1) > max_resource_count %}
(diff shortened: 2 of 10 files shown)


ckanext/canada/helpers.py
```diff
@@ -901,3 +901,9 @@ def is_user_locked(user_name):
         return True
 
     return False
+
+
+def max_resources_per_dataset():
+    max_resource_count = config.get('ckanext.canada.max_resources_per_dataset', None)
```
wardi, line 907: Use an integer instead of None here to prevent an exception

ckanext/canada/templates/package/new_resource.html
```diff
@@ -6,6 +6,27 @@
   </div>
 {% endblock %}
 
+{% block form %}
+  {% set max_resource_count = h.max_resources_per_dataset() %}
+  {% if g.is_registry and max_resource_count and (pkg_dict.resources|length + 1) > max_resource_count %}
```
wardi, line 11:
```suggestion
  {% if g.is_registry and max_resource_count and pkg_dict.resources|length >= max_resource_count %}
```

ckanext/canada/validators.py
```diff
@@ -531,3 +531,35 @@ def protect_registry_access(key, data, errors, context):
...
+
+    new_resource_count = len(set(k[1] for k in data.keys() if k[0] == 'resources'))
+    if not new_resource_count:
+        return
+
+    current_resource_count = model.Session.query(model.Resource.id)\
+        .filter(model.Resource.package_id == package_id)\
+        .filter(model.Resource.state == 'active').count()
+    if not current_resource_count:
+        return
```
wardi, line 554: I'd move this into the `if new_resource_count > int(max_resource_count):` block so you can skip the extra db query most of the time

PR #1531, opened 2024-11-14
add new indigenous language to presets.yaml

RabiaSajjad:


ckanext/canada/schemas/presets.yaml
```diff
@@ -3842,6 +3842,10 @@ presets:
         en: Indonesian
         fr: indonésien
       value: id
+    - label:
+        en: Innu-Aimun
+        fr: Innu-Aimun
+      value: innuaimun
```
JVickery-TBS, line 3848: @RabiaSajjad I think we are using ISO codes for the values. https://iso639-3.sil.org/code/moe <- soe should be `moe` I guess.

JVickery-TBS (changes requested):
just the above comment

JVickery-TBS (approved):

PR #1532, opened 2024-11-14
TravelQ Destination Fields Revamp

JVickery-TBS:
This will be one of the first time we stop doing parity between PSQL and Excel. Business requirements are reaching a point in which we will just have to stop doing crazy things in Excel.


JVickery-TBS:
Have to fix the regex for multi-destination field. Trying to figure out the capture grouping, not sure if it needs to be some dynamic grouping, or if we can do some really scary looping in PSQL

ckanext/canada/tables/travelq.yaml
```diff
@@ -187,37 +187,189 @@ resources:
   - datastore_id: destination_en
     datastore_type: text
     label:
-      en: Places visited (English)
+      en: Place visited (English)
       fr: Endroits visités (anglais)
     description:
-      en: To include names of all places visited during the trip in question, in English
+      en: To include name of the place visited during the trip in question, in English.
```
wardi, line 193:
```suggestion
      en: To include name of the first place visited during the trip in question, in English.
```
?

Also really unfortunate to see these fields collected in French and English. How often will they be different? Wouldn't using a geocoder on this data to normalize and present both languages be much better?

ckanext/canada/tests/test_travelq.py
```diff
@@ -7,6 +10,8 @@
 
 from ckanext.recombinant.tables import get_chromo
 
+DESTINATION_ENDPOINT = 'https://restcountries.com/v3.1/all?fields=name,capital'
```
wardi, line 13: We have a specific list of countries recognized by the GoC that is maintained with https://github.com/open-data/ckanext-canada/blob/master/bin/download_country.py to add XK and TW

ckanext/canada/tables/travelq.yaml
```diff
@@ -187,37 +187,189 @@ resources:
...
       fr: Cette zone indique, en anglais, le nom de tous les endroits visités pendant le voyage.
+    form_help:
+      en: Note, use the format {City Name}, {Country Name} (e.g. Ottawa, Canada or New York City, United States of America)
+      fr: Note, use the format {City Name}, {Country Name} (e.g. Ottawa, Canada or New York City, United States of America)
     obligation: Mandatory
     excel_required: true
+    # Dev NOTE: excel_error_formula: Excel and Open Office Calc parity issues with REGEX. REGEX does not exist in Excel, and sheetname references in OOC work differently.
```
wardi, line 200: open office calc? I've found the sheet references that recombinant generates e.g. for validation and links work fine

ckanext/canada/tables/travelq.yaml
```diff
@@ -187,37 +187,189 @@ resources:
...
       fr: Endroits visités (anglais)
     description:
-      en: To include names of all places visited during the trip in question, in English
+      en: To include name of the place visited during the trip in question, in English.
       fr: Cette zone indique, en anglais, le nom de tous les endroits visités pendant le voyage.
+    form_help:
+      en: Note, use the format {City Name}, {Country Name} (e.g. Ottawa, Canada or New York City, United States of America)
+      fr: Note, use the format {City Name}, {Country Name} (e.g. Ottawa, Canada or New York City, United States of America)
     obligation: Mandatory
     excel_required: true
```
wardi, line 199: this is close: check for one comma and something non-whitespace on either side
```suggestion
    excel_required: true
    excel_error_formula: 'OR({default_formula},LEN({cell})<>LEN(SUBSTITUTE({cell},",","")+1,LEN(TRIM(LEFT({cell},FIND(",",{cell})-1)))>0,LEN(TRIM(RIGHT({cell},LEN({cell})-FIND({cell},","))))>0)
```

ckanext/canada/tables/travelq.yaml
JVickery-TBS: Yeah it is really stupid to collect these fields in both languages. I doubt that many would be different in French. Idk how we would do the geo stuff though, maybe an improvement in the future as the timeline for this is tight and not sure which library would be best and the network limitations for any API stuff we would possibly need.

So I guess we will just collect them for now...

ckanext/canada/tests/test_travelq.py
JVickery-TBS: @wardi so just use pyyaml and the choices file?? I don't specifically need the capital city names I guess, I could just double up the country name

ckanext/canada/tables/travelq.yaml
JVickery-TBS: @wardi `err:508` in excel on this

ckanext/canada/tests/test_travelq.py
wardi: might be better than adding a dependency on another 3rd party service. We could collect the capital names from the same source that script uses if you really want.

Was the purpose to check that your regex covers the characters required to represent common city names?

ckanext/canada/tables/travelq.yaml
wardi: I'm sure I lost my way pasting in some of the parens or something. I'll take another look

wardi: Try it again now

wardi:
For both travelq and hospitalityq is it better to pair the 1,2,3,other English and french fields instead of having them separated? Just a general usability question, not sure the best person to ask.

ckanext/canada/tests/test_travelq.py
JVickery-TBS: yeah was just trying to do city names as well

JVickery-TBS:
Yeah it probably is a lot better to group them that way, will do that.

ckanext/canada/tables/travelq.yaml
```diff
@@ -187,37 +187,189 @@ resources:
...
+    description:
+      en: To include names of all other places visited during the trip in question, in English.
+      fr: Cette zone indique, en anglais, le nom de tous les endroits visités pendant le voyage.
+    form_help:
+      en: Note, use the format {City Name}, {Country Name};{City 2 Name}, {Country 2 Name} (e.g. Ottawa, Canada;New York City, USA)
+      fr: Note, use the format {City Name}, {Country Name};{City 2 Name}, {Country 2 Name} (e.g. Ottawa, Canada;New York City, USA)
+    obligation: Optional
+    excel_required: false
+    # Dev NOTE: excel_error_formula: Excel and Open Office Calc parity issues with REGEX. REGEX does not exist in Excel.
+    excel_required_formula: NOT(ISBLANK({destination_other_fr}))
```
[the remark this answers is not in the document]
JVickery-TBS: missing rest of formula here?

ckanext/canada/tables/travelq.yaml
JVickery-TBS: still broken sadly. I worked on these formulas in different ways for like 6 hours and couldn't get them. So ultimately gave up and easily convinced people to drop the parity between PSQL and Excel

JVickery-TBS:
@wardi okay this should be good for a re-review now. And since we are supporting even more optional formatting here, I doubt we would be able to do a nice and reliable Excel formula. So I think we can just exclude that from these fields unless if you really want to bash your head against it haha

ckanext/canada/tables/travelq.yaml
wardi: @JVickery-TBS this one works with either 1 or 2 commas
excel_brain_damage

JVickery-TBS: `{1,2}` will cause error in python as Recombinant formats the formulae. Ima just executive decision to not do excel error formulas for these fields and call it a day.

wardi: right, there's a format() in there. I just doubled them up to fix.

JVickery-TBS: there is actually also an assertion in Recombinant which forces the format values to be exact, so everything needs a replacement. We are just not gunna support it here with this field. And I doubt we could get a formula for the multi-destination field either

PR #1534, opened 2024-11-15
Dataset & Resource Relationship Fields

JVickery-TBS:


ckanext/canada/plugins.py
```diff
@@ -407,6 +407,16 @@ def before_index(self, data_dict):
...
                 cr.pop('__extras', None)
 
+        if data_dict.get('relationship'):
+            related_relationships = []
+            related_types = []
+            for rel in data_dict['relationship']:
+                related_relationships.append(rel.get('related_relationship'))
+                related_types.append(rel.get('related_type'))
+            data_dict['related_relationship'] = related_relationships
+            data_dict['related_type'] = related_types
```
wardi, line 417: list comprehensions are shorter and run a little faster than for loops
```suggestion
            related_relationships = [rel['related_relationship'] for rel in data_dict['relationship']]
            related_types = [rel['related_type'] for rel in data_dict['relationship']]
```

ckanext/canada/schemas/presets.yaml
```diff
@@ -2755,25 +2752,64 @@ presets:
...
+        en: The URL for online access to the related record
+        fr: FR The URL for online access to the related record FR
+    # copied from fluent_text preset
+    form_snippet: fluent_text.html
+    display_snippet: fluent_link.html
+    display_attributes:
+      style: "word-wrap: break-word"
+    error_snippet: fluent_text.html
+    validators: fluent_text
+    output_validators: fluent_core_translated_output
```
wardi, line 2804: not a core field so this isn't the right validator

ckanext/canada/schemas/presets.yaml
```diff
@@ -2802,11 +2835,45 @@ presets:
...
+    - label:
+        en: Referenced by
+        fr: FR Referenced par FR
+      value: referenced_by
+    form_include_blank_choice: true
+    form_snippet: select.html
+    display_snippet: select.html
+    validators: scheming_required scheming_choices
+    required: true
+    form_panel: relationships
```
wardi, line 2876: is the form_panel stuff used anymore?

JVickery-TBS: It is in Resources template, but I think that the form_panel was mainly for the old related resources panel? So we can probably just not do panels? and then just call `super()` in our `resource_form.html`.

ckanext/canada/schemas/presets.yaml
JVickery-TBS: just use `fluent_text_output`?

ckanext/canada/schemas/presets.yaml
JVickery-TBS: hmmm the form_panels seem to be used to also hide and order a lot of the resource form fields. So just gunna leave it in for now

ckanext/canada/schemas/presets.yaml
```diff
@@ -2755,25 +2752,64 @@ presets:
...
     form_snippet: select.html
     display_snippet: select.html
     validators: scheming_required scheming_choices
+    required: true
+    form_panel: relationships
+
+# Field = Related Record URL.
+- preset_name: canada_dataset_related_url
+  values:
+    field_name: related_url_translated
```
wardi, line 2776: the `_translated` bit was added for fluent versions of core fields, it's not necessary for new custom fields (and it's a little odd to call urls translated)

JVickery-TBS: I guess we might not need to have a URL (English) and URL(French) then? otherwise the `_translated` is really just for the get_translated helper.

ckanext/canada/templates/scheming/display_snippets/dataset_relationship.html
```diff
@@ -0,0 +1,17 @@
...
+          {%- snippet 'scheming/snippets/display_field.html',
+            field=subfield,
+            data=field_data,
+            entity_type=entity_type,
+            object_type=object_type -%}
+        {%- endfor -%}
+      </p>
+    {%- endfor -%}
+  </div>
+{% endblock %}
```
wardi, line 17: shouldn't this be automatic, or is there a modification to the default here?

ckanext/canada/templates/scheming/package/snippets/resource_form.html
```diff
@@ -11,41 +11,31 @@
...
+        {{ _("Metadata Fields for Resource Relationships") }}
+      </summary>
+      <div class="tgl-panel">
+        {%- for field in schema.resource_fields -%}
+          {%- if field.form_panel == 'relationships' -%}
+            {%- snippet 'scheming/snippets/form_field.html', field=field, data=data, errors=errors, entity_type='dataset', object_type=dataset_type -%}
+          {%- endif -%}
+        {%- endfor -%}
+      </div>
+    </details>
```
wardi, line 38: Before resources were used *as* relationships for datasets because we didn't have repeating subfields.

Now resource relationships are relationships between resources, and the dataset relationships are defined in the dataset form. Will it be confusing for users to have relationships defined at both levels? Are there many cases where users will want to define relationships between resources (and if they do, maybe there are too many resources in a dataset?)

JVickery-TBS: Well the original request for bringing this all back was that Pat wanted to have references between Resources. E.g. this resource is a JSON schema which DEFINES this other resource which is a dataset.

That is not what the old way was, it was a resource being used to have some relationship to the dataset that the resource belonged to. So did not meet the requirement Pat wanted. So the original resource as relationship is moved to the repeating fields in the Dataset. And the new requirement that Pat wants are the repeating fields in the Resources.

I think both make sense. Are they confusing? absolutely, but so was the old resource as a relationship, it didn't even really make much sense and using a resource as not really a resource was bad. These at least make more sense. There will for sure need to be updates to the Registry Operations Guide for this.

But the Dataset repeating fields make a lot of sense for all of those users that keep segmenting their datasets for fiscal years. E.g. Fish Data 2024 continues Fish Data 2023, and/or Fish Data 2023 is Continued by Fish Data 2024

ckanext/canada/schemas/presets.yaml
wardi: There will be cases where users need to provide separate EN and FR urls so I don't think we can get rid of that. Yes the `_translated` is just for `get_translated` which isn't used for custom fields, only core fields.

ckanext/canada/templates/scheming/display_snippets/dataset_relationship.html
wardi: thanks, maybe add a comment about why it's required and when we might be able to remove it

ckanext/canada/schemas/presets.yaml
JVickery-TBS: ah okay I see, `_translated` has been removed.

ckanext/canada/templates/scheming/display_snippets/dataset_relationship.html
JVickery-TBS: comments added.

JVickery-TBS:
@wardi this has been changed to target the 2.10 branch.

PR #1536, opened 2024-11-21
add missing translations for validation

RabiaSajjad:


JVickery-TBS:
@RabiaSajjad this looks good, but could you also fix a small issue with the Info schema?

We need to include these two fields in the schema so resource_show will display them for Info type records:

- https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/schemas/dataset.yaml#L490
- https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/schemas/dataset.yaml#L495

You might also need to add the translations into the label dict for those fields in the YAMLs, in the scenario that some of our templates (or validation templates) do not call gettext and only use the scheming label dict.

JVickery-TBS:
@RabiaSajjad as discussed, Info schema also needs:
```
- preset: canada_resource_validation_schema
- preset: canada_resource_validation_options
```

changes/1536.bugfix
```diff
@@ -0,0 +1,3 @@
+Add missing translations for validation:
```
JVickery-TBS, line 1: @RabiaSajjad could you also just note the "fix" to adding the validation fields to the Info schema? Thanks!

JVickery-TBS (changes requested):

PR #1537, opened 2024-11-22
Filesize Display Snippet

JVickery-TBS:
feat(templates): filesize display snippet;

- Added a filesize scheming display snippet.


ckanext/canada/templates/scheming/display_snippets/filesize.html
```diff
@@ -0,0 +1 @@
+<span title="{{ data[field.field_name] }} bytes" class="canada-abbr">{{ data[field.field_name]|filesizeformat }}</span>
```
wardi, line 1: Does this work on mobile or only hovering with a desktop browser?

Might be good to format the bytes number with commas or spaces so it's easier to read.

JVickery-TBS: @wardi no titles do not show tooltips on mobile. I added a mobile display for it now, just shows a small element after it in brackets.

The "bytes" is now formatted too.

PR #1538, opened 2024-11-22
Fix Reference Assignment Error in PortalUpdater Delete Views

JVickery-TBS:
fix(dev): syntax;

- PortalUpdater reference error in delete views.

### DIFF
--- a/ckanext/canada/cli.py
+++ b/ckanext/canada/cli.py
@@ -208,6 +208,8 @@ class PortalUpdater(object):
                 activity_date.isoformat()
             )
 
+            has_errored = False
+
             for packages, next_date in (
                     changed_package_id_runs(activity_date, verbose=self.verbose)):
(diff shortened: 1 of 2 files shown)


ckanext/canada/cli.py
```diff
@@ -320,6 +328,9 @@ def _copy_datasets(source_datastore_uri: Optional[Union[str, None]], user: Optio
 
         packages = iter(sys.stdin.readline, '')
         for package in packages:
+            sys.stdout.write(json.dumps(['package_id', 'action', 'reason', 'error', 'failure_reason', 'failure_trace', False]) + '\n')
+            sys.stdout.flush()
+            continue
```
wardi, line 333: unconditionally skip all the rest of the code in this loop?

JVickery-TBS: oh damn my debug code MY BAD

PR #1539, opened 2024-11-22
display translated validation status

RabiaSajjad:


ckanext/canada/templates/scheming/display_snippets/validation_status.html
```diff
@@ -0,0 +1,12 @@
...
+    'failure': _('Invalid data'),
+    'error': _('Error during validation'),
+    'unknown': _('Data validation unknown'),
+}
+-%}
+{%- if data[field.field_name] in ['success', 'failure', 'error'] -%}
+    {{ _(messages[data[field.field_name]]) }}
+{% else %}
+    {{ _(messages['unknown']) }}
+{% endif %}
```
wardi, line 12:
```suggestion
{%- set messages = {
    'success': _('Valid data'),
    'failure': _('Invalid data'),
    'error': _('Error during validation'),
}
-%}
{{ messages.get([data[field.field_name]], _('Data validation unknown')) }}
```

wardi: this is simpler and doesn't call `_()` twice on the same string

wardi: modified to not display 'created' and 'running'

PR #1541, opened 2024-11-28
Code Changes for Re-index Additions

JVickery-TBS:
Some code changes related to the additions here: https://github.com/open-data/ckan/pull/185

### DIFF
--- a/ckanext/canada/logic.py
+++ b/ckanext/canada/logic.py
@@ -71,7 +71,7 @@ JOB_MAPPING = {
         'icon': 'fa-trash',
         'rid': lambda job_args: job_args,
     },
-    'ckan.lib.search.jobs.reindex': {
+    'ckan.lib.search.jobs.reindex_packages': {
         'icon': 'fa-database',
         'gid': lambda job_kwargs: job_kwargs.get('group_id')
     }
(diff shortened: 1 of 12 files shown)


ckanext/canada/logic.py
```diff
@@ -442,44 +446,70 @@ def canada_job_list(up_func, context, data_dict):
...
+                    resource = get_action('resource_show')({'user': current_user}, {'id': rid})
+                except (ObjectNotFound, NotAuthorized):
+                    continue
+                job_info = {'name_translated': resource.get('name_translated'),
+                            'resource_id': rid,
+                            'url': h.url_for('dataset_resource.read',
+                                            id=resource.get('package_id'),
+                                            resource_id=rid)}
+            if gid:
+                if gid == config.get('ckan.site_id'):
```
wardi, line 493: would be cleaner to use a different key or None value here to indicate that the job is for the whole site. It's unlikely but possible that the site id matches a group or org name value.

ckanext/canada/logic.py
```diff
@@ -442,44 +446,70 @@ def canada_job_list(up_func, context, data_dict):
...
+                    job_info = {'name': _('Entire Site')}
+                else:
+                    try:
```
wardi, line 496: using a `'reindex_type'` key for group/organization in the job would mean not having to try/except group/organization here too

ckanext/canada/logic.py
JVickery-TBS: @wardi yeah sorry just fixed this a bit with a different condition because of the changes to the background jobs stuff in `open-data/ckan`

ckanext/canada/logic.py
JVickery-TBS: we actually don't have any groups in our site, and the capability is fully shut off, so this is to just cover some 500 error bases in the view really. And as this is just an admin view that is not really used much at all, it could just be fine.

Unless if you think that having a `reindex_type` key in core would be beneficial to core?

PR #1543, opened 2024-12-10
Allow White Listed DS Sources

JVickery-TBS:
feat(dev): white list source domains;

- White list source domains.

### DIFF
--- a/ckanext/canada/plugins.py
+++ b/ckanext/canada/plugins.py
@@ -86,6 +86,7 @@ class CanadaValidationPlugin(FrictionlessPlugin):
         field_candidates.append({'type':'string'})
 
 
+@p.toolkit.blanket.config_declarations
 class CanadaThemePlugin(p.SingletonPlugin):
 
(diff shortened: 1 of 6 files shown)


ckanext/canada/plugins.py
```diff
@@ -564,17 +576,29 @@ def get_auth_functions(self):
...
+                    'Only uploaded resources and white listed sources can be added to the Data Store.')
                 return False
 
+            if not res.get('url_type'):
+                allowed_domains = p.toolkit.config.get('ckanext.canada.datastore_source_domain_allow_list', [])
+                url = res.get('url')
+                url_parts = urlsplit(url)
+                if url_parts.netloc not in allowed_domains:
+                    log.error(
+                        'Only uploaded resources and white listed sources can be added to the Data Store.')
```
wardi, line 599: is it appropriate to log an error when a resource can't be uploaded? This method is here to check if it can be uploaded so "no" seems like a normal case not an error, right?

PR #1544, opened 2024-12-13
Implement IUsage

JVickery-TBS:
Implement plugin: https://github.com/NorwegianRefugeeCouncil/ckanext-api-tracking

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,7 +86,7 @@ jobs:
             pip install -e 'git+https://github.com/ckan/ckantoolkit.git#egg=ckantoolkit' -r 'https://raw.githubusercontent.com/ckan/ckantoolkit/master/requirements.txt'
             pip install -e 'git+https://github.com/open-data/frictionless-py.git@canada-v2.10#egg=frictionless' -r 'https://raw.githubusercontent.com/open-data/frictionless-py/canada-v2.10/requirements.txt'
             pip install -e 'git+https://github.com/open-data/ckanext-security.git@canada-v2.10#egg=ckanext-security' -r 'https://raw.githubusercontent.com/open-data/ckanext-security/canada-v2.10/requirements.txt'
-            pip install -e 'git+https://github.com/NorwegianRefugeeCouncil/ckanext-api-tracking.git#egg=ckanext-api-tracking' -r 'https://raw.githubusercontent.com/NorwegianRefugeeCouncil/ckanext-api-tracking/main/requirements.txt'
+            pip install -e 'git+https://github.com/open-data/ckanext-api-tracking.git@canada-v2.10#egg=ckanext-api-tracking' -r 'https://raw.githubusercontent.com/open-data/ckanext-api-tracking/canada-v2.10/requirements.txt'
             find \/srv\/app\/ -name '*.pyc' -delete
       - run:
           name: Setup extension (CKAN 2.10 PY3.9)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,7 @@ jobs:
             pip install -e 'git+https://github.com/ckan/ckantoolkit.git#egg=ckantoolkit' -r 'https://raw.githubusercontent.com/ckan/ckantoolkit/master/requirements.txt'
             pip install -e 'git+https://github.com/open-data/frictionless-py.git@canada-v2.10#egg=frictionless' -r 'https://raw.githubusercontent.com/open-data/frictionless-py/canada-v2.10/requirements.txt'
             pip install -e 'git+https://github.com/open-data/ckanext-security.git@canada-v2.10#egg=ckanext-security' -r 'https://raw.githubusercontent.com/open-data/ckanext-security/canada-v2.10/requirements.txt'
+            pip install -e 'git+https://github.com/NorwegianRefugeeCouncil/ckanext-api-tracking.git#egg=ckanext-api-tracking' -r 'https://raw.githubusercontent.com/NorwegianRefugeeCouncil/ckanext-api-tracking/main/requirements.txt'
             find \/srv\/app\/ -name '*.pyc' -delete
       - run:
           name: Setup extension (CKAN 2.10 PY3.9)
--- a/ckanext/canada/helpers.py
+++ b/ckanext/canada/helpers.py
@@ -1,5 +1,7 @@
 from typing import Optional, Union
 
+import os
+from cryptography.fernet import Fernet
 import json
 import re
 import inspect
@@ -949,3 +951,15 @@ def max_resources_per_dataset():
     max_resource_count = config.get('ckanext.canada.max_resources_per_dataset', None)
     if max_resource_count:
         return int(max_resource_count)
+
+
+def api_tracking_extras(id: str):
+    fernet_key = os.environ.get('CKAN_API_TRACKING_SECRET')
+    if not fernet_key:
+        return
+
+    fernet_key = fernet_key if isinstance(fernet_key, bytes) else fernet_key.encode()
+    fernet = Fernet(fernet_key)
+
+    #TODO: get the extras from TrackingUsage model.
+    #      decrypt and json loads.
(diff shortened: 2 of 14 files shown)


.circleci/config.yml
```diff
@@ -86,6 +86,7 @@ jobs:
             pip install -e 'git+https://github.com/ckan/ckantoolkit.git#egg=ckantoolkit' -r 'https://raw.githubusercontent.com/ckan/ckantoolkit/master/requirements.txt'
             pip install -e 'git+https://github.com/open-data/frictionless-py.git@canada-v2.10#egg=frictionless' -r 'https://raw.githubusercontent.com/open-data/frictionless-py/canada-v2.10/requirements.txt'
             pip install -e 'git+https://github.com/open-data/ckanext-security.git@canada-v2.10#egg=ckanext-security' -r 'https://raw.githubusercontent.com/open-data/ckanext-security/canada-v2.10/requirements.txt'
+            pip install -e 'git+https://github.com/NorwegianRefugeeCouncil/ckanext-api-tracking.git#egg=ckanext-api-tracking' -r 'https://raw.githubusercontent.com/NorwegianRefugeeCouncil/ckanext-api-tracking/main/requirements.txt'
```
wardi, line 89: likely want to fork this so we can control releases

ckanext/canada/helpers.py
```diff
@@ -949,3 +952,42 @@ def max_resources_per_dataset():
...
+                   'user_id': tracking_obj.user_id}
+
+    if tracking_obj.user_id:
+        user_obj = model.User.get(tracking_obj.user_id)
+        if user_obj:
+            return_dict['user_name'] = user_obj.name
+            return_dict['user_fullname'] = user_obj.fullname
+
+    if tracking_obj.extras and (tracking_obj.extras.get('REQUEST_ARGS') or tracking_obj.extras.get('REQUEST_BODY')):
+        fernet_key = fernet_key if isinstance(fernet_key, bytes) else fernet_key.encode()
```
wardi, line 979: maybe `encode('ascii')` otherwise it's going to default to the system encoding, but I see that Fernet() accepts either bytes or str, so why the conversion here?

JVickery-TBS: it screamed at me that it needed bytes

.circleci/config.yml
JVickery-TBS: I did do a bunch of upstream contribs this week that were accepted. But added some more this morning and last night that would need to be brought in. So a fork might be needed, but I really don't want to Fork another repo hahaha

wardi: what if they merge or change something that breaks your tidy release process though? I guess you could pin commit ids here..

ckanext/canada/helpers.py
wardi: works for me?
```python
Fernet('YWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWE=')
<cryptography.fernet.Fernet object at 0x7f5d1a9fa530>
```

ckanext/canada/helpers.py
```diff
@@ -949,3 +952,42 @@ def max_resources_per_dataset():
...
+        fernet = Fernet(fernet_key)
+
+        value = tracking_obj.extras.get('REQUEST_ARGS', tracking_obj.extras.get('REQUEST_BODY'))
+        value = value if isinstance(value, bytes) else value.encode()
```
wardi, line 983: again, `encode()` here is relying on the system encoding. `encode('utf-8')` would be safer.

.circleci/config.yml
JVickery-TBS: hmm yeah, okay I will fork the repo and make a v2.10 branch

ckanext/canada/helpers.py
JVickery-TBS: ah, it is the encrypt/decrypt methods that scream for the bytes. Will remove the encode on the key

ckanext/canada/helpers.py
```diff
@@ -949,3 +952,42 @@ def max_resources_per_dataset():
...
+        value = fernet.decrypt(value)
+        value = value if isinstance(value, str) else value.decode()
```
wardi, line 985: fernet.decrypt always returns bytes. Most byte values would fail to decode here, right?

.circleci/config.yml
wardi: I always went with canada-XX for the branch names to not confuse with upstream releases

ckanext/canada/plugins.py
```diff
@@ -921,31 +1016,6 @@ def get_validators(self):
...
-                #FIXME: make sure username special chars are handled
-                # the values in the tuple HAVE to be str types.
-                extra = [('X-LogExtra', f'user={contextual_user} {log_extra}')]
-
-            return start_response(
-                status,
-                response_headers + extra,
-                exc_info)
-
-        return self.app(environ, _start_response)
```
wardi, line 946: This will render the graphana reporting useless for APIs. Consider removing it when graphana and log reporting are no longer used

.circleci/config.yml
JVickery-TBS: forked and branched with my additional code changes for us. (they are all already in upstream PRs)

ckanext/canada/templates/ajax_snippets/api_tracking_info.html
```diff
@@ -0,0 +1,56 @@
...
+
+Example
+
+    {% snippet 'ajax_snippets/api_tracking_info.html', id=id, embedded=true %}
+
+#}
+
+{% set id = h.sanitize_id(id) %}
+{% if id and h.check_access('sysadmin') %}
+  {% set api_tracking_info = h.canada_api_tracking_info(id) %}
```
wardi, line 15: template helpers are normally used for small bits of formatting code, not code that fetches a bunch of information from the DB. You probably used a helper just because that was your only option with ajax snippets, but that's because ajax snippets are terrible https://github.com/ckan/ckan/issues/7619

It would be much better to create a normal flask view and a new action to fetch the data. That way the permission checking can be done in the normal ckan way in the action instead of a template if statement.

ckanext/canada/plugins.py
JVickery-TBS: Our graphana "fields" broke a while ago and we never got them "fixed." So for CKAN logs in Graphana, we really only have the raw log files. And I am still logging to the syslog so it helps us debug and we still have those logs going to LogAnalytics.

I still have to figure out how to purge all these DB logs though, and how often we want to do that.

ckanext/canada/templates/ajax_snippets/api_tracking_info.html
JVickery-TBS: yeah that would be better. I still want to have it in an ajax modal however :/ as I think it would be not great to have to go fetch and decrypt all payloads in the sysadmin view. That way, we would only fetch and decrypt on user request

wardi: There's nothing preventing us from ajax-ing in the response to a normal flask view.

ckanext/canada/plugins.py
wardi: Even without LogAnalytics custom fields this lets us quickly check the slowest api calls and immediately tell which user and org were making them.

I'm just suggesting removing this code as a later PR because it's been very useful in the past.

JVickery-TBS: which part of this code logs the execution time? is it just some CKAN core thing with the log_extra? the logging here isn't enough eh? https://github.com/open-data/ckanext-canada/blob/removal/api-logging-middleware/ckanext/canada/plugins.py#L866

wardi: nginx logs include execution time

wardi:
Copying all data the user sends from every API call into the application DB could make performance of our site much worse, especially under load, and make it significantly harder to debug while a performance problem is happening. It may also bloat our application DB significantly.

Rate and size limits on data collection and regular pruning of data would be a good start. Even better would be sending this data to a separate DB or logging service.

JVickery-TBS:
this is garbo, try again :'(

PR #1547, opened 2024-12-19
CKAN Coding Standards

JVickery-TBS:


wardi:
I kept the old migration scripts around as a record of the data schema changes in the past in case we ever need to compare data from backups with newer data. The other deletions should be fine but it's worth mentioning them in the changelog at least.

JVickery-TBS:
> I kept the old migration scripts around as a record of the data schema changes in the past in case we ever need to compare data from backups with newer data. The other deletions should be fine but it's worth mentioning them in the changelog at least.

@wardi I will just add a note to the changelogs for the removal of all this stuff. as we are cross training soon, I am trying to slim down all the actual code to operational things. So old migration scripts can just be found in git history if anyone needs them

.github/workflows/pyright.yml
```diff
@@ -29,9 +29,10 @@ jobs:
             source /home/runner/work/app/bin/activate
             pip install --upgrade setuptools==70.0.0
             pip install --upgrade pip==23.2.1
-            pip install -e 'git+https://github.com/open-data/ckan@canada-v2.10#egg=ckan' -r 'https://raw.githubusercontent.com/open-data/ckan/canada-v2.10/requirements.txt' -r 'https://raw.githubusercontent.com/open-data/ckan/canada-v2.10/dev-requirements.txt'
+            pip install -e 'git+https://github.com/open-data/ckan@canada-v2.10#egg=ckan' -r 'https://raw.githubusercontent.com/open-data/ckan/canada-v2.10/requirement-setuptools.txt' -r 'https://raw.githubusercontent.com/open-data/ckan/canada-v2.10/requirements.txt' -r 'https://raw.githubusercontent.com/open-data/ckan/canada-v2.10/dev-requirements.txt'
```
wardi, line 32: if requirements change on a branch they won't be picked up as part of the pyright run until they are merged. Can we reference the files in `/home/runner/work/app/src/ckanext-canada` instead?

ckanext/canada/cli.py
```diff
@@ -540,35 +566,37 @@ def _capture_exception_details(_reason: str, _package_id: str):
...
-def _changed_datasets(since_date, server, brief):
+def _changed_datasets(since_date: str, server: Optional[str],
+                      brief: Optional[bool]):
     """
     Produce a list of dataset ids and requested dates. Each package
     id will appear at most once, showing the activity date closest
     to since_date. Requested dates are preceeded with a "#"
     """
-    since_date = isodate(since_date, None)
+    since_datetime: Optional[datetime] = isodate(since_date, cast(Context, {}))
```
wardi, line 576: why is `since_datetime` optional? If we don't get a datetime wouldn't it be better to exit with an error here?

ckanext/canada/cli.py
```diff
@@ -683,25 +713,45 @@ def _load_suggested(use_created_date, filename):
             "state": "active",
             "id": uuid,
             "title_translated": {
-                'en': str(row['title_en'], 'utf-8'),
-                'fr': str(row['title_fr'], 'utf-8')
+                # type_ignore_reason: checking existance
+                'en': str(row['title_en'],  # type: ignore
+                          'utf-8') if row['title_en'] else '',
```
wardi, line 718:
```suggestion
                'en': row['title_en'],
```
Should work fine now that `.encode('utf-8')` has been removed above?

For the type checks we could define the set of keys with a `class Suggested(TypedDict): ...` and then cast the row? There must be a clean way of doing DictReader dict key checks with pyright.

ckanext/canada/helpers.py
```diff
@@ -218,11 +227,13 @@ def canada_date_str_to_datetime(date_str: str) -> 'datetime.datetime':
         if not m:
             raise ValueError('Unable to parse %s as seconds.microseconds' %
                              time_tuple[5])
-        seconds = int(m.groupdict().get('seconds'))
+        # type_ignore_reason: incomplete typing
+        seconds = int(m.groupdict().get('seconds'))  # type: ignore
```
wardi, line 231:
```suggestion
        seconds = int(m.groupdict()['seconds'])
```
would this let you remove the `type: ignore`? You know that there will be a seconds group and its value will be a string.

ckanext/canada/helpers.py
```diff
@@ -243,6 +254,7 @@ def normalize_strip_accents(s: Union[str, bytes]) -> str:
     """
     if isinstance(s, str):
         return s
```
wardi, line 256: this would skip the normalization. that doesn't seem right.

ckanext/canada/helpers.py
```diff
@@ -505,19 +508,19 @@ def json_loads(value: str) -> Dict[str, Any]:
 
 
 def get_datapreview(res_id: str) -> str:
-    dsq_results = ckan.logic.get_action('datastore_search')(
-        {}, {'resource_id': res_id, 'limit': 100})
+    dsq_results = t.get_action('datastore_search')(
```
wardi, line 511: we *really* shouldn't be using `t` as a global variable for ckan.toolkit. One-letter variables are almost always short-lived loop variables and the like.

There is precedence for using `tk` for the toolkit module, so we could change it to that.

ckanext/canada/logic.py
```diff
@@ -325,30 +341,35 @@ def canada_guess_mimetype(context: Context, data_dict: DataDict) -> str:
...
-                mimetype = f['value']
-            if 'replaces' in f:
-                for r in f['replaces']:
-                    if mimetype.lower() == r.lower():
-                        mimetype = f['value']
+        preset_field = scheming_get_preset('canada_resource_format')
+        fmt_choices = []
+        if preset_field and 'choices' in preset_field:
+            fmt_choices = preset_field['choices']
+        if fmt_choices:
```
wardi, line 348: shouldn't need an if around this loop if `fmt_choices` is always a list, right?

ckanext/canada/validators.py
```diff
@@ -112,8 +112,9 @@ def user_read_only_json(key: FlattenKey,
 
     if hasattr(package, key[0]):
         data[key] = json.loads(getattr(package, key[0]))
-    else:
-        data[key] = json.loads(package.extras.get(key[0], 'None'))
+    elif hasattr(package, 'extras'):
+        # type_ignore_reason: checking attribute
+        data[key] = json.loads(package.extras.get(key[0], 'None'))  # type: ignore
```
wardi, line 117:
```suggestion
        data[key] = json.loads(package.extras.get(key[0], 'null'))  # type: ignore
```
None is invalid json

ckanext/canada/view.py
```diff
@@ -201,13 +240,26 @@ class CanadaUserRegisterView(UserRegisterView):
     def post(self):
         params = parse_params(request.form)
         email = params.get('email', '')
+        if isinstance(email, list):
+            email = email[0]
```
wardi, line 244: All of these could be replaced like this:
```suggestion
        email = request.form.get('email')
```
MultiDict returns the first item with `[]` and `get()` when multiple are passed.

.github/workflows/pyright.yml
JVickery-TBS: that is for our `ckan` branch, the install of the `ckanext-canada` is done a couple lines below and uses the checkout action from the previous step.

ckanext/canada/helpers.py
JVickery-TBS: Yeah, I probably should have searched for the usage of this helper anyways. Some of this stuff is pretty old code or even fully unused code at this point. Will check this out more

ckanext/canada/helpers.py
JVickery-TBS: I personally hate the use of `t` alias for toolkit so thank god hahaha. I also don't like `tk`, but that is just my personaly preference because I am weird. (* continues to use underscored named variables in python)

I will just do `toolkit`

@wardi in the topic of `toolkit` do we even need this anymore: https://github.com/ckan/ckantoolkit

ckanext/canada/cli.py
JVickery-TBS: `isodate` can return `None`

ckanext/canada/cli.py
JVickery-TBS: Yeah, I personally have never used this command. Nor had to do anything with suggested datasets to this date. So didn't really put much effort into this one.

I have cleaned up this encoding stuff now. We are on Python3 now so hopefully it is all just fine. We will find out if we ever actually have to use this command I guess.

ckanext/canada/helpers.py
JVickery-TBS: yes it does.

ckanext/canada/helpers.py
JVickery-TBS: Okay yeah this is used in the facets list template:
```
{%- macro _text_key(d) -%}
    {{- h.normalize_strip_accents(d.label) -}}
{%- endmacro -%}
{%- do items.sort(key=_text_key) -%}
```

which makes sense. So yeah it should for sure be something more like:
```
if isinstance(s, bytes):
    s = s.decode('utf-8')
```
and then continue with the normalization. This also is only strictly used in the sort method, so not like it is actually altering any values here.

ckanext/canada/helpers.py
JVickery-TBS: Okay, `t` fixed up here and just used imports from `ckan.plugins.toolkit`

So just the big 'ol question about the `ckantoolkit` repo

ckanext/canada/validators.py
JVickery-TBS: Okay so this validator is only used in the `prop` (Suggested Datasets) dataset type. in the fields `Keywords` and `Subjects`.

This has been changed to `null` now.

ckanext/canada/view.py
JVickery-TBS: Updated to do it how CKAN Core views do it:
```
clean_dict(unflatten(tuplize_dict(parse_params(request.form))))
```

ckanext/canada/helpers.py
wardi: All the compatibility fixes in ckantoolkit were for long forgotten versions of ckan. We can remove it.

wardi (approved):

PR #1554, opened 2025-01-29
Even More Resource Formats

JVickery-TBS:
feat(schema): resource formats;

- Add more resource formats.


JVickery-TBS:
can ignore the linting and typing, will be fixed in a different PR

PR #1555, opened 2025-01-30
Handle Portal Update Exceptions with Click

JVickery-TBS:
feat(cli): click exit 1;

- Click exit code 1 for exceptions.

Need to raise the click exception so that the command gets the proper exit code in the terminal. We handle during the pool, but we do not handle if the worker pool fails to start. So this will handle that.

### DIFF
--- a/changes/1555.changes
+++ b/changes/1555.changes
@@ -1,0 +1,1 @@
+The Portal Update command will send correct exit codes when the worker pool fails.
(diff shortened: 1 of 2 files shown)


JVickery-TBS:
Like the other PR, can ignore the flake and pyright, they are fixed in another PR

wardi (approved):

PR #1556, opened 2025-01-30
New Translations and Template Code for Delete Resource Confirmation

JVickery-TBS:
feat(templates): new code;

- Update code and add translations.

req: https://github.com/open-data/ckan/pull/189

### DIFF
--- a/ckanext/canada/templates/package/confirm_delete_resource.html
+++ b/ckanext/canada/templates/package/confirm_delete_resource.html
@@ -1,10 +1,13 @@
 {% extends "page.html" %}
 
+{% import 'macros/form.html' as form %}
+
 {% block subtitle %}{{ _("Confirm Delete") }}{% endblock %}
 
 {% block primary_content_inner %}
   <div class="col-md-12">
     {% block form %}
+      {% block errors %}{{ form.errors(errors=errors, error_summary=error_summary) }}{% endblock %}
       <p>{{ _('Are you sure you want to delete resource - {name}?').format(name=h.resource_display_name(g.resource_dict)) }}</p>
       <p class="form-actions">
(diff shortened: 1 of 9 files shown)


JVickery-TBS:
annnnd once again, we can ignore the flake8 and pyright in this one...

wardi (approved):

JVickery-TBS:
@wardi okay so some bigger changes in this one now. Because of the issues with the `dataset_errors` variable in https://github.com/open-data/ckan/pull/189 I have just done the ckanext-scheming templating in our plugin here. It is pretty much a copy of the `errors.html` template from ckanext-scheming, but just rigged to work with the new ErrorDict from the open-data/ckan PR.

ckanext/canada/templates/package/confirm_delete_resource.html
```diff
@@ -5,6 +5,19 @@
 {% block primary_content_inner %}
   <div class="col-md-12">
     {% block form %}
+      {% block errors %}
+        {%- set schema = h.scheming_get_dataset_schema(dataset_type) -%}
+        {% if dataset_errors %}
+          {%- snippet 'scheming/snippets/dataset_errors.html',
+            errors=dataset_errors, resource_fields=schema.resource_fields,
+            dataset_fields=schema.dataset_fields, pkg_name=pkg_id,
```
wardi, line 13: instead of using helpers to repeatedly call package_show/resource_show in the snippet can we not pass the whole pkg_dict here?

JVickery-TBS: Yes, just have to modify the DeleteView in the core view to pass in the pkg_dict, it only passes the pkg_id and resource_dict for the current Resource, need names for the dataset and other resources. So I can do that for sure. I hated writing these helper, they felt bad haha

wardi (approved):

PR #1557, opened 2025-02-07
Fix Service Standard Target Met

JVickery-TBS:
fix(pd): service std filter;

- Fixed logic for `target_met`

### DIFF
--- a/bin/filter/filter_service_std.py
+++ b/bin/filter/filter_service_std.py
@@ -43,8 +43,8 @@ def main():
         den = int(row['total_volume']) if row['total_volume'] else 0
 
         # performance = volume_meeting_target / total_volume
-        if num <= 0 or den <= 0:
-            row['performance'] = None  # nand division of zeros
+        if den <= 0:  # NaN calculation
+            row['performance'] = None
         else:
             row['performance'] = max(round(num / den, 4), 0)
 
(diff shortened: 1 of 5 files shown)


bin/filter/filter_service_std.py
```diff
@@ -43,19 +43,25 @@ def main():
...
+            # NaN or 0
+            if not row['performance']:
+                if row['volume_meeting_target'] is not None and row['total_volume'] is not None:
+                    # volume_meeting_target and total_volume are defined,
+                    # either can be 0, target_met is not met
+                    row['target_met'] = 'N'
+                else:
+                    # volume_meeting_target or total_volume are not defined,
+                    # target_met is not possible
+                    row['target_met'] = 'NA'
```
wardi, line 64: Should be able to simplify this:
```suggestion
            if row['volume_meeting_target'] is None or row['total_volume'] is None:
                # volume_meeting_target or total_volume are not defined,
                # target_met is not possible
                row['target_met'] = 'NA'
```

bin/filter/filter_service_std.py
```diff
@@ -43,22 +43,22 @@ def main():
...
 
         # calculate target_met
-        if row['target']:
+        if row['target'] is not None:  # target can be 0
             target = float(row['target'])
 
-            # if no performance(volume_meeting_target/total_volume)
-            # then target_met is not possible
-            if row['performance'] is None:
+            if row['volume_meeting_target'] is None or row['total_volume'] is None:
```
wardi, line 55: If you want `row['target_met'] = 'NA'` when row['total_volume'] is <= 0 then you could instead
```suggestion
            if row['volume_meeting_target'] is None or row['total_volume'] is None or row['performance'] is None:
```

JVickery-TBS: hmmm yeah I think that would make sense. As we could not calculate the performance if the `total_volume` is zero, thus the `target_met` should be `NA`

JVickery-TBS: I don't know anymore. this logic is getting silly and they agreed on it originally. like `N` vs `NA` ugh

PR #1558, opened 2025-02-10
display resource format labels in additional information

RabiaSajjad:


RabiaSajjad:
> Do you think we should also show the label in these pills or nah?

@JVickery-TBS I think it makes sense to show the format label in the pills as it is more readable. I have added this change but I am also confirming with business if they are ok with it.

PR #1566, opened 2025-02-26
Open 3925 simpler search logging

thriuin:
Deploying minor change to logging of the Search `import_data_csv` command that is used in the PD Makefile. Removing the deprecated `--quiet` option. This should substantially reduce the unneeded progress messages in the log files.


RabiaSajjad (approved):
Looks good. Thanks Ross.

PR #1567, opened 2025-03-05
CiteProc Plugin Support

JVickery-TBS:
Implements https://github.com/open-data/ckanext-citeproc

### DIFF
--- a/ckanext/canada/plugin/public_plugin.py
+++ b/ckanext/canada/plugin/public_plugin.py
@@ -215,13 +215,16 @@ class CanadaPublicPlugin(p.SingletonPlugin, DefaultTranslation):
     def get_blueprint(self) -> List[Blueprint]:
         return [canada_views]
 
-    def _canada_cite_map(self, cite_data: DataDict,
-                         pkg_dict: DataDict) -> Dict[str, Any]:
+    def _canada_citation_map(self, cite_data: DataDict,
+                             pkg_dict: DataDict) -> Dict[str, Any]:
         lang = 'en'
         if has_request_context():
             lang = h.lang()
 
-        if pkg_dict.get('org_section'):
+        if(
+          pkg_dict.get('org_section') and
+          lang in pkg_dict['org_section']
+          and pkg_dict['org_section'][lang]):
             cite_data['publisher'] += ' - ' + pkg_dict['org_section'][lang]
 
         cite_data['author'] = []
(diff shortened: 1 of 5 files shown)


ckanext/canada/plugin/public_plugin.py
```diff
@@ -212,6 +215,56 @@ def column_types(self, existing_types: Dict[str, Type[ColumnType]]):
     def get_blueprint(self) -> List[Blueprint]:
         return [canada_views]
 
+    def _canada_citation_map(self, cite_data: DataDict,
+                             pkg_dict: DataDict) -> Dict[str, Any]:
+        lang = 'en'
+        if has_request_context():
+            lang = h.lang()
```
wardi, line 222: does `h.lang()` raise an exception when there's no request context? If so it's considered better style to do this with a `try:` `except:`

ckanext/canada/plugin/public_plugin.py
```diff
@@ -212,6 +215,56 @@ def column_types(self, existing_types: Dict[str, Type[ColumnType]]):
...
+
+        if (
+          pkg_dict.get('org_section') and
+          lang in pkg_dict['org_section']
+          and pkg_dict['org_section'][lang]):
+            cite_data['publisher'] += ' - ' + pkg_dict['org_section'][lang]
```
wardi, line 228: IIRC the suggested style for indenting if conditions is to add an extra (4-space) indent. Also this could be simplified to:

```suggestion
        if pkg_dict.get('org_section', {}).get(lang):
            cite_data['publisher'] += ' - ' + pkg_dict['org_section'][lang]
```

ckanext/canada/plugin/public_plugin.py
```diff
@@ -212,6 +215,56 @@ def column_types(self, existing_types: Dict[str, Type[ColumnType]]):
...
+            cite_data.pop('author', None)
+
+        return cite_data
+
+    # ICiteProcMappings
+    def dataset_citation_map(self, cite_data: DataDict,
+                             pkg_dict: DataDict) -> Tuple[bool, Dict[str, Any]]:
+        cite_data['container_title'] = _(cite_data['container_title'])
+        cite_data = self._canada_citation_map(cite_data, pkg_dict)
+        return False, cite_data
```
wardi, line 259: modifying a passed parameter and *also* returning that parameter is an anti-pattern. I know a number of existing interfaces in ckan do this but we shouldn't repeat this for new interfaces.

JVickery-TBS:
@wardi okay I made some modifications to the ckanext-citeproc plugin to make it better, and updated the code here.

ckanext/canada/plugin/public_plugin.py
```diff
@@ -216,54 +216,47 @@ def get_blueprint(self) -> List[Blueprint]:
         return [canada_views]
 
     def _canada_citation_map(self, cite_data: DataDict,
-                             pkg_dict: DataDict) -> Dict[str, Any]:
+                             pkg_dict: DataDict):
         lang = 'en'
-        if has_request_context():
+        try:
             lang = h.lang()
+        except Exception:
```
wardi, line 223: always better to use the a more specific exception class so that this doesn't hide some other problem

wardi (approved):
otherwise LGTM

ckanext/canada/plugin/public_plugin.py
JVickery-TBS: Yeah the lang helper tries to get current language from the flask environ which does not exist in the CLI of course. Exception is `RuntimeError`

wardi (approved):

PR #1569, opened 2025-03-11
Limit UI List for Resources Sidebar

JVickery-TBS:
feat(templates): limit resource list;

- Toggle resource list.


wardi (approved):
would be great as an upstream contrib

PR #1579, opened 2025-05-20
Support Email Config Option

JVickery-TBS:
feat(dev): support email config;

- Configurized the support email.

### DIFF
--- a/ckanext/canada/helpers.py
+++ b/ckanext/canada/helpers.py
@@ -1113,3 +1113,7 @@ def max_resources_per_dataset() -> Optional[int]:
     max_resource_count = config.get('ckanext.canada.max_resources_per_dataset', None)
     if max_resource_count:
         return int(max_resource_count)
+
+
+def support_email_address() -> str:
+    return config.get('ckanext.canada.support_email_address')
--- a/ckanext/canada/templates/recombinant/snippets/xls_upload.html
+++ b/ckanext/canada/templates/recombinant/snippets/xls_upload.html
@@ -25,7 +25,12 @@
   {% if 'upload_warn' in geno %}
     <div class="col-md-12">
       <div class="alert alert-warning" style="font-weight:bold;">
-        <p>{{ h.recombinant_language_text(geno.upload_warn) | safe }}</p>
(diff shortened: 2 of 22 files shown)


ckanext/canada/templates/recombinant/snippets/xls_upload.html
```diff
@@ -25,7 +25,12 @@ <h4>{{ _("Create and update multiple records") }}</h4>
   {% if 'upload_warn' in geno %}
     <div class="col-md-12">
       <div class="alert alert-warning" style="font-weight:bold;">
-        <p>{{ h.recombinant_language_text(geno.upload_warn) | safe }}</p>
+        {%- set warn_text = h.recombinant_language_text(geno.upload_warn) -%}
+        {%- if '{support}' in warn_text -%}
```
wardi, line 29: this check shouldn't be necessary

ckanext/canada/validators.py
```diff
@@ -233,12 +233,15 @@ def canada_maintainer_email_default(key: FlattenKey,
...
+    Set to ckanext.canada.support_email_address
+    if not given and no contact form given.
+
+    This is an output validator.
     """
     em = data[key]
     cf = data.get(('maintainer_contact_form',), '')
     if (not em or em is missing) and (not cf or cf is missing or cf == '{}'):
-        data[key] = 'open-ouvert@tbs-sct.gc.ca'
+        data[key] = config.get('ckanext.canada.support_email_address')
```
wardi, line 244: we can use the helper here too

ckanext/canada/helpers.py
```diff
@@ -1113,3 +1113,7 @@ def max_resources_per_dataset() -> Optional[int]:
     max_resource_count = config.get('ckanext.canada.max_resources_per_dataset', None)
     if max_resource_count:
         return int(max_resource_count)
+
+
+def support_email_address() -> str:
+    return config.get('ckanext.canada.support_email_address')
```
wardi, line 1119: since we have the config_declaration we can use `[]` instead of `.get()` here (get indicates that it might fail and return None which is not the case)

PR #1580, opened 2025-05-20
Check Registry Domain Instead of Internal Plugin Loaded

JVickery-TBS:
feat(dev): is registry global;

- Moved app global to request context helper.

@wardi I THINK this works how we want to set up our CKAN. So instead of checking for canada_internal to be loaded, just checking the request URI subdomain. As we are still gunna have users login and use the "Registry" via that domain, and the "Public" portal will still be at `open.canada.ca/data/`. But now just both `registry.open.canada.ca` and `open.canada.ca/data/` will do proxy passes to the same uWSGI App in NGINX, so one CKAN instance/config. And we already have some NGINX configs to redirect user login from `/data/` to `registry.open` so I think this makes sense to do?

PS - the helper does `'registry' in subdomain` to support our test and staging environs.

### DIFF
--- a/ckanext/canada/helpers.py
+++ b/ckanext/canada/helpers.py
@@ -58,6 +58,12 @@ GEO_MAP_TYPE_DEFAULT = 'static'
 RELEASE_DATE_FACET_STEP = 100
 
 
+def is_registry_domain() -> bool:
+    uri_parts = urlsplit(request.url)
+    subdomain = uri_parts.netloc.split('.')[0]
+    return 'registry' in subdomain
+
+
 def get_translated_t(data_dict: Dict[str, Any],
                      field: str) -> Tuple[str, bool]:
     '''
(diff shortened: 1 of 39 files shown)


JVickery-TBS:
I think I still need to change some places that are checking `plugin_loaded("canada_internal")`

ckanext/canada/helpers.py
```diff
@@ -58,6 +58,12 @@
 RELEASE_DATE_FACET_STEP = 100
 
 
+def is_registry_domain() -> bool:
+    uri_parts = urlsplit(request.url)
```
wardi, line 62: If some command from the CLI or a background job with no associated request ends up hitting this helper what happens?

ckanext/canada/templates/admin/base.html
```diff
@@ -1,15 +1,17 @@
 {% ckan_extends %}
 
+{% set is_registry = h.is_registry_domain() %}
```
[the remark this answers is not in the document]
wardi: variables are cheap but jinja2 variables are a little messy because of how they sometimes are inherited into blocks and other templates and sometimes not. Calling the helper anywhere it's needed instead of storing the result in a variable makes the templates more consistent and means not having to worry about variable inheritance.

ckanext/canada/helpers.py
wardi: also tests. We wouldn't need all those mocks added.

JVickery-TBS: k catching RuntimeError and returning True for CLI and background processes, added more comments in there

JVickery-TBS:
@wardi are you able to review this again? And confirm if this can be merged prior to making the registry public. I don't think there is anything blocking it? Only thing is that the French domain would not work without this new plugin: https://github.com/open-data/ckanext-language-domains

But otherwise, we are just checking subdomain and not plugin loaded

PR #1582, opened 2025-06-13
Dataset Scheming Form Pages [OPEN-4034]

wardi:
split dataset type into separate pages using scheming form pages and dataset completeness feature


JVickery-TBS:
Had to do some more css and template stuff, mainly for the Add Resource in Draft part. and then some small screen stuff.

Have to do a couple more to polish this up. But this is in staging and working fairly well now.

JVickery-TBS:
Okay small update on this one, I think there is some changes/fixes needed to ckanext-fluent for empty dicts. There are scenarios in which the fluent_text validator outputs empty dicts {} - I don't think it will fully solve the other issues here with some state things. But will work on that further in the future.

As for the templates, I have added the Stages Sidebar back into the Resource edit template as we want it for when creating a dataset, there is the "Add Resource" step:
image

For the above, I was able to just do `{%- if pages and data.get('state') -%}` in the sidebar template. As when Editing a Resource, or adding a new Resource to an existing Dataset, the edit templates will not have the pkg data passed to them. So this solves that, and then also handles the stages not showing up in the Resources' Data Dictionary and Validation Report views.

We still have the issue with the resource and dataset sharing the same field name for date_published:
image

So visually, we are almost there. Logically, there is still some more work to do I think.

JVickery-TBS:
@wardi for now, I have made the use of scheming pages configurable via `ckanext.canada.use_scheming_pages` just so I can make new Qase test cases with screenshots etc for the staging and local combined instance project.

I have added the `.pages` schemas to master for now (just copies of normal schemas) so we can get the changes still tracking here.

EDIT: cuz the new config value controls setting the schemas in update_config, whichever plugin is doing it must be AFTER scheming_datasets (in our case canada_datasets as the subclassed plugin). Of course, in the combined instance almost all logic outside of Jinja, Helpers, and i18n POs are inside of canada_logic which is always the last loaded plugin.

EDIT 2: our current production setup actually does not support canada_public being below recombinant. So I think we can just check the config value and assert for now...

PR #1583, opened 2025-06-18
Rework Code Base for Single Instance

JVickery-TBS:


ckanext/canada/templates/base.html
```diff
@@ -56,10 +56,7 @@
 
 {%- block custom_styles -%}
   {% asset 'canada_public/css' %}
-  {% if h.is_registry_domain() %}
-    {# Remove when WET v4.0.21 is out #}
-    {% asset 'canada_internal/css' %}
-  {% endif %}
+  {% asset 'canada_internal/css' %}
```
wardi: These could be merged at some point

wardi (commented):
looks good, just a few minor suggestions

ckanext/canada/templates/user/list.html
```diff
@@ -59,6 +59,11 @@ <h2 class="wb-inv">{{ _('Users') }}</h2>
                       {% else %}
                         <span class="text-muted"><span class="fa fa-unlock-alt">&nbsp;</span></span>
                       {% endif %}
+                      {% if user.default_dataset_visibility == 'public' or (user.User and user.User.plugin_extras and user.User.plugin_extras['default_dataset_visibility'] == 'public') %}
+                        <span class="text-muted" title="{{- _('Creates Public Datasets') -}}"><i class="fa fa-eye"></i>&nbsp;</span>
+                      {% else %}
+                        <span class="text-muted" title="{{- _('Creates Private Datasets') -}}"><i class="fa fa-eye-slash"></i>&nbsp;</span>
```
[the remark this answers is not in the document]
JVickery-TBS: Yessss that makes more sense for our sysadmins

ckanext/canada/schemas/presets.yaml
```diff
@@ -4291,3 +4302,23 @@ presets:
...
+
+- preset_name: ckanapi_dump_datastore_fields
+  values:
+    field_name: datastore_fields
+    form_snippet: null
+    display_snippet: null
+    required: false
+    validators: ignore
+    form_panel: _exclude_from_form
+    output_validators: canada_output_none
```
wardi: It would be much better for ckanapi to remove these fields before passing them to ckan, since they're not real fields and it's the tool generating them in the first place

PR #1586, opened 2025-06-26
fixed dropdown

gabmor38:
Fixed language in new user form dropdown

### DIFF
--- a/ckanext/canada/templates/user/new_user_form.html
+++ b/ckanext/canada/templates/user/new_user_form.html
@@ -14,7 +14,7 @@
     {% call form.input("email", id="field-email", label=_("Email"), type="email", value=data.email, error=errors.email, attrs=valid_dict2, is_required=true, classes=["control-medium"]) %}
       {{ form.info(_('Example: joe.bloggs@tbs-sct.gc.ca')) }}
     {% endcall %}
-    {% set format_attrs={ 'oninvalid': "setCustomValidity('"+_("Please fill out this field.")+"')", 'onchange':"setCustomValidity('')", 'data-module': 'autocomplete', 'data-module-source': '/organization/autocomplete?q=?', 'data-module-key': 'title', 'data-module-label': 'title', 'data-module-dropdownClass': 'org-select2-dropdown', 'data-module-containerClass': 'org-select2-container', 'data-module-tags': 'false', 'data-module-createtags': 'false', 'class':'control-medium'} %}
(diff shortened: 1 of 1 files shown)


ckanext/canada/templates/user/new_user_form.html
```diff
@@ -14,7 +14,7 @@
...
       {{ form.info(_('Example: joe.bloggs@tbs-sct.gc.ca')) }}
     {% endcall %}
-    {% set format_attrs={ 'oninvalid': "setCustomValidity('"+_("Please fill out this field.")+"')", 'onchange':"setCustomValidity('')", 'data-module': 'autocomplete', 'data-module-source': '/organization/autocomplete?q=?', 'data-module-key': 'title', 'data-module-label': 'title', 'data-module-dropdownClass': 'org-select2-dropdown', 'data-module-containerClass': 'org-select2-container', 'data-module-tags': 'false', 'data-module-createtags': 'false', 'class':'control-medium'} %}
+    {% set format_attrs={ 'oninvalid': "setCustomValidity('"+_("Please fill out this field.")+"')", 'onchange':"setCustomValidity('')", 'data-module': 'autocomplete', 'data-module-source': h.lang()~ '/organization/autocomplete?q=?', 'data-module-key': 'title', 'data-module-label': 'title', 'data-module-dropdownClass': 'org-select2-dropdown', 'data-module-containerClass': 'org-select2-container', 'data-module-tags': 'false', 'data-module-createtags': 'false', 'class':'control-medium'} %}
```
wardi: it's better to use `h.url_for('canada.organization_autocomplete')` for generating this url so that the language and site path are both included.

ckanext/canada/templates/organization/snippets/organization_item.html
```diff
@@ -4,7 +4,7 @@
   {% set html_title = organization.description|truncate(80, true) %}
 {% endif %}
 <dt style="width:auto;"><a href="{{ url }}" title="{{ html_title }}">
-  {{ h.split_piped_bilingual_field(organization.display_name, h.lang())|truncate(80, true) }}</a>
+ _{{ h.split_piped_bilingual_field(organization.display_name, h.lang())|truncate(80, true) }}</a>
```
wardi, line 7: add an underline before the org title?

gabmor38: I removed it, I didn't mean to add this change to the repo

JVickery-TBS (changes requested):
Just the above usage for `h.url_for`

JVickery-TBS (changes requested):
@gabmor38 for the 3 new instances of url_for you will want to use the Flask blueprint name for the endpoint. So instead of `/organization/autocomplete` use `h.url_for('canada.organization_autocomplete', q='?')`

url_for takes first parameter as the flask blueprint name which will resolve to the url route even if we change it in the future. And then all other parameters after are the blueprint/url args

JVickery-TBS:
@gabmor38 as for the tests failing, it looks like it is just all the frontend Recombinant webform tests. Because of one of the change shere in the org select for that:
```
>   data-module-source="{{ h.url_for(/organization/autocomplete?q=?) }}"
E   jinja2.exceptions.TemplateSyntaxError: unexpected '/'

ckanext/canada/templates/recombinant/snippets/org_select.html:21: TemplateSyntaxError
```
but should be fixed if using the `canada.organization_autocomplete`

PR #1595, opened 2025-08-01
AI Strategy Tracker 2025

RabiaSajjad:


ckanext/canada/tables/aistrategy.yaml
```diff
@@ -0,0 +1,818 @@
...
+    obligation: Mandatory
+    excel_required: true
+    form_required: true
+    format_type: Controlled List
+    datastore_type: text
+    validation:
+      en: Must match expected completion options in the controlled list.
+      fr: Doit correspondre aux options d'achèvement prévues de la liste contrôlée.
+    occurrence: Single
+    excel_full_text_choices: true
```
wardi, line 64: I suggest setting this to false here for easier data entry

ckanext/canada/tables/aistrategy.yaml
```diff
@@ -0,0 +1,818 @@
...
+      en: The key action described in the AI Strategy
+      fr: L'action clé décrite dans la stratégie en matière d'IA
+    obligation: Mandatory
+    excel_required: true
+    form_required: true
+    format_type: Controlled List
+    datastore_type: text
+    validation:
+      en: Must match predefined action statements.
+      fr: Doit correspondre à des déclarations d'action prédéfinies.
```
wardi, line 138: adapt the nap5 excel_choice_range_formula and excel_error_formula values for this field so that the drop-downs are hierarchical and errors correctly displayed

wardi: (same for sub-action and activity)

wardi: also add:
```yaml
form_attrs:
  cascading_select_field: priority_area
```
and similar for sub-action and activity

ckanext/canada/tables/aistrategy.yaml
wardi: this is a style thing though, see what you prefer in the generated excel template

ckanext/canada/tables/aistrategy.yaml
```diff
@@ -0,0 +1,818 @@
...
+        errors := errors || required_error(NEW.priority_area, 'priority_area');
+        errors := errors || choice_error(NEW.priority_area, {priority_area}, 'priority_area');
+
+        errors := errors || required_error(NEW.key_action, 'key_action');
+        errors := errors || choice_error(NEW.key_action, {key_action}, 'key_action');
+
+        errors := errors || required_error(NEW.sub_action, 'sub_action');
+        errors := errors || choice_error(NEW.sub_action, {sub_action}, 'sub_action');
+
+        errors := errors || required_error(NEW.activity, 'activity');
```
wardi, line 767: for key_action, sub_action and activity adapt the nap5 sql validation that limits the choices available based on the previous selections, e.g.
```sql
errors := errors || choice_error(NEW.activity,
  array(SELECT unnest from unnest({activity}) where left(unnest, length(NEW.sub_action)) = NEW.sub_action),
  'activity')
```

ckanext/canada/tables/aistrategy.yaml
```diff
@@ -0,0 +1,818 @@
...
+  - update_record_modified_created_trigger
+
+  examples:
+    record:
+      ref_number: "TBS-001"
+      reporting_period: "2025-Q3"
+      priority_area: "A"
+      key_action: "A.1"
+      sub_action: "A.1.1"
+      activity: "1.3.1"
```
wardi, line 805:
```suggestion
      activity: "A.1.3.1"
```

wardi: add a basic test module that covers loading the example record at least, even better would be testing specific error cases. See test_nap5.py for minimal tests and test_service.py for some extensive ones.

ckanext/canada/templates/home/quick_links.html
```diff
@@ -135,6 +135,9 @@ <h2 class="h5">{{ _('Proactive Publication') }}</h2>
             {{ quick_link_pd('wrongdoing') }}
             {{ quick_link_pd('contractsa') }}
             {{ quick_link_pd('ati') }}
+            {% if org_name == 'tbs-sct' %}
+              {{ quick_link_pd('aistrategy') }}
+            {% endif %}
```
wardi, line 140: add to the menu navigation too?

ckanext/canada/tables/aistrategy.yaml
```diff
@@ -0,0 +1,818 @@
...
+
+  # Dev: req dynamic field
+  - datastore_id: user_modified
+    label: User Last Modified Record
+    import_template_include: false
+    visible_to_public: false
+    datastore_type: text
+    preview_class: bg-info
+
+  datastore_primary_key: [ref_number, reporting_period]
```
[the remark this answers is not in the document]
wardi: I'm curious why this change was rolled back. This will mean more data to enter and allows duplicate entries for the same indicator and reporting period in the published data and tracker web page.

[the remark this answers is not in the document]
RabiaSajjad: We looked at the sample data and found that some rows were overwriting because TBS will be entering data for multiple departments leading the same activity in a reporting period. Having a unique id mitigates this issue. Since only TBS will be entering the data, it shouldn't be too complex to maintain the reference number for the records.

ckanext/canada/tables/aistrategy.yaml
wardi: This should fix the example data test failure, the fix below is just for the docs

PR #1627, opened 2025-10-03
Replane Tool Example

JVickery-TBS:
feat(pd): aistrategy updates;

- Used new replane tool.

Compare to the human done one: https://github.com/open-data/ckanext-canada/pull/1626


JVickery-TBS:
This would also help us normalize line lengths in the YAMLS too

PR #1668, opened 2026-04-09
Organization Validation for Service Inventory

JVickery-TBS:
We are able to consistently use the releases from the SI github to generate the CSV ref data we need. We have commands which load it into the tables which we define in an sql script in here. The part we were tripped on was relation between Service IDs and Program IDs, but there is no relation that is being defined or maintained currently so don't have to worry about that.

Just extended the temp datastore table to add an optional org_name to it, which we then use in the triggers. The Service ID one works fine, need to make the error message for it still.

The Program IDs one is gunna be interesting as Program IDs is a multiple select.


ckanext/canada/plugin/internal_plugin.py
```diff
@@ -62,7 +63,20 @@ def configure(self, config: 'CKANConfig'):
...
+            resource_id = data_dict.get('resource_id')
+            if resource_id:
+                res = model.Resource.get(resource_id)
+                if res and res.name in get_resource_names():
+                    try:
+                        pkg_dict = p.toolkit.get_action('package_show')(
+                            {'ignore_auth': True}, {'id': res.package_id})
+                        org_name = pkg_dict.get('organization', {}).get('name', None)
+                    except (ObjectNotFound, NotAuthorized):
+                        pass
```
wardi: package_show makes db queries and a solr call to respond with the complete package dict. If all we need is the org name and we're already accessing the models directly then we could use:
```suggestion
                if res and res.name in get_resource_names():
                    org_name = res.package.owner_org
```

Or we can query the org name and the resource name as part of the same db call and skip the model objects. Something like:
```python
org_name, res_name = session.query(
    model.Group.name, model.Resource.name,
).join(
    model.Resource.id == resource_id,
    model.Package.id == model.Resource.package_id,
    model.Group.id == model.Package.owner_id,
).one()
```
Maybe with some `state == 'active'`s in the filters there to be safe.

JVickery-TBS: thanks yeah I was trying to get the cheapest, cheekiest thing to do for this as it would fire off on all datastore upsert calls

bin/service_generate_reference_data.py
```diff
@@ -0,0 +1,218 @@
...
+                continue
+    assert org_variants_uri
+    assert program_ids_uri
+    assert service_ids_uri
+
+    # compile list of Open Canada org abbreviations and Service Inventory org_ids
+    org_id_abbr_map = {}
+    with requests.get(org_variants_uri, stream=True) as response:
+        response.encoding = 'utf-8-sig'
+        c = csv.DictReader((l.decode('utf-8-sig') for l in response.iter_lines()))
```
wardi: Our CSV reading is is a little odd. We're expecting a UTF-8 BOM on every line which doesn't make sense.

`io.TextIOWrapper` was the thing I was trying to remember last week. This is cleaner:

```suggestion
        c = csv.DictReader(TextIOWrapper(response.raw, 'utf-8-sig'))
```

(and same suggestion for the other DictReaders)

ckanext/canada/tables/service.yaml
```diff
@@ -1600,16 +1612,31 @@ resources:
...
+        refval := (
+          SELECT array_agg(service_id) FROM "ref_service_service_ids"
+          WHERE org_years ? org_name
+          AND service_id = NEW.service_id
+        );
+        IF refval IS NULL THEN
+          errors := errors || ARRAY[['service_id', {service_id_org_limit_error} || '' || org_name || '' || NEW.service_id]];
+        ELSE
+          NEW.service_id := refval[1];
+        END IF;
```
wardi: Would something like this work instead?
```suggestion
        IF NOT EXISTS(
          SELECT FROM "ref_service_service_ids"
            WHERE org_years ? org_name
            AND service_id = NEW.service_id
        ) THEN
          errors := errors || ARRAY[['service_id', {service_id_org_limit_error} || '' || org_name || '' || NEW.service_id]];
        END IF;
```

ckanext/canada/tables/service.yaml
```diff
@@ -1644,10 +1671,17 @@ resources:
...
         NEW.client_target_groups := crval.clean;
 
-        errors := errors || required_error(NEW.program_id, 'program_id');
-        crval := choices_clean_error(NEW.program_id, {program_id}, 'program_id');
-        errors := errors || crval.error;
-        NEW.program_id := crval.clean;
+        refval := (
+          SELECT array_agg(program_id) FROM "ref_service_program_ids"
+          WHERE org_years ? org_name
+          AND program_id = ANY(NEW.program_id)
```
wardi: It's good to normalize the order of program_ids (add an ORDER BY here) so that users providing the same list of program ids in a different order isn't treated as a change, and so that records in the db with the same program ids always have the program ids in the same order.

ckanext/canada/tables/service.yaml
```diff
@@ -2589,16 +2636,30 @@ resources:
...
 
+        refval := (
+          SELECT array_agg(service_id) FROM "ref_service_service_ids"
+          WHERE org_years ? org_name
+          AND service_id = NEW.service_id
+        );
+        IF refval IS NULL THEN
+          errors := errors || ARRAY[['service_id', {service_id_org_limit_error} || '' || org_name || '' || NEW.service_id]];
+        ELSE
+          NEW.service_id := refval[1];
```
wardi: see my service_id comment above

wardi:
just those minor things otherwise LGTM

bin/service_generate_reference_data.py
JVickery-TBS: can't use buffers cuz the response is a Stream. But `response.iter_lines(decode_unicode=True)` seems to work.

ckanext/canada/triggers.py
```diff
@@ -749,3 +749,43 @@ def update_triggers():
...
+            {'argname': 'max', 'argtype': 'int4', 'argdefault': 'NULL'}],
+        rettype='_text',
+        definition='''
+    DECLARE
+        min_val int4 := min;
+        max_val int4 := max;
+    BEGIN
+        IF min_val IS NULL THEN
+            min_val := 2005;
+        END IF;
```
wardi: This should work, right?
```suggestion
            {'argname': 'min', 'argtype': 'int4', 'argdefault': 2025},
            {'argname': 'max', 'argtype': 'int4', 'argdefault': 'NULL'}],
        rettype='_text',
        definition='''
    DECLARE
        min_val int4 := min;
        max_val int4 := max;
    BEGIN
```

ckanext/canada/tables/service.yaml
```diff
@@ -1600,16 +1578,28 @@ resources:
...
         errors text[][] := '{{}}';
+        org_name text NOT NULL := (SELECT org_name FROM datastore_user LIMIT 1);
         crval RECORD;
+        refval text[];
+        referr text[];
       BEGIN
+        NEW.fiscal_yr := trim(both E'\t\n\x0b\x0c\r \xc2\xa0' from NEW.fiscal_yr);
         errors := errors || required_error(NEW.fiscal_yr, 'fiscal_yr');
-        errors := errors || choice_error(NEW.fiscal_yr, {fiscal_yr}, 'fiscal_yr');
+        errors := errors || fiscal_year_error(NEW.fiscal_yr, 'fiscal_yr'::text, 2014);
```
wardi: I wonder why `::text` is necessary here. Are literal strings in postgresql not text?

wardi (approved):

ckanext/canada/tables/service.yaml
JVickery-TBS: yeah not sure still with this one, it works locally but the tests are just failing on it in github

ckanext/canada/triggers.py
JVickery-TBS: yeah might as well just use 2005 as the default

ckanext/canada/tables/service.yaml
JVickery-TBS: riiiggght the argdefault core code was not merged in our ckan fork yet haha. Okay hopefully these all go then.

JVickery-TBS:
TODO: update any tests for test_service.py and its make tests if needed.

bin/filter/filter_service.py
```diff
@@ -79,17 +96,33 @@ def process_row(row: Dict[str, Any]) -> Dict[str, Any]:
...
     for id in program_ids:
+        # TODO: update suffix labels
+        # handle suffix "-INV"
+        #   en: (Invalid for Organization EN)
+        #   fr: (Invalid for Organization FR)
+        suffix_en = ''
+        suffix_fr = ''
+        if id.endswith('-INV'):
+            suffix_en = ' (Invalid for Organization EN)'
+            suffix_fr = ' (Invalid for Organization FR)'
```
wardi: Do we need a FIXME to remember to get translated strings here?

bin/filter/filter_service.py
```diff
@@ -79,17 +96,33 @@ def process_row(row: Dict[str, Any]) -> Dict[str, Any]:
...
         if id not in PROGRAM_IDS:
+            row['program_name_en'].append('"Unknown Program EN%s"' % suffix_en)
+            row['program_name_fr'].append('"Unknown Program FR%s"' % suffix_fr)
             continue
         # NOTE: we add double quotes as Program Names can have
         #       single quotes and commas in them
-        row['program_name_en'].append('"%s"' % PROGRAM_IDS[id]['en'])
-        row['program_name_fr'].append('"%s"' % PROGRAM_IDS[id]['fr'])
+        row['program_name_en'].append('"%s%s"' % (PROGRAM_IDS[id]['en'], suffix_en))
+        row['program_name_fr'].append('"%s%s"' % (PROGRAM_IDS[id]['fr'], suffix_fr))
```
wardi: Maybe a little cursed, but you are allowed to use the same types of quotes inside f-strings that you use outside.
```suggestion
        row['program_name_en'].append(f'"{PROGRAM_IDS[id]['en']}{suffix_en}"')
        row['program_name_fr'].append(f'"{PROGRAM_IDS[id]['fr']}{suffix_fr}"')
```

bin/service_generate_reference_data.py
```diff
@@ -0,0 +1,233 @@
...
+SERVICE_ID_HEADERS = ['service_id', 'label_en', 'label_fr', 'org_years']
+
+PROGRAM_ID_OUTPUT_FILE = os.path.join(
+    os.path.split(__file__)[0],
+    '../ckanext/canada/tables/references/data/ref_service_program_ids.csv')
+PROGRAM_ID_HEADERS = ['program_id', 'label_en', 'label_fr', 'org_years']
+
+REQUEST_HEADERS = {'User-Agent': 'CKAN/open-gov/service/gen_ref_data'}
+
+WHITE_SPACE_SUBS = re.compile(r'\xa0|\t|\n|\s+')
```
wardi: I think you mean
```suggestion
WHITE_SPACE_SUBS = re.compile(r'[\xa0\s]+')
```
`\t` and `\n` are covered by `\s` and you want to allow any number of repeats of all of them, not only `\s`

bin/service_generate_reference_data.py
```diff
@@ -0,0 +1,233 @@
...
+                continue
+    assert org_variants_uri
+    assert program_ids_uri
+    assert service_ids_uri
+
+    # compile list of Open Canada org abbreviations and Service Inventory org_ids
+    org_id_abbr_map = {}
+    with requests.get(org_variants_uri, stream=True) as response:
+        response.encoding = 'utf-8-sig'
+        c = csv.DictReader(response.iter_lines(decode_unicode=True))
```
wardi, line 99: if we've manually set the encoding do we need detect_unicode=True? If so that's a strange API

ckanext/canada/logic.py
```diff
@@ -302,24 +306,36 @@ def datastore_create_temp_user_table(context: Context,
...
             '''.format(
                 drop_statement=' ON COMMIT DROP' if drop_on_commit else '',
                 username=literal_string(username),
-                sysadmin='TRUE' if is_sysadmin(username) else 'FALSE'))
+                sysadmin='TRUE' if is_sysadmin(username) else 'FALSE',
+                cli_importing='TRUE' if 'recombinant_import' in
+                contextual_flags else 'FALSE',
+                org_name=literal_string(org_name) if org_name else None,
+                flag_definitions=(',' if contextual_flags else '') +
+                ', '.join('%s boolean' % m for m in contextual_flags),
```
wardi: we'd want an `identifier()` here wrapping column names, but as we discussed it would be even better to use a `text[]` field for flags

ckanext/canada/triggers.py
```diff
@@ -265,7 +265,8 @@ def update_triggers():
         rettype='_text',
         definition='''
     BEGIN
-        IF value IS NOT NULL AND value <> '' AND LENGTH(value) > max_chars THEN
+        IF value IS NOT NULL AND value <> '' AND LENGTH(REGEXP_REPLACE(
+        value, E'[\r\n]', '', 'g')) > max_chars THEN
```
wardi: as discussed we want to include all characters when computing length

ckanext/canada/model.py
```diff
@@ -15,6 +15,74 @@
...
 
 
+# type_ignore_reason: incomplete typing
+class RefData(Base):  # type: ignore
+    __tablename__ = 'ref_data'
+
+    table_name = Column(Unicode, primary_key=True)
+    last_sync = Column(DateTime, nullable=False,
+                       default=datetime.datetime.now(datetime.timezone.utc))
+    file_hash = Column(Unicode, nullable=False)
```
wardi: if we're only ever storing sha265 hexdigest values here it would be better to call the column sha256 or similar

bin/filter/filter_service.py
JVickery-TBS: there is a todo above, but can change to fixme

bin/service_generate_reference_data.py
JVickery-TBS: sweet thanks, this caught one: `BWD05,High Performance  Computing Solution,Solution` -> `Performance Computing` double space between "Performance  Computing"

bin/service_generate_reference_data.py
JVickery-TBS: yesss it is a weird API, the response.encoding is not passed to iter_lines it was a really really annoying thing to deal with.

JVickery-TBS: the response.encoding part is just for the possible BOMs for csv.DictReader to understand.

ckanext/canada/triggers.py
JVickery-TBS: reverted, and just handling newline normalization elsewhere.

ckanext/canada/logic.py
JVickery-TBS: changed to text[] which is a lot better.

JVickery-TBS:
@wardi I think this is ready for another review now.

ckanext/canada/assets/datatables/pd_datatables.js
```diff
@@ -237,12 +246,14 @@ function load_pd_datatable(CKAN_MODULE, HAS_TRANSLATIONS){
         let str = _data.toString();
         let htmlStr = $($.parseHTML(str)).text();
         if( str.length < _cutoff || htmlStr.length < _cutoff ){
+          _data = _data.replaceAll(/\r?\n/g, '<br>');
```
wardi, line 304: This is spooky. Are we rendering all HTML tags that might appear in the data?

wardi:
otherwise LGTM

ckanext/canada/assets/datatables/pd_datatables.js
JVickery-TBS: mmmm good question. I'll make a ticket to look into that. this is in production already and the firewall will catch any JS scripts trying to be posted to the registry. so for now, not a critical security thing - but yeah will go look into it.

PR #1672, opened 2026-04-22
minor change to nap6 commitment

RabiaSajjad:


changes/1672.changes
```diff
@@ -0,0 +1 @@
+Small change to commitment C04.5.
```
JVickery-TBS: @RabiaSajjad are u able to specify that this is for NAP6 in the change log? thanksssss

JVickery-TBS (changes requested):

PR #1676, opened 2026-05-30
fixes for loading service.csv

wardi:
some changes I needed for loading service.csv

### DIFF
--- a/ckanext/canada/tables/service.yaml
+++ b/ckanext/canada/tables/service.yaml
@@ -1600,14 +1600,6 @@ resources:
           errors := errors || ARRAY[['service_id', {service_id_org_limit_error} || '' || org_name || '' || NEW.service_id]];
         END IF;
 
-        errors := errors || required_error(NEW.service_name_en, 'service_name_en');
-        NEW.service_name_en := trim_lead_trailing(NEW.service_name_en);
-        errors := errors || max_char_error(NEW.service_name_en, 350, 'service_name_en');
-
-        errors := errors || required_error(NEW.service_name_fr, 'service_name_fr');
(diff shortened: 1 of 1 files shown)


ckanext/canada/tables/service.yaml
```diff
@@ -1638,7 +1630,6 @@ resources:
           SELECT array_agg(program_id) FROM "ref_service_program_ids"
           WHERE org_years ? org_name
           AND program_id = ANY(NEW.program_id)
-          ORDER BY program_id
```
wardi, line 1641: Maybe this works on newer postgres but it doesn't on my 16.9

wardi:
service.csv error report:
[output.csv](https://github.com/user-attachments/files/28428490/output.csv)

JVickery-TBS:
sweet that is not too many. so we can get this to the Policy Team and hopefully they can engage to get these fixed by September

ckanext/canada/tables/service.yaml
JVickery-TBS: yeaahh we added that cuz u mentioned that we didn't want to take a different ordering of Program IDs as an actual update to the field/record. But I think we can put the `ORDER BY` inside of the `array_agg`? r u able to try that to see if that works?

wardi:
Here are the service-std errors I get: 
[output2.csv](https://github.com/user-attachments/files/28672017/output2.csv)

This needs one more change in recombinant to prevent failures on load with FK errors: https://github.com/open-data/ckanext-recombinant/commit/90aa79dbcea11adb0cba9bb61777c828d1510010

ckanext/canada/tables/service.yaml
wardi: sure I'll look at updating the ordering for better compatibility

PR #1678, opened 2026-06-05
update url for fgp viewer

RabiaSajjad:


JVickery-TBS (approved):
@RabiaSajjad okay so openmaps would be https://open.canada.ca/openmaps ?? I think this would work then as our nginx rewrites the portal /data to / and then the ckan helper would resolve to /openmaps right?

I guess I will take note that in the Combined CKAN instance, I will have to change the `if fgp_url` to `if not h.is_registry_domain() and fgp_url` so that the openmaps button does not display in the Registry.

PR #1679, opened 2026-06-09
Resource View Embed Host List

JVickery-TBS:
Bringing back the resource view embed button, adds a list into the modal to display which hosts are allowed to embed our site contents.

### DIFF
--- a/ckanext/canada/helpers.py
+++ b/ckanext/canada/helpers.py
@@ -1173,3 +1173,7 @@ def mail_to_with_params(email_address: str, name: str,
 
 def get_inline_script_nonce() -> str:
     return str(request.environ.get('CSP_NONCE', ''))
+
+
+def get_allowed_frame_hosts() -> Optional[list]:
+    return config.get('ckanext.canada.allowed_frame_hosts')
--- a/ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
+++ b/ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PROJECT VERSION\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2026-03-19 17:41+0000\n"
+"POT-Creation-Date: 2026-06-09 14:38+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language: fr\n"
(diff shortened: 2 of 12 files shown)


ckanext/canada/templates/package/snippets/resource_view.html
```diff
@@ -17,3 +30,33 @@
   {%- endif -%}
 </p>
 {% endblock %}
+
+{% block resource_view_bottom %}
+  {% set allowed_hosts = ['canada.ca', '*.canada.ca', '*.gc.ca'] %}
```
wardi: these would make sense as a configuration option since the allowed list isn't tied to any code in this repo, right?

ckanext/canada/templates/package/snippets/resource_view.html
```diff
@@ -17,3 +30,33 @@
...
+      <div class="modal-content">
+        <div class="modal-header">
+          <h3 class="modal-title">{{ _("Embed resource view") }}</h3>
+          <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="{{ _('Close') }}"></button>
+        </div>
+        <div class="modal-body">
+          <p class="embed-content">{{ _("You can copy and paste the embed code into a CMS or blog software that supports raw HTML") }}</p>
+          <p class="embed-content">
+            {{ _("Only certain websites are allowed to embed content from open.canada.ca:") }}
+          <pre>{%- for h in allowed_hosts -%}{{- h -}}<br/>{%- endfor -%}</pre>
```
wardi: maybe a list with <code>s instead of a <pre> block with <br/>s? That's more semantic and should make it easier to tweak with CSS, right?

ckanext/canada/templates/snippets/citeproc_button.html
```diff
@@ -0,0 +1,7 @@
+{% ckan_extends %}
+
+{% block citeproc_button %}
+  {% if not h.is_registry_domain() %}
+    {{ super() }}
+  {% endif %}
+{% endblock %}
```
wardi, line 7: instead of overriding the snippet why not add the `not h.is_registry_domain()` check to https://github.com/open-data/ckanext-canada/pull/1679/changes#diff-748ec90e95b84b2c2e52c8bf7cc57bbb9d7aa264bf8e425df0f0369b9d2d1075R146 ?

JVickery-TBS: there are places in the citeproc plugin which the button snippet is added to some core CKAN templates, like the Resource Read one.

ckanext/canada/i18n/fr/LC_MESSAGES/ckanext-canada.po
```diff
@@ -3974,6 +3974,43 @@ msgstr "Etes-vous sûr de vouloir supprimer cette ressource?"
...
+msgid "Embed"
+msgstr "Embarquer sur un site"
+
+#: ckanext/canada/templates/package/snippets/resource_view.html:40
+msgid "Embed resource view"
+msgstr "Incorporer cette visualisation de ressource"
+
+#: ckanext/canada/templates/package/snippets/resource_view.html:41
+msgid "Close"
+msgstr ""
```
wardi, line 3987: missed one or do we inherit this translation?

ckanext/canada/helpers.py
```diff
@@ -1173,3 +1173,7 @@ def mail_to_with_params(email_address: str, name: str,
 
 def get_inline_script_nonce() -> str:
     return str(request.environ.get('CSP_NONCE', ''))
+
+
+def get_allowed_frame_hosts() -> Optional[list]:
```
wardi, line 1178: the config declaration should guarantee the type for us, so maybe
```suggestion
def get_allowed_frame_hosts() -> list[str]:
```

PR #1680, opened 2026-06-09
single Page Package Type Display

JVickery-TBS:
I personally have no idea what I am looking at when I am in the single page for dataset and resources. So adding the package type here really helps determine if this is Open Data or Open Info


ckanext/canada/templates/package/snippets/package_type_label.html
```diff
@@ -0,0 +1,28 @@
+{#
+Label/pill style display for a Package Type
+
+pkg_dict  - package dictionary
+for_page  - wraps html in a <span class="lb-page-display"></span>
```
wardi, line 5: why is this a parameter it it's always true?

ckanext/canada/templates/package/read.html
```diff
@@ -2,20 +2,65 @@
 
 {% set client_lang = h.lang() %}
 {% set title, machine_translated = h.get_translated_t(pkg_dict, 'title') %}
+{% if h.is_registry_domain() %}
+  {% set portal_link = h.portal_url() ~ "/data/" ~ client_lang ~ "/" ~ pkg_dict.get('type', 'dataset') ~ "/" ~ pkg_dict.name %}
+  {% set portal_status_alerts = [] %}
+  {% set is_type_for_alerts = pkg_dict.get('type') == 'dataset' or pkg_dict.get('type') == 'info' %}
+  {% set do_not_show_alerts = pkg_dict.state.startswith('draft') or pkg_dict.state.startswith('deleted') %}
+  {% set no_ready_to_publish = pkg_dict['ready_to_publish'] == 'false' or not pkg_dict['ready_to_publish'] %}
+  {% set no_has_approval = pkg_dict['imso_approval'] == 'false' or not pkg_dict['imso_approval'] %}
```
wardi: I definitely prefer variables like this to be positive rather than negative so that we never end up with awkward syntax like `{% if not no_ready_to_publish %}`

ckanext/canada/templates/package/read.html
```diff
@@ -2,20 +2,65 @@
...
+  {% endif %}
+  {% if no_has_approval %}
+    {% do portal_status_alerts.append({
+        'label': _('Action required:'),
+        'sub_label': _('Seek out departmental approval and mark as approved to continue.'),
+        'alert_class': 'warning',
+      })
+    %}
+  {% endif %}
+  {% if not no_ready_to_publish and not no_has_approval and queued_for_publishing %}
```
wardi: oh look, it happened!

ckanext/canada/templates/package/snippets/package_type_label.html
JVickery-TBS: its used in `snippets/package_item.html` with for_page=False

ckanext/canada/templates/package/read.html
JVickery-TBS: k all variables and logic updated to be positive. and specified approval var with imso

PR #1682, opened 2026-06-16
update permissions for adminaircraft menu link

RabiaSajjad:


ckanext/canada/templates/home/quick_links.html
```diff
@@ -196,7 +196,10 @@ <h2 class="h5">{{ _('Proactive Publication') }}</h2>
               </ul>
             <li>
 
-            {% if h.canada_check_access('98abeb62-7c76-4dfb-a134-1551f55ede55') %}
+            {% set orgs = h.organizations_available('read') %}
```
JVickery-TBS: we already have the `set orgs` in the same block https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/templates/home/quick_links.html#L61 so should not need to redefine and call the helper here.

ckanext/canada/templates/snippets/cdts/header.html
```diff
@@ -132,6 +132,15 @@
                 "subhref": {{ h.url_for('info_new', collection='parliament_committee_deputy') | tojson }},
                 "subtext": {{ _('Parliamentary Committee appearances for deputy heads') | tojson }},
               },
+              {% set orgs = h.organizations_available('read') %}
```
JVickery-TBS: same thing as above, we already have `set orgs` in the same if block https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/templates/snippets/cdts/header.html#L42 so you can probably bring the original one up to the top of the if block context.

JVickery-TBS (changes requested):
@RabiaSajjad just the above comments for var reuse.

JVickery-TBS (approved):

PR #1689, opened 2026-07-14
Persistant User plugin_extras & Test Coverage

JVickery-TBS:
feat(tests): coverage;

- User schema test coverage;
- Fix user schema for custom extra fields.

This code makes more sense when there are more user fields like in the single-instance and staging branches. As fields are omitted across the site in the data_dict

EDIT: just more explanation... so for instance the 'default_publish_visibility' for the combined instance only displays for sysadmin users in the Webform. So if a normal user changes their name, email, or bio, then that field would not get POSTed to the action endpoint and thus would be removed/default validator'd into the plugin_extras. Because we are kind of just piggy-backing off of the plugin_extras core field here, so we don't have to actually make DB Model for our custom user fields. Similarly, the user_opt_in_feature__ fields are only POSTed via the JS Module, so if a user updated their bio in the webform, it would remove/default validator this plugin extra too. So just had to add some extra code in here to fix that. Also, there is a check for if schema is in the context now, as the ckanext-security extension also extends the schema.

### DIFF
--- a/ckanext/canada/logic.py
+++ b/ckanext/canada/logic.py
@@ -781,7 +781,9 @@ def canada_user_update(up_func: Action,
     user_obj = model.User.get(id)
     if user_obj is None:
         raise ObjectNotFound('User was not found.')
-    context['user_obj'] = user_obj
+
+    # fresh context copy
+    context = cast(Context, dict(context, user_obj=user_obj))
     extras = context['user_obj'].plugin_extras or {}
 
     # get custom validators for custom fields
@@ -813,8 +815,8 @@ def canada_user_update(up_func: Action,
         data['opt_in_features__pd_datatables'] = extras.get(
             'opt_in_features__pd_datatables', False)
 
-    if 'plugin_extras' not in data_dict:
-        data_dict['plugin_extras'] = {}
+    # always use current("before") extras to prevent arbitrary key/values
+    data_dict['plugin_extras'] = extras
 
     # add our validated values to plugin_extras for core to save
     data_dict['plugin_extras']['opt_in_features__pd_datatables'] = \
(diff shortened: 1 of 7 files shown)


ckanext/canada/logic.py
```diff
@@ -773,41 +773,64 @@ def canada_user_update(up_func: Action,
...
     Add new fields to the User schema.
+
+    - opt_in_features__pd_datatables
     """
+    # get user object as we need existing plugin_extras
+    id = get_or_bust(data_dict, 'id')
+    user_obj = model.User.get(id)
+    if user_obj is None:
+        raise ObjectNotFound('User was not found.')
+    context['user_obj'] = user_obj
```
wardi: I prefer to create a new context instead of modifying the one passed by the caller. modifying the passed context (or reusing passed contexts) can cause all sorts of difficult-to-diagnose bugs.

```suggestion
    context = cast(Context, dict(context, user_obj=user_obj))
```

ckanext/canada/logic.py
```diff
@@ -773,41 +773,64 @@ def canada_user_update(up_func: Action,
...
         model.Session.rollback()
         raise ValidationError(errors)
+
     if 'opt_in_features__pd_datatables' not in data:
-        return up_func(context, data_dict)
+        # use existing or default value
+        data['opt_in_features__pd_datatables'] = extras.get(
+            'opt_in_features__pd_datatables', False)
+
     if 'plugin_extras' not in data_dict:
```
wardi: just realizing, normally plugin_extras can only be set directly in the API by sysadmin users but this override looks like it allows any user to set arbitrary plugin_extras. That's not something we want, right?

JVickery-TBS: mmm yeah I suppose so. we only want to limit to what we put into plugin_extras via this schema, ignoring anything else. So basically just always set plugin_extras to an empty dict here? and then set our key values?

wardi: Or set it to what was in the db before in case there are other fields we need to maintain.

PR #1690, opened 2026-07-20
recombinant edit_using__id: true support for forms and preview

wardi:
required changes to the recombinant functionality that lives in the ckanext-canada extension for https://github.com/open-data/ckanext-recombinant/pull/155

here we're displaying and labeling the datastore `_id` column as `open_canada_id` with the expectation that we will export a column named `open_canada_id` in the combined CSV version of PD types that use `edit_using__id: true`

Only the deprecated preview template was updated as that's the one I'm familiar with. Similiar changes will be required in the new PD preview.


JVickery-TBS:
yeaahhh we really gotta move the HTML webforms to Recombinant plugin. Will see about the new PD datatables for this.

JVickery-TBS:
@wardi okay I have added in the `edit_using__id` stuff into the new PD Datatable stuff. So the `open_canada_id` (_id) will display if chromo.edit_using__id. And then the "Edit in Excel" and "Edit Online" buttons have been updated to use the _id. And then the custom Datatable Editor has the _id stuff in it now, allowing editing the Primary Keys.

the datatable editor already only does insert or update: `'method': editingRows.length > 0 ? 'update' : 'insert',`

JVickery-TBS:
@wardi okay so would we just do something like this in the triggers:
```
IF NEW._id IS NULL THEN
    NEW._id := nextval('reclassification_open_canada_id_seq');
END IF;
```

wardi:
No, we set the `_id` on `INSERT` and require it to exist on `UPDATE`, like in
https://github.com/open-data/ckanext-recombinant/pull/155/changes#diff-d85953e2b34aacb9f6b4d28dc596f71504c377d40a3f098e64b4d4a12cd65edcR67-R72

Although I guess when migrating data we'll need to be able to set during INSERT for sysadmin users :thinking: I wonder what the cleanest way to handle that would be. Noting that we'll need to reset any sequence that we depend on when we copy data.

wardi:
@JVickery-TBS ok, I've got `_id` import working (for data migrations) but it took changes here as well as in ckan and ckanext-canada. I've added an `importing` flag to the `datastore_user` table for the trigger and a `datastore_import` context variable that is set in the recombinant load-csv cli.

Next I'll update the `datastore_sequence_create` API to allow setting the next sequence number so we can sync from the original to the target ckan instance. If we adopt my suggested numbering scheme we can just set the sequence to `max(_id) / 100` after migrating data.

PR #1703, opened 2026-08-06
Dataset Scheming Form Pages [OPEN-4034]

JVickery-TBS:
recreation of https://github.com/open-data/ckanext-canada/pull/1582 due to partial merge to prevent ongoing merge conflicts


ckanext/canada/schemas/presets.pages.yaml
```diff
@@ -246,7 +246,7 @@ presets:
     form_snippet: fluent_text.html
     display_snippet: fluent_text.html
     error_snippet: fluent_text.html
-    validators: fluent_text
+    validators: fluent_text scheming_required
```
wardi, line 249: hmm. this is fine for now but might be better to fix in ckanext-fluent

wardi (approved):

PR #1707, opened 2026-09-04
add translations for data dictionary field types

RabiaSajjad:


wardi:
as I said in the ckan PR, if we add the gettext call to a template here instead this looks good to me.

ckanext/canada/templates/datastore/snippets/dictionary_view.html
```diff
@@ -1,5 +1,8 @@
 {% ckan_extends %}
 
+{% block dictionary_field_type%}
```
JVickery-TBS: needs space after dictionary_field_type

JVickery-TBS (changes requested):

